PR #3, opened 2024-03-19
[KDESKTOP-136] Post-pone the deletion of synchronizations, drives and users with a timer to get a more responsive GUI

luc-guyot-infomaniak:
This merge request addresses the issue [KDESKTOP-136](https://infomaniak.atlassian.net/browse/KDESKTOP-136), formerly [#726](https://github.com/infomaniak/desktop-app/multi/kdrive/-/issues/726).

The proposed changes aim at keeping the GUI thread responsive when some long tasks related to synchronization deletion are triggered by the user.
Our changes post-pone, within the event loop of the Qt application server, each of the following three tasks by means of a `QTimer`:
- the deletion of an individual synchronization,
- the deletion of a drive,
- the deletion of a user.

Because each of the aforementioned tasks calls stopVfs with the unregister parameter set with true, these tasks can be long on the Windows platform and cause the GUI to be frozen with no explicit reason.
The proposed changes explicitly lock all the actions which could interfere with synchronization deletions, letting the user know that a deletion process is ongoing. The GUI actions are unlock on deletion completion or when deletion fails.

**Note.** Tests on Windows show that the deletion of a drive, can be extremely or infinitely long. To be investigated.

[KDESKTOP-136]: https://infomaniak.atlassian.net/browse/KDESKTOP-136?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/gui/drivepreferenceswidget.cpp
```diff
@@ -1042,17 +1057,30 @@ void DrivePreferencesWidget::onErrorAdded() {
 }
 
 void DrivePreferencesWidget::onRemoveDrive(bool checked) {
-    Q_UNUSED(checked)
+    Q_UNUSED(checked);
```
ChristopheLarchier, line 1060: Q_UNUSED macro already contains the ";" ;)
#define Q_UNUSED(x) (void)x;

src/gui/folderitemwidget.cpp
```diff
@@ -253,89 +264,114 @@ void FolderItemWidget::closeFolderView() {
     //    emit displayFolderDetail(_syncDbId, _isExpanded);
 }
 
-void FolderItemWidget::onMenuButtonClicked() {
-    MenuWidget *menu = new MenuWidget(MenuWidget::Menu, this);
 
+SyncInfoClient *FolderItemWidget::getSyncInfoClient() const noexcept {
```
ChristopheLarchier, line 268: I would prefer returning an std::optional<SyncInfoClient &>

src/gui/guirequests.h
```diff
@@ -112,13 +114,15 @@ struct GuiRequests {
...
         static ExitCode deleteDrive(int driveDbId);  // !!! Use COMM_LONG_TIMEOUT !!!
-        static ExitCode deleteSync(int syncDbId);    // !!! Use COMM_LONG_TIMEOUT !!!
+        static ExitCode deleteSync(int syncDbId);    // Asynchronous because it can be time consuming
         static ExitCode propagateSyncListChange(int syncDbId, bool restartSync);
         static ExitCode bestAvailableVfsMode(VirtualFileMode &mode);
         static ExitCode propagateExcludeListChange();  // !!! Use COMM_LONG_TIMEOUT !!!
         static ExitCode hasSystemLaunchOnStartup(bool &enabled);
         static ExitCode hasLaunchOnStartup(bool &enabled);
         static ExitCode setLaunchOnStartup(bool enabled);
+        static ExitCode isSyncDeletionOnGoing(
```
ChristopheLarchier, line 124: Still used?

ChristopheLarchier (commented):
Done. See comments.

src/gui/drivepreferenceswidget.cpp
```diff
@@ -1070,24 +1098,25 @@ void DrivePreferencesWidget::onUnsyncTriggered(int syncDbId) {
...
+        folderBloc->setCustomToolTipText(tr("This synchronization is being deleted."));
+
+        // Remove sync
+        const ExitCode exitCode = GuiRequests::deleteSync(syncDbId);
+        if (exitCode != ExitCodeOk) {
+            qCWarning(lcDrivePreferencesWidget()) << "Error in Requests::removeSync";
+            CustomMessageBox msgBox(QMessageBox::Warning,
+                                    tr("Failed to stop syncing the folder <i>%1</i>.").arg(syncInfoMapIt->second.localPath()),
+                                    QMessageBox::Ok, this);
+            return;
```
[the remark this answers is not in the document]
luc-guyot-infomaniak: Thanks, done.

src/gui/drivepreferenceswidget.cpp
luc-guyot-infomaniak: Thanks, done.

src/gui/guirequests.h
luc-guyot-infomaniak: Ooops! Fixed, thanks (obsolete declaration removed).

src/gui/folderitemwidget.cpp
luc-guyot-infomaniak: The C++ compiler won't let me declare any type of the form `std::optional<T&>` for some type `T` (this was my first idea and first attempt). After a second thought, I came to the conclusion that access via pointer is a simple and common interface while it isn't much more risky that access via reference.

src/gui/clientgui.cpp
```diff
@@ -1233,9 +1233,17 @@ void ClientGui::onDriveRemoved(int driveDbId) {
 
         // Erase drive
         driveInfoMapIt = _driveInfoMap.erase(driveInfoMapIt);
+
+        // Handle the case of the current drive being erased
         if (_currentDriveDbId == driveDbId) {
-            // Select next drive if there is one
-            _currentDriveDbId = !_driveInfoMap.empty() ? driveInfoMapIt->first : 0;
+            // Select the next drive if there is one.
+            if (driveInfoMapIt != _driveInfoMap.cend()) {
```
luc-guyot-infomaniak, line 1240: This the fix: if the erased drive was the last item of the drive info map, then `driveInfoMapIt` is the `end()` iterator and cannot be dereferenced.

ClementKunz: Tested on Windows, all good 👍

ClementKunz (approved):

src/gui/folderitemwidget.cpp
ChristopheLarchier: OK...

PR #24, opened 2024-03-27
Fixes two issues related to IoHelper::getItemType and the detection of links

luc-guyot-infomaniak:
The first issue was reported by @ChristopheLarchier today: the value of `_linkType` is not set anymore in `uploadjob.cpp` due to an accidental change while introducing `IoHelper::getItemType` in this file.

The second issue is also visible in this file: the error returned by `IoHelper::getItemType` should be `IoErrorNoSuchFileOrDirectory` if and only if the item argument could not be found. Before these changes, the function would return `IoErrorNoSuchFileOrDirectory` if the item is an existing symlink with an invalid target path.

Unit tests have been adapted.

### DIFF
--- a/src/libcommonserver/io/iohelper.cpp
+++ b/src/libcommonserver/io/iohelper.cpp
@@ -148,11 +148,12 @@ bool IoHelper::_isExpectedError(IoError ioError) noexcept {
 bool IoHelper::_setTargetType(ItemType &itemType) noexcept {
     std::error_code ec;
     const bool isDir = _isDirectory(itemType.targetPath, ec);
-    itemType.ioError = stdError2ioError(ec);
+    IoError ioError = stdError2ioError(ec);
 
-    if (itemType.ioError != IoErrorSuccess) {
-        const bool expected = _isExpectedError(itemType.ioError);
+    if (ioError != IoErrorSuccess) {
+        const bool expected = _isExpectedError(ioError);
         if (!expected) {
+            itemType.ioError = ioError;
             LOGW_WARN(Log::instance()->getLogger(), L"Failed to check if the item is a directory: "
                                                         << Utility::formatStdError(itemType.targetPath, ec).c_str());
         }
--- a/src/libsyncengine/jobs/network/uploadjob.cpp
+++ b/src/libsyncengine/jobs/network/uploadjob.cpp
@@ -146,41 +146,41 @@ void UploadJob::setQueryParameters(Poco::URI &uri, bool &canceled) {
     canceled = false;
 }
 
+namespace details {
+inline bool isLink(LinkType linkType) {
+    return linkType == LinkTypeSymlink || linkType == LinkTypeHardlink ||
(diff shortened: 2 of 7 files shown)


src/libsyncengine/jobs/network/uploadjob.cpp
```diff
@@ -146,41 +146,41 @@ void UploadJob::setQueryParameters(Poco::URI &uri, bool &canceled) {
...
-        LOGW_DEBUG(KDC::Log::instance()->getLogger(), L"Item misses search permission - path=" << Path2WStr(_filePath).c_str());
+        LOGW_DEBUG(_logger, L"Item misses search permission - " << Utility::formatSyncPath(_filePath).c_str());
         canceled = true;
         return;
     }
 
-    if (_linkType == LinkTypeSymlink || _linkType == LinkTypeHardlink ||
-        (_linkType == LinkTypeFinderAlias && OldUtility::isMac()) || (_linkType == LinkTypeJunction && OldUtility::isWindows())) {
-        // Link
+    _linkType = itemType.linkType;
```
luc-guyot-infomaniak, line 176: The missing assignment.

src/libcommonserver/io/iohelper.cpp
```diff
@@ -148,11 +148,12 @@ bool IoHelper::_isExpectedError(IoError ioError) noexcept {
...
     const bool isDir = _isDirectory(itemType.targetPath, ec);
-    itemType.ioError = stdError2ioError(ec);
+    IoError ioError = stdError2ioError(ec);
 
-    if (itemType.ioError != IoErrorSuccess) {
-        const bool expected = _isExpectedError(itemType.ioError);
+    if (ioError != IoErrorSuccess) {
+        const bool expected = _isExpectedError(ioError);
         if (!expected) {
+            itemType.ioError = ioError;
```
luc-guyot-infomaniak, line 156: Change the returned error only if it is unexpected.

src/libsyncengine/jobs/network/uploadjob.cpp
```diff
@@ -146,41 +146,41 @@ void UploadJob::setQueryParameters(Poco::URI &uri, bool &canceled) {
...
-        LOGW_DEBUG(KDC::Log::instance()->getLogger(), L"Item does not exist anymore - path=" << Path2WStr(_filePath).c_str());
-        canceled = true;
+        LOGW_DEBUG(_logger, L"Item does not exist anymore - " << Utility::formatSyncPath(_filePath).c_str());
         return;
     }
 
     if (itemType.ioError == IoErrorAccessDenied) {
-        LOGW_DEBUG(KDC::Log::instance()->getLogger(), L"Item misses search permission - path=" << Path2WStr(_filePath).c_str());
+        LOGW_DEBUG(_logger, L"Item misses search permission - " << Utility::formatSyncPath(_filePath).c_str());
         canceled = true;
```
ClementKunz, line 172: Canceled is already true with your latest changes. For readability sake, we might want to remove this line

luc-guyot-infomaniak: Good catch, thanks! Done.

ClementKunz (approved):

PR #31, opened 2024-04-03
KDESKTOP-701-Add-logs-in-case-of-upload-issue

ClementKunz:
Add more log to understand upload issues

### DIFF
--- a/src/libsyncengine/jobs/network/abstractnetworkjob.cpp
+++ b/src/libsyncengine/jobs/network/abstractnetworkjob.cpp
@@ -319,7 +319,7 @@ bool AbstractNetworkJob::sendRequest(Poco::Net::HTTPSClientSession &session, con
         stream.push_back(session.sendRequest(req));
         if (ioOrLogicalErrorOccurred(stream[0].get())) {
             int err = session.socket().getError();
-            LOG_WARN(_logger, "sendRequest failed for job " << jobId() << " - err=" << err);
+            LOG_WARN(_logger, "sendRequest failed for job " << jobId() << " - socket err=" << err);
             return processSocketError(session, err);
         }
     } catch (Poco::Exception &e) {
@@ -343,7 +343,7 @@ bool AbstractNetworkJob::sendRequest(Poco::Net::HTTPSClientSession &session, con
             stream[0].get() << std::string(itBegin, itEnd);
             if (ioOrLogicalErrorOccurred(stream[0].get())) {
                 int err = session.socket().getError();
-                LOG_WARN(_logger, "sendRequest failed for job " << jobId() << " - err=" << err);
+                LOG_WARN(_logger, "sendRequest failed for job " << jobId() << " - socket err=" << err);
                 return processSocketError(session, err);
             }
         } catch (Poco::Exception &e) {
@@ -524,7 +524,14 @@ bool AbstractNetworkJob::processSocketError(Poco::Net::HTTPSClientSession &sessi
 }
 
 bool AbstractNetworkJob::ioOrLogicalErrorOccurred(std::ios &stream) {
-    return ((stream.fail() || stream.bad()) && !stream.eof());
+    bool res = (stream.fail() || stream.bad()) && !stream.eof();
+    if (res) {
+        LOG_DEBUG(_logger, "sendRequest failed for job " << jobId()
+                                                         << " - stream fail=" << stream.fail()
+                                                         << " - stream bad=" << stream.fail()
+                                                         << " - stream eof=" << stream.fail());
+    }
+    return res;
 }
 
 const std::string AbstractNetworkJob::errorText(const Poco::Exception &e) const {


src/libsyncengine/jobs/network/abstractnetworkjob.cpp
```diff
@@ -319,7 +319,7 @@ bool AbstractNetworkJob::sendRequest(Poco::Net::HTTPSClientSession &session, con
         stream.push_back(session.sendRequest(req));
         if (ioOrLogicalErrorOccurred(stream[0].get())) {
             int err = session.socket().getError();
-            LOG_WARN(_logger, "sendRequest failed for job " << jobId() << " - err=" << err);
+            LOG_WARN(_logger, "sendRequest failed for job " << jobId() << " - socket err=" << err);
```
luc-guyot-infomaniak, line 322: If such error message formatting can be reused more than twice in the code, you may create and use a formatter function.

src/libsyncengine/jobs/network/abstractnetworkjob.cpp
```diff
@@ -524,7 +524,14 @@ bool AbstractNetworkJob::processSocketError(Poco::Net::HTTPSClientSession &sessi
 }
 
 bool AbstractNetworkJob::ioOrLogicalErrorOccurred(std::ios &stream) {
-    return ((stream.fail() || stream.bad()) && !stream.eof());
+    bool res = (stream.fail() || stream.bad()) && !stream.eof();
+    if (res) {
+        LOG_DEBUG(_logger, "sendRequest failed for job " << jobId()
+                                                         << " - stream fail=" << stream.fail()
+                                                         << " - stream bad=" << stream.fail()
```
luc-guyot-infomaniak, line 531:
```suggestion
                                                         << " - stream bad=" << stream.bad()
```

src/libsyncengine/jobs/network/abstractnetworkjob.cpp
```diff
@@ -524,7 +524,14 @@ bool AbstractNetworkJob::processSocketError(Poco::Net::HTTPSClientSession &sessi
...
+                                                         << " - stream eof=" << stream.fail());
```
luc-guyot-infomaniak, line 532:
```suggestion
                                                         << " - stream eof=" << stream.eof());
```

src/libsyncengine/jobs/network/abstractnetworkjob.cpp
ClementKunz: Indeed, thanks for your careful review

src/libsyncengine/jobs/network/abstractnetworkjob.cpp
ClementKunz: I don't know because we want to avoid copying string by using StringStream here. Do you have any suggestion?

luc-guyot-infomaniak: I was thinking about something like this:

```
std::string formatSocketJobError(UniqueId jobId, int errorCode) {
    std::stringstream ss;
    ss << "job='" << jobId << "', socket error='" << errorCode << "'";

    return ss.str();
}
```

that would be used in the following way:

```

LOG_WARN(_logger, "sendRequest:  " << formatSocketJobError(jobId, errorCode));
```

It becomes interesting if it can be reused in numerous places. This way we ensure a uniform format that can be improved or fixed by modifying a single function. 

Performance becomes a concern if many logs of this kind can be issued during normal use of the application.
It is indeed true that the use of `std::stringstream` is much higher than string concatenation (see e.g., https://stackoverflow.com/questions/30254175/is-stringstream-better-than-strings-operator-for-string-objects-concatenati).

Alternatively:

 ```
std::string formatSocketJobError(UniqueId jobId, int errorCode) {
    std::string result;
    result.reserve(100); // There might be a better guess.
    
   return result.append("job='").append(std::to_string(jobId)).append("', socket error='").append(std::to_string(errorCode)).append("'"); 
}
```
 If there is no strong potential for re-use, just skip this comment.

Comment (ClementKunz) [reply src/libsyncengine/jobs/network/abstractnetworkjob.cpp:322]: Those solution implies string copy. You will build a first stream, extract the string, copy it as function returned value to inject it again into another stream.
It is ok for logs that does not happen to often (which is the case here, with Warning and Erros).
Also, as mention, this particuliar log is used twice only in the code. Perhaps we should think about an even more generic function that could be used for any error, instead of doing multiple formatter functions.
If you agreed I would prefer to make it the topic of another PR

**Review by luc-guyot-infomaniak — APPROVED** [review]

File: src/libsyncengine/jobs/network/abstractnetworkjob.cpp (hunk 1 above)
Comment (luc-guyot-infomaniak) [inline src/libsyncengine/jobs/network/abstractnetworkjob.cpp:322]: I have been reading the logs attached to redmine issues today and I realize that `Poco::getMessage(int errorCode)` could be useful here.

File: src/libsyncengine/jobs/network/abstractnetworkjob.cpp [hunk 4]
```diff
@@ -524,7 +524,14 @@ bool AbstractNetworkJob::processSocketError(Poco::Net::HTTPSClientSession &sessi
 }
 
 bool AbstractNetworkJob::ioOrLogicalErrorOccurred(std::ios &stream) {
-    return ((stream.fail() || stream.bad()) && !stream.eof());
+    bool res = (stream.fail() || stream.bad()) && !stream.eof();
```
luc-guyot-infomaniak, line 527: Is it possible to reuse `ioOrLogicalErrorOccurred` here?

src/libsyncengine/jobs/network/abstractnetworkjob.cpp
ClementKunz: Agreed, its done

PR #39, opened 2024-04-08
[KDESKTOP-688] Windows CI configuration

MaximePouce:


.github/workflows/windows.yml
```diff
@@ -0,0 +1,42 @@
...
+  pull_request:
+    types: [synchronize, review_requested]
+
+concurrency:
+  group: ${{ github.head_ref }}
+  cancel-in-progress: true
+
+jobs:
+  build-kDrive:
+    runs-on: [ self-hosted, Windows, desktop-kdrive ]
```
luc-guyot-infomaniak, line 13: As the name of the repository is `desktop-kDrive` and `kDrive` with "D" in upper-case, we may apply the same convention here.

extensions/windows/cfapi/Vfs/Vfs.vcxproj
```diff
@@ -189,11 +189,8 @@ copy /Y "$(TargetDir)$(TargetName).pdb" "F:\Projects\build-kdrive-Desktop_Qt_6_2
     </Link>
     <Manifest />
     <PostBuildEvent>
-      <Command>copy /Y "$(ProjectDir)..\Common\debug.h" "F:\Projects\kdrive\src\server\vfs\win\."
-copy /Y "$(ProjectDir)$(TargetName).h" "F:\Projects\kdrive\src\server\vfs\win\."
```
luc-guyot-infomaniak, line 193: Should we adapt the build instructions for Windows? If I am correct, users who want to install `desktop-kDrive` are invited to edit the Build Events of the `Vfs` project in the documentation, i.e., to replace  `F:\Projects` with the appropriate folder.

infomaniak-build-tools/windows/build-drive.ps1
```diff
@@ -313,6 +315,14 @@ if ($LASTEXITCODE -ne 0)
 	exit $LASTEXITCODE
 }
 
+# The CI tester will exit before creating and signing the package
```
luc-guyot-infomaniak, line 318:
```suggestion
# The CI tester will exit before creating and signing the package.
```

extensions/windows/cfapi/Vfs/Vfs.vcxproj
MaximePouce: The Post-Build instructions need to be updated to remove that hardcoded`F:\Projects` path in Debug, however this change is only removing all the Post-Build instructions for **Release** builds, as the extension file management is done by my powershell script

.github/workflows/windows.yml
MaximePouce: This section is selecting runners based on tags, so I'm using full lower-case as this is how I created the tag, I cannot 100% confirm that they are case sensitive but to avoid potential errors I'm keeping it in lower-case (the `Windows` tag was automatically created)

luc-guyot-infomaniak (approved):

PR #45, opened 2024-04-11
KDESKTOP-684-Operation-generator-fail-if-an-item-with-DELETE-operation-is-blacklisted

ClementKunz:
Following [issue](https://infomaniak.atlassian.net/browse/KDESKTOP-393) the local items that were blacklisted remains in the update tree. This should not be the case, no excluded items should appear in the update tree.

### DIFF
--- a/src/libsyncengine/reconciliation/platform_inconsistency_checker/platforminconsistencycheckerworker.cpp
+++ b/src/libsyncengine/reconciliation/platform_inconsistency_checker/platforminconsistencycheckerworker.cpp
@@ -38,8 +38,9 @@ void PlatformInconsistencyCheckerWorker::execute() {
 
     ExitCode exitCode = checkTree(_syncPal->_remoteUpdateTree->rootNode(), _syncPal->_remoteUpdateTree->rootNode()->name());
 
-    for (const auto &id : _idsToBeRemoved) {
-        _syncPal->_remoteUpdateTree->deleteNode(id);
+    for (const auto &idItem : _idsToBeRemoved) {
+        _syncPal->_remoteUpdateTree->deleteNode(idItem.first);
+        _syncPal->_localUpdateTree->deleteNode(idItem.second);
     }
 
     std::chrono::duration<double> elapsed_seconds = std::chrono::steady_clock::now() - start;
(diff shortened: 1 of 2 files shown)


src/libsyncengine/reconciliation/platform_inconsistency_checker/platforminconsistencycheckerworker.h
```diff
@@ -38,7 +38,7 @@ class PlatformInconsistencyCheckerWorker : public OperationProcessor {
         bool checkPathAndName(std::shared_ptr<Node> remoteNode);
         void checkNameClashAgainstSiblings(std::shared_ptr<Node> remoteParentNode);
 
-        std::list<NodeId> _idsToBeRemoved;
+        std::list<std::pair<NodeId, NodeId>> _idsToBeRemoved;   // first item = remote ID (mandatory), second item = local ID (optional)
```
luc-guyot-infomaniak, line 41:
```suggestion
        struct NodeIdPair {
              NodeId local;
              NodeId remote;
        };
        std::list<NodeIdPair> _idsToBeRemoved;   // first item = remote ID (mandatory), second item = local ID (optional)
```

ClementKunz: Done

src/libsyncengine/reconciliation/platform_inconsistency_checker/platforminconsistencycheckerworker.cpp
```diff
@@ -132,9 +136,10 @@ void PlatformInconsistencyCheckerWorker::blacklistNode(const std::shared_ptr<Nod
         default:
             break;
     }
-    LOGW_SYNCPAL_INFO(_logger, L"Blacklisting remote item '" << Utility::formatSyncPath(relativePath).c_str() << L"' because " << causeStr.c_str() << L".");
+    LOGW_SYNCPAL_INFO(_logger, L"Blacklisting remote item with " << Utility::formatSyncPath(relativePath).c_str() << L" because " << causeStr.c_str() << L".");
```
MaximePouce, line 139: With current change a log would look like `Blacklisting remote item with [path]`
If this change is to indicate both the remote and local items were blacklisted, it will be clearer to just remove the `remote` part

```suggestion
    LOGW_SYNCPAL_INFO(_logger, L"Blacklisting item " << Utility::formatSyncPath(relativePath).c_str() << L" because " << causeStr.c_str() << L".");
```

ClementKunz: Most of the time, at least on first sync, only remote items will be blacklisted because items are not yet synchronized.
The only case where we blacklist both is when a rename operation generated a name clash on 2 synchronized items.

luc-guyot-infomaniak (approved):

PR #57, opened 2024-04-16
KDESKTOP-757 - provide to logger as function parameter

ClementKunz:
The app crashed after deleteing the folder used as root for an advanced synchronisation.
This issue reveal a deeper problem: the logger instance used in `Utility::logGenericServerError` was not correctly defined

### DIFF
--- a/src/libcommonserver/utility/utility.cpp
+++ b/src/libcommonserver/utility/utility.cpp
@@ -273,13 +273,13 @@ std::string Utility::formatGenericServerError(std::istream &inputStream, const P
     return errorStream.str();   // str() return a copy of the underlying string
 }
 
-void Utility::logGenericServerError(const std::string &errorTitle, std::istream &inputStream,
+void Utility::logGenericServerError(const log4cplus::Logger &logger, const std::string &errorTitle, std::istream &inputStream,
                                     const Poco::Net::HTTPResponse &httpResponse) {
     std::string errorMsg = formatGenericServerError(inputStream, httpResponse);
 #ifdef NDEBUG
     sentry_capture_event(sentry_value_new_message_event(SENTRY_LEVEL_WARNING, errorTitle.c_str(), errorMsg.c_str()));
 #endif
-    LOG_WARN(_logger, errorTitle.c_str() << ": " << errorMsg.c_str());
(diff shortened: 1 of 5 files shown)


src/libcommonserver/utility/utility.cpp
```diff
@@ -273,13 +273,13 @@ std::string Utility::formatGenericServerError(std::istream &inputStream, const P
...
 }
 
-void Utility::logGenericServerError(const std::string &errorTitle, std::istream &inputStream,
+void Utility::logGenericServerError(const log4cplus::Logger &logger, const std::string &errorTitle, std::istream &inputStream,
                                     const Poco::Net::HTTPResponse &httpResponse) {
     std::string errorMsg = formatGenericServerError(inputStream, httpResponse);
 #ifdef NDEBUG
     sentry_capture_event(sentry_value_new_message_event(SENTRY_LEVEL_WARNING, errorTitle.c_str(), errorMsg.c_str()));
 #endif
-    LOG_WARN(_logger, errorTitle.c_str() << ": " << errorMsg.c_str());
```
luc-guyot-infomaniak, line 282: What to do with all the other uses of `Utility::_logger`? It seems to relate to https://infomaniak.atlassian.net/browse/KDESKTOP-738.

luc-guyot-infomaniak (approved):

src/libcommonserver/utility/utility.cpp
ClementKunz: I did not see any crash related to other call to logger. Do you think we should apply the same strategy everywhere in this Utility file?

luc-guyot-infomaniak: I cannot tell without further investigation. I just wanted to know if you have clear understanding of the reason of this specific crash and if this fix can or should apply at other places. We can certainly wait and address this with https://infomaniak.atlassian.net/browse/KDESKTOP-738.

ClementKunz: I definitly agree that we should address https://infomaniak.atlassian.net/browse/KDESKTOP-738, perhaps we should discuss the priority of this issue in next meeting.
Since this fix will be released in the next version, I prefer keeping the changes to a minimum. I will open other issues if I detect other crashes

PR #59, opened 2024-04-17
Kdesktop 735 enhance the log sender 

herve-er:
- Generate a zip archive containing user logs and parmsDb.
- Provide an option to estimate the file size.
- Offer the choice between sending all logs or only those from the current session.
- Implement the initial version of UI integration (Parameters -> Preferences -> Debug -> SEND LOG).
- Send the generated archive to support via the new backend route. [DRAFT | IN PROGRESS]
- Display a success message to the user, including the logName for easy identification by the support (useful in cases where a non-logged user encounters issues referencing Jira issue KDESKTOP-747).


herve-er:
This pull request had been divided in smaller ones.

PR #60, opened 2024-04-17
[HOTFIX] Open local folder of a new drive instead of its parent 

luc-guyot-infomaniak:
This bug has been detected on MacOSX, but should be visible on every operaring system: the method `QFileInfo:path()` removes the name of the item from its output path, see https://doc.qt.io/qt-6/qfileinfo.html#path.

### DIFF
--- a/src/gui/guiutility.cpp
+++ b/src/gui/guiutility.cpp
@@ -480,6 +480,10 @@ qint64 GuiUtility::folderDiskSize(const QString &dirPath) {
     return total;
 }
 
+QString GuiUtility::getFolderPath(const QString &path, NodeType nodeType) {
+    return nodeType == NodeTypeDirectory ? path : QFileInfo(path).path();
+}
+
 bool GuiUtility::openFolder(const QString &dirPath) {
     if (dirPath.isEmpty()) return true;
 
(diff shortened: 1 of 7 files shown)


src/gui/guiutility.cpp
```diff
@@ -481,28 +481,20 @@ qint64 GuiUtility::folderDiskSize(const QString &dirPath) {
...
-        } else {
-            return false;
-        }
+    if (dirPath.isEmpty()) return true;
+
+    QFileInfo fileInfo(dirPath);
+    if (fileInfo.exists()) {
+        const QUrl url = getUrlFromLocalPath(QDir::cleanPath(fileInfo.filePath()));
+        if (url.isValid() && !QDesktopServices::openUrl(url)) return false;
+    } else if (fileInfo.dir().exists()) {
```
luc-guyot-infomaniak, line 489: @ClementKunz  Here is the fallback strategy followed when the path indicates an invalid item: we try attempt to open the parent folder, if any.

src/libsyncengine/jobs/network/downloadjob.cpp
```diff
@@ -207,16 +207,15 @@ bool DownloadJob::handleResponse(std::istream &is) {
     } else {
```
luc-guyot-infomaniak, line 207: These changes are required for successful compilation on MacOSX but are not part, per se, of this PR. See earlier dedicated PR fixing the compilation issue.

ClementKunz (approved):

PR #76, opened 2024-04-26
Kdesktop 783 implement a new table key value in parms db

herve-er:
This new table will allow us to store value under a key and to retrieve it.
It can be useful in different situation:
- Save the last time of serverSelfRestart.
- idem for the client.
- save the timestamp of the last successful log upload.
- save the path of the log archive in case of upload fail
- ….

A request will also be set up to allow the client to set/retrieve data from this table.

PR #77, opened 2024-04-26
Kdesktop 786 update self restarter

herve-er:
Update the self restarter to use the new keyValue tabs in the parms database, and remove the selfrestarter tab.


src/libparms/db/parmsdb.h
```diff
@@ -122,24 +122,18 @@ class PARMS_EXPORT ParmsDb : public Db {
...
     private:
         static std::shared_ptr<ParmsDb> _instance;
         bool _test;
 
         ParmsDb(const std::filesystem::path &dbPath, const std::string &version, bool autoDelete, bool test);
 
         bool insertDefaultParameters();
-        bool insertDefaultSelfRestarterData();
         bool insertDefaultAppState();
+        bool insertDefaultAppState(AppStateKey key, const std::string &value);
```
ClementKunz, line 136: not very clear to me why this is declared in `parmsdb.h` but defined in `parmsdbappstate.cpp` while there is a `parmsdbappstate.h`.
Is it possible to make ParmsDbAppState as a separated class, perhaps a child of ParmsDb is there is some methods defined in ParmsDb that you need in ParmsDbAppState

src/libparms/db/parmsdbappstate.h
```diff
@@ -35,4 +35,5 @@
 #define UPDATE_APP_STATE_REQUEST_ID "update_value_with_key"
 #define UPDATE_APP_STATE_REQUEST "UPDATE app_state SET value=?2 WHERE key=?1;"
 
-#define APP_STATE_KEY_DEFAULT_Test "Test"
\ No newline at end of file
+#define APP_STATE_KEY_DEFAULT_LastServerSelfRestart "0"
+#define APP_STATE_KEY_DEFAULT_LastClientSelfRestart "0"
```
ClementKunz, line 39: As described in the SonarCloud issue, prefer using const value instead of macro (the remaining of the code will be refactored later, but let's apply it now on new code)

src/server/appserver.h
```diff
@@ -125,6 +125,7 @@ class AppServer : public SharedTools::QtSingleApplication {
         bool _vfsActivationDone = false;
         bool _vfsConnectionDone = false;
         bool _crashRecovered = false;
+        QString _crashMessage = tr("kDrive application will close due to a fatal error.");
```
ClementKunz, line 128: This does not need to be a attribute of the class since it never change. Make it a static const value defined in the cpp file (or wherever you find it relevant)

src/libparms/db/parmsdb.h
herve-er: As discussed with @luc-guyot-infomaniak in previous PR ([link to discussion](https://github.com/Infomaniak/desktop-kDrive/pull/76#discussion_r1580962129)), the goal here is to reduce the size of `parmsdb.cpp`, which is already more than 3000 lines long.

This reduction will improve the readability of `parmsdb` functions. For instance:

```cpp
ASSERT(queryCreate(CREATE_APP_STATE_TABLE_ID));
if (!queryPrepare(CREATE_APP_STATE_TABLE_ID, CREATE_APP_STATE_TABLE, false, errId, error))
{
    queryFree(CREATE_APP_STATE_TABLE_ID);
    return sqlFail(CREATE_APP_STATE_TABLE_ID, error);
}
if (!queryExec(CREATE_APP_STATE_TABLE_ID, errId, error)) {
    queryFree(CREATE_APP_STATE_TABLE_ID);
    return sqlFail(CREATE_APP_STATE_TABLE_ID, error);
}
queryFree(CREATE_APP_STATE_TABLE_ID);

//repeat for all tables
```

will become:

```cpp
createAppState();
createOtherTable();
...
```

The same simplification will apply to the prepare function.

Currently, only `parmsdbappstate.h/cpp` are done, but the plan is to do the same with other parts of `parmsdb.cpp` (e.g., `insertDrive()`, `updateDrive()`, `deleteDrive()` -> `parmsdbdrive.h/cpp`).

Making it a separate or child class seems to be quite complicated as we are still within the same database, which is protected with a mutex (this would require making the mutex a protected static member). Additionally, there's the question of database creation/upgrade, which needs to be synchronized for all potential child classes.

However, what could be done is to remove `parmsdbappstate.h`, whose only goal is to define constants (all the requests), but this can also be done in the cpp file, it was just to separate the sql queries definition.

src/libparms/db/parmsdbappstate.cpp
```diff
@@ -65,13 +65,25 @@ bool ParmsDb::prepareAppState() {
...
+    }
+
+    if (!insertDefaultAppState(AppStateKey::LastClientSelfRestart, APP_STATE_KEY_DEFAULT_LastClientSelfRestart)) {
+        return false;
+    }
+
+    return true;
+}
+
+bool ParmsDb::insertDefaultAppState(AppStateKey key, const std::string &value) {
```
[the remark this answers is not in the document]
herve-er: Exact, I will change in next commit.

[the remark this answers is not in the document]
herve-er: [812884e](https://github.com/Infomaniak/desktop-kDrive/pull/77/commits/812884e864c235bdbbf25a90ba5860b4696f8b89)

src/server/appserver.h
herve-er: I will edit this in next commit.

src/libparms/db/parmsdb.cpp
```diff
@@ -1407,20 +1317,6 @@ bool ParmsDb::upgrade(const std::string &fromVersion, const std::string & /*toVe
...
-            return sqlFail(CREATE_SELF_RESTARTER_TABLE_ID, error);
-        }
-
-        if (!queryExec(CREATE_SELF_RESTARTER_TABLE_ID, errId, error)) {
-            queryFree(CREATE_SELF_RESTARTER_TABLE_ID);
-            return sqlFail(CREATE_SELF_RESTARTER_TABLE_ID, error);
-        }
-        queryFree(CREATE_SELF_RESTARTER_TABLE_ID);
-
         createAppState();
```
ChristopheLarchier, line 1320: Return code to check

src/server/appserver.h
herve-er: [e25a99e](https://github.com/Infomaniak/desktop-kDrive/pull/77/commits/e25a99e535064117fe127658134fba60114dc153)

src/libparms/db/parmsdb.cpp
herve-er: [81ea943](https://github.com/Infomaniak/desktop-kDrive/pull/77/commits/81ea943b54acb16d9f335cbdf5ab41dc17c78bfe)

ClementKunz (approved):

PR #79, opened 2024-04-29
KDESKTOP-760-Manage-unvailable-items-when-LiteSync-is-activated

ClementKunz:
Cancel the hydration of the file if an error 404 is received and display an error message in the app.


src/libsyncengine/syncpal/syncpal.cpp
```diff
@@ -788,6 +788,23 @@ void SyncPal::directDownloadCallback(UniqueId jobId) {
         return;
     }
 
+    if (directDownloadJobsMapIt->second->getStatusCode() == Poco::Net::HTTPResponse::HTTP_NOT_FOUND) {
+        Error error(syncDbId()
```
luc-guyot-infomaniak, line 792: Just a thought for later: we may define derived error classes with suggestive names and simpler constructors. In this specific case, we would ideally pass only three "dynamic parameters" to the constructor let all the remaining parameters be initialized in the derived class constructor.

src/server/vfs/mac/litesyncextconnector.mm
```diff
@@ -1128,7 +1128,7 @@ bool checkIoErrorAndLogIfNeeded(IoError ioError, const std::string &itemType, co
                         }
 
                         if (![fileHandle writeData:buffer error:&error]) {
-                            LOGW_ERROR(_logger, L"Error while writting to file - path=" << QStr2WStr(filePath).c_str()
+                            LOGW_ERROR(_logger, L"Error while writing to file - path=" << QStr2WStr(filePath).c_str()
```
luc-guyot-infomaniak, line 1131:
```suggestion
                            LOGW_ERROR(_logger, L"Error while writing to: " << Utility::formatSyncPath(filePath).c_str()
```
If there is no dependency issue preventing us from using `KDC::Utility` here.

src/server/vfs/mac/litesyncextconnector.mm
```diff
@@ -1185,6 +1185,14 @@ bool checkIoErrorAndLogIfNeeded(IoError ioError, const std::string &itemType, co
     return true;
 }
 
+bool LiteSyncExtConnector::vfsCancelHydrate(const QString &filePath) {
+    if (!_private->updateFetchStatus(filePath, QString("CANCEL"))) {
+        LOGW_WARN(_logger, L"Call to updateFetchStatus failed - path=" << QStr2WStr(filePath).c_str());
```
luc-guyot-infomaniak, line 1190: Idem.

src/server/vfs/mac/litesyncextconnector.mm
ClementKunz: Since 'filePath' is a 'QString' here, I would suggest to keep it like that until the refactoring that will remove Qt everywhere

src/libsyncengine/syncpal/syncpal.cpp
ClementKunz: Yes I agree. The errors will also need refactoring to simplify its use.
Why not a very simple constructor and a few setters to specialize the error?

ClementKunz: Done, but this is only a minor changes in the error management system, we should deeply refactor it later

luc-guyot-infomaniak (approved):

PR #86, opened 2024-05-01
KDESKTOP-778-Client-crash-sometimes-after-removing-a-sync

ClementKunz:
This PR aim to fix crash in kDrive client when creating back a synchronisation. The crash seems linked to the way we disable child widgets during drive removal


ChristopheLarchier (approved):

src/gui/drivepreferenceswidget.cpp
```diff
@@ -1043,17 +1043,6 @@ void DrivePreferencesWidget::onDriveBeingRemoved() {
     assert(driveInfoIt != _gui->driveInfoMap().cend());
     driveInfoIt->second.setIsBeingDeleted(true);
 
-//    // Lock all GUI drive-related actions during drive deletion.
-//    GuiUtility::setEnabledRecursively<QLayout>(_mainVBox, false);
```
luc-guyot-infomaniak: I can confirm that removing this specific line prevents the application from crashing in `QWidget::testAttribute` after the operation **Remove all synchronizations** immediately followed by **Add drive**. 

I don't understand the problem yet, but this a truly good lead.

PR #92, opened 2024-05-02
Windows - Added an environment variable to force getRights fallback method

ChristopheLarchier:


src/libcommonserver/utility/utility_win.cpp
```diff
@@ -133,6 +133,13 @@ static bool init_private() {
 
     initTrusteeWithUserSID();
 
+    const char *getUseRightsFallbackMethod = std::getenv("KDRIVE_USE_GETRIGHTS_FALLBACK_METHOD");
```
ClementKunz, line 136: `CommonUtility::envVarValue` instead of `std::getenv`?

ChristopheLarchier: Oups... Done

ClementKunz (approved):

src/libcommon/utility/utility.cpp
```diff
@@ -664,6 +664,26 @@ bool CommonUtility::fileNameIsValid(const SyncName &name) {
...
 
+std::string CommonUtility::envVarValue(const std::string &name) {
+#ifdef _WIN32
+    char *value = nullptr;
+    size_t sz = 0;
+    if (_dupenv_s(&value, &sz, name.c_str()) == 0 && value != nullptr) {
+        std::string valueStr(value);
+        free(value);
+        return valueStr;
+    }
```
herve-er, line 675: Sonar is suggesting to avoid usage of free. What about :
`
if (size_t sz = 0; getenv_s(&sz, nullptr, 0, name.c_str()) == 0 && sz != 0) {
    auto value = std::make_unique<char[]>(sz);
    if (getenv_s(&sz, value.get(), sz, name.c_str()) == 0) {
        return std::string(value.get());
    }
}
`

PR #108, opened 2024-05-08
KDESKTOP-391-Still-issue-with-filenames-containing-on-Windows

ClementKunz:
Items that contains double quote in their name should be ignored


src/libsyncengine/jobs/network/csvfullfilelistwithcursorjob.cpp
```diff
@@ -63,6 +63,7 @@ bool CsvFullFileListWithCursorJob::getItem(SnapshotItem &item, bool &error, bool
     bool prevCharDoubleQuotes = false;
     bool readNextLine = true;
     std::string tmp;
+    uint doubleQuoteCounter = 0;
```
luc-guyot-infomaniak, line 66: Is it correct that the scope of this variable is a line of the CSV text file? If yes, then the declaration can be moved into the `while` loop below.

Should we reset this variable to `0` every time we encounter the `,` character? An. end of line?

luc-guyot-infomaniak (commented):
Juste a question about the scope of `doubleQuoteCounter`.

src/libsyncengine/jobs/network/csvfullfilelistwithcursorjob.cpp
luc-guyot-infomaniak, line 66:
```suggestion
    uint doubleQuoteCount = 0;
```

Simply because it is shorter to type!

ClementKunz: Yes it is but this function `getItem` called on each line. The while loop is here to handle special cases where the file name contains a line return.
In this case, we can imagine a file name that contains both double quote and line return. Therefor, we should ignore it anyway.

ClementKunz: Ok, done 👍

ChristopheLarchier (approved):

PR #115, opened 2024-05-13
Kdesktop 849 add tests and documentation for get rights

herve-er:
From Microsoft documentation:
> https://learn.microsoft.com/en-us/windows/win32/api/aclapi/nf-aclapi-geteffectiverightsfromacla
> 
> The GetEffectiveRightsFromAcl function fails and returns ERROR_INVALID_ACL if the specified ACL contains an inherited access-denied ACE.

From my personal test on Windows 11 Pro 23H2, the function does work as expected and returns the rights for the trustee even if the ACL contains an inherited access-denied ACE.

If we get ERROR_INVALID_ACL, we will consider to be in the case described in the documentation and consider the file as not existing (we can't get the rights).

This PR also add tests for inherited permission and for files/folders on without inherited permission. 

To test inheritance, I introduced `IoHelper::_setRightsWindowsApiInheritance`, which is false by default.

The `_setRightsWindowsApi` then uses this variable to determine whether it should set permissions on sub-objects. 
This is necessary to test the `getRights` method on files/folders that inherit permissions from the parent folder.
This method avoids the need to add a parameter to the `getRights` function, which is generic across Linux, MacOS, and Windows.

### DIFF
--- a/src/libcommonserver/io/iohelper_win.cpp
+++ b/src/libcommonserver/io/iohelper_win.cpp
@@ -454,7 +454,13 @@ static bool getRightsWindowsApi(const SyncPath &path, bool &read, bool &write, b
     ACCESS_MASK rights = 0;
     result = GetEffectiveRightsFromAcl(pfileACL, &Utility::_trustee, &rights);
     ioError = dWordError2ioError(result);
-    if (result == ERROR_INVALID_ACL) { // The GetEffectiveRightsFromAcl function fails and returns ERROR_INVALID_ACL if the specified ACL contains an inherited access-denied ACE.
+
+    /* The GetEffectiveRightsFromAcl function fails and returns ERROR_INVALID_ACL if the specified ACL contains an inherited access-denied ACE.
+     * see: https://learn.microsoft.com/en-us/windows/win32/api/aclapi/nf-aclapi-geteffectiverightsfromacla
+     * From my personal test on Windows 11 Pro 23H2, the function does work as expected and returns the rights for the trustee even if the ACL contains an inherited access-denied ACE.
+     * If we get ERROR_INVALID_ACL, we will consider to be in the case described in the documentation and consider the file as not existing (we can't get the rights).
+     */
+    if (result == ERROR_INVALID_ACL) { 
         LOGW_INFO(logger, L"getRightsWindowsApi: path='" << Utility::formatSyncPath(path) << L"', the specified ACL contains an inherited access - denied ACE. Considerring the file as not existing.");
         read = false;
         write = false;
--- a/src/libcommonserver/io/iohelper_win.cpp
+++ b/src/libcommonserver/io/iohelper_win.cpp
@@ -93,6 +93,7 @@ time_t FileTimeToUnixTime(LARGE_INTEGER filetime, DWORD *remainder) {
 }  // namespace
 
 int IoHelper::_getAndSetRightsMethod = 0;
+bool IoHelper::_setRightsWindowsApiInheritance = false;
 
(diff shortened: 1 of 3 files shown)


src/libcommonserver/io/iohelper_win.cpp
```diff
@@ -334,7 +335,8 @@ bool IoHelper::checkIfFileIsDehydrated(const SyncPath &itemPath, bool &isDehydra
 }
 
 static bool setRightsWindowsApi(const SyncPath &path, DWORD permission, ACCESS_MODE accessMode, IoError &ioError,
-                                log4cplus::Logger logger) noexcept {  // Always return false if ioError != IoErrorSuccess, caller
+                                log4cplus::Logger logger,
+                                bool inherite = false) noexcept {             // Always return false if ioError != IoErrorSuccess, caller
```
ClementKunz, line 339: The comment is written in a strange way on 2 lines. Put it on the line above the definition since it is a description of the function behavior.

herve-er: CLang split the comment on two lines as it was too long. 
I moved it above the definition as suggested (commit: https://github.com/Infomaniak/desktop-kDrive/pull/115/commits/032300bac6dbd7e4fe32c6d68e7aa5176dc6a044), this solved the problem.

src/libcommonserver/io/iohelper_win.cpp
```diff
@@ -541,10 +553,12 @@ bool IoHelper::setRights(const SyncPath &path, bool read, bool write, bool exec,
         } else {
             grantedPermission |= FILE_GENERIC_EXECUTE;
         }
-        bool res =
-            setRightsWindowsApi(path, grantedPermission, ACCESS_MODE::SET_ACCESS, ioError, logger()) || _isExpectedError(ioError);
+        bool res = setRightsWindowsApi(path, grantedPermission, ACCESS_MODE::SET_ACCESS, ioError, logger(),
+                                       _setRightsWindowsApiInheritance) ||
```
ClementKunz, line 561: Strange line return, hard to read

src/libcommonserver/io/iohelper_win.cpp
```diff
@@ -541,10 +553,12 @@ bool IoHelper::setRights(const SyncPath &path, bool read, bool write, bool exec,
...
+                   _isExpectedError(ioError);
         if (res) {
-            res &= setRightsWindowsApi(path, deniedPermission, ACCESS_MODE::DENY_ACCESS, ioError, logger()) ||
+            res &= setRightsWindowsApi(path, deniedPermission, ACCESS_MODE::DENY_ACCESS, ioError, logger(),
+                                       _setRightsWindowsApiInheritance) ||
```
ClementKunz, line 565: Same

src/libcommonserver/io/iohelper_win.cpp
herve-er: As discussed, we may want to change the clang format behaviour in this case.
https://github.com/Infomaniak/desktop-kDrive/pull/115/commits/262fff3cd330c7cc3cc36e8d01b7f60924439b38

src/libcommonserver/io/iohelper_win.cpp
herve-er: As discussed, we may want to change the clang format behaviour in this case.
https://github.com/Infomaniak/desktop-kDrive/pull/115/commits/262fff3cd330c7cc3cc36e8d01b7f60924439b38

ClementKunz (approved):

PR #116, opened 2024-05-13
KDESKTOP-823-Emptying-the-trash-from-webapp-generate-a-lot-of-unneccessary-log

ClementKunz:
Actions "file_delete" are now ignored because it is triggered only on files already in the trash


src/libsyncengine/jobs/network/getfileinfojob.cpp
```diff
@@ -35,39 +35,47 @@ GetFileInfoJob::GetFileInfoJob(int driveDbId, const NodeId &nodeId)
...
+        return false;
+    }
+    if (!JsonParserUtility::extractValue(dataObj, lastModifiedAtKey, _modtime)) {
+        return false;
+    }
+    std::string tmp;
+    if (!JsonParserUtility::extractValue(dataObj, typeKey, tmp)) {
+        return false;
+    }
+    bool isDir = tmp == dirKey;
```
luc-guyot-infomaniak, line 58:
```suggestion
   const bool isDir = tmp == dirKey;
```

src/libsyncengine/update_detection/file_system_observer/snapshot/snapshot.cpp
```diff
@@ -84,6 +85,12 @@ bool Snapshot::updateItem(const SnapshotItem &newItem) {
         startUpdate();
     }
 
+    if (ParametersCache::instance()->parameters().extendedLog()) {
+        LOGW_DEBUG(Log::instance()->getLogger(), L"File/directory: " << SyncName2WStr(newItem.name()).c_str() << L" ("
```
luc-guyot-infomaniak, line 89:
```suggestion
        LOGW_DEBUG(Log::instance()->getLogger(), L"Item: " << SyncName2WStr(newItem.name()).c_str() << L" ("
```

src/libsyncengine/update_detection/file_system_observer/snapshot/snapshot.cpp
```diff
@@ -84,6 +85,12 @@ bool Snapshot::updateItem(const SnapshotItem &newItem) {
         startUpdate();
     }
 
+    if (ParametersCache::instance()->parameters().extendedLog()) {
```
luc-guyot-infomaniak, line 88: Out of scope: It would be nice to have s shorthand for this call, for instance: `ParametersCache::isLogExtended()`.

src/libsyncengine/update_detection/file_system_observer/remotefilesystemobserverworker.cpp
```diff
@@ -390,14 +392,16 @@ ExitCode RemoteFileSystemObserverWorker::getItemsInDir(const NodeId &dirId, cons
     auto nodeIdIt = nodeIds.begin();
     while (nodeIdIt != nodeIds.end()) {
         if (_snapshot->isOrphan(*nodeIdIt)) {
+            LOG_SYNCPAL_DEBUG(_logger, L"Node " << SyncName2WStr(_snapshot->name(*nodeIdIt)).c_str() << L" ("
```
luc-guyot-infomaniak, line 395:
```suggestion
            LOG_SYNCPAL_DEBUG(_logger, L"Node '" << SyncName2WStr(_snapshot->name(*nodeIdIt)).c_str() << L"' ("
```

src/libsyncengine/update_detection/file_system_observer/remotefilesystemobserverworker.cpp
```diff
@@ -474,225 +478,241 @@ ExitCode RemoteFileSystemObserverWorker::sendLongPoll(bool &changes) {
...
-                                                           SyncTime &createdAt, SyncTime &modtime) {
-    if (action == accessRightInsert || action == accessRightUpdate || action == accessRightUserInsert ||
-        action == accessRightUserUpdate || action == accessRightTeamInsert || action == accessRightTeamUpdate ||
-        action == accessRightMainUsersInsert || action == accessRightMainUsersUpdate) {
-        bool hasRights = false;
-        ExitCode exitCode = hasAccessRights(nodeId, hasRights, createdAt, modtime);
-        rightsAdded = hasRights;
-        return exitCode;
+        default:
+            LOGW_SYNCPAL_DEBUG(_logger, L"Unknown operation received on file: " << SyncName2WStr(actionInfo.name).c_str() << L" ("
```
luc-guyot-infomaniak, line 702:
```suggestion
            LOGW_SYNCPAL_DEBUG(_logger, L"Unknown operation received on item: " << SyncName2WStr(actionInfo.name).c_str() << L" ("
```

src/libsyncengine/update_detection/file_system_observer/remotefilesystemobserverworker.cpp
```diff
@@ -474,225 +478,241 @@ ExitCode RemoteFileSystemObserverWorker::sendLongPoll(bool &changes) {
...
-                // Ignore these actions
-            } else {
-                LOGW_SYNCPAL_DEBUG(_logger, L"Unknown operation received on file: " << SyncName2WStr(name).c_str() << L" ("
-                                                                                    << Utility::s2ws(id).c_str() << L")");
+        // Item removed
+        case ActionCode::actionCodeAccessRightRemove:
+        case ActionCode::actionCodeAccessRightUserRemove:
+        case ActionCode::actionCodeAccessRightTeamRemove:
+        case ActionCode::actionCodeAccessRightMainUsersRemove: {
+            bool hasRights = false;
```
luc-guyot-infomaniak, line 663: This case looks like the `ActionCode::actionCodeAccessRightMainUsersUpdate` case. Should they be handled the same way? If not, could we avoid code duplication in some way?

src/libsyncengine/update_detection/file_system_observer/snapshot/snapshot.cpp
ClementKunz: Agreed. Task created in backlog: https://infomaniak.atlassian.net/browse/KDESKTOP-864

src/libsyncengine/update_detection/file_system_observer/remotefilesystemobserverworker.cpp
ClementKunz: The differences are on those lines:
```if (!hasRights) break;  // Current user does not have the right to access this item, ignore action.```
```if (hasRights) break;  // Current user still have the right to access this item, ignore action.```

And also those cases does not break but fall through the next statement, which is why they must be handled separately.
But I agree that there is some duplication here. Any suggestions are welcome ^^ I did not find a better/clearer way yet

Comment (luc-guyot-infomaniak) [reply src/libsyncengine/update_detection/file_system_observer/remotefilesystemobserverworker.cpp:663]: I would use a lambda function of the following form: 

```
const auto hasRights = [this](NodeId &nodeId) -> bool {
            SyncTime createdAt = 0;
            SyncTime modtime = 0;
            int64_t size = 0;
            if (getFileInfo(nodeId, hasRights, createdAt, modtime, size) != ExitCodeOk) {
               ...
}
```

and then call twice

`if (hasRights(actionInfo.nodeId)) ...`

with different side effects (break of fall through).

**Review by ChristopheLarchier — APPROVED** [review]

File: src/libsyncengine/update_detection/file_system_observer/remotefilesystemobserverworker.cpp (hunk 6 above)
Comment (ClementKunz) [reply src/libsyncengine/update_detection/file_system_observer/remotefilesystemobserverworker.cpp]: Done 👍

File: src/libsyncengine/update_detection/file_system_observer/remotefilesystemobserverworker.cpp [hunk 7]
```diff
@@ -474,225 +478,247 @@ ExitCode RemoteFileSystemObserverWorker::sendLongPoll(bool &changes) {
...
+        case ActionCode::actionCodeAccessRightInsert:
+        case ActionCode::actionCodeAccessRightUpdate:
+        case ActionCode::actionCodeAccessRightUserInsert:
+        case ActionCode::actionCodeAccessRightUserUpdate:
+        case ActionCode::actionCodeAccessRightTeamInsert:
+        case ActionCode::actionCodeAccessRightTeamUpdate:
+        case ActionCode::actionCodeAccessRightMainUsersInsert:
+        case ActionCode::actionCodeAccessRightMainUsersUpdate: {
+            bool rightsOk = false;
+            SyncTime createdAt = 0;
```
luc-guyot-infomaniak, line 620: These 3 local variables are not used anywhere. I would recommend to isolate them in the body of the lambda function.

ClementKunz: Done 👍

PR #121, opened 2024-05-14
KDESKTOP-857-Conflict-resolution-window-Wrong-color-of-text-in-dark-theme

ClementKunz:
Fix color in dark mode

### DIFF
--- a/src/gui/parametersdialog.cpp
+++ b/src/gui/parametersdialog.cpp
@@ -66,22 +66,7 @@ static const int maxLogFilesToSend = 25;
 
 Q_LOGGING_CATEGORY(lcParametersDialog, "gui.parametersdialog", QtInfoMsg)
 
-ParametersDialog::ParametersDialog(std::shared_ptr<ClientGui> gui, QWidget *parent)
-    : CustomDialog(false, parent),
-      _gui(gui),
-      _currentDriveDbId(0),
-      _backgroundMainColor(QColor()),
-      _pageStackedWidget(nullptr),
-      _driveMenuBarWidget(nullptr),
-      _preferencesMenuBarWidget(nullptr),
-      _errorsMenuBarWidget(nullptr),
-      _preferencesWidget(nullptr),
-      _drivePreferencesWidget(nullptr),
-      _drivePreferencesScrollArea(nullptr),
-      _noDrivePagewidget(nullptr),
-      _sendLogsWidget(nullptr),
-      _errorsStackedWidget(nullptr),
-      _defaultTextLabel(nullptr) {
+ParametersDialog::ParametersDialog(std::shared_ptr<ClientGui> gui, QWidget *parent) : CustomDialog(false, parent), _gui(gui) {
     initUI();
 
     connect(this, &ParametersDialog::exit, this, &ParametersDialog::onExit);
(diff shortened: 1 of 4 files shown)


src/gui/abstractfileitemwidget.cpp
```diff
@@ -38,10 +38,13 @@ static const int boxVSpacing = 10;
 static const int shadowBlurRadius = 20;
 static const QSize iconSize = QSize(15, 15);
 
+static const QColor folderIconColorLight = QColor("#666666");
```
ChristopheLarchier, line 41: Not possible to use QSS instead ?

ClementKunz: I did not find a solution with qss because of call to method `KDC::GuiUtility::getIconWithColor`:
`QColor color = GuiUtility::isDarkTheme() ? folderIconColorDark : folderIconColorLight;
    _driveIconLabel->setPixmap(
        KDC::GuiUtility::getIconWithColor(":/client/resources/icons/actions/icon-folder-empty.svg", color).pixmap(iconSize));`

luc-guyot-infomaniak, line 41:
```suggestion
static const QColor folderIconColorLight = QColor(0x666666);
```

`Qt` Creator recommends this int-based constructor because it is faster. It doesn't matter in this case, but it would demonstrate the principle and serves as an example for other constructor calls.

src/gui/abstractfileitemwidget.cpp
```diff
@@ -177,18 +183,22 @@ void AbstractFileItemWidget::setFileTypeIcon(const QString &ressourcePath) {
...
 }
 
 void AbstractFileItemWidget::setPath(const QString &path) {
+    QColor color = GuiUtility::isDarkTheme() ? folderIconColorDark : folderIconColorLight;
     _driveIconLabel->setPixmap(
-        KDC::GuiUtility::getIconWithColor(":/client/resources/icons/actions/icon-folder-empty.svg").pixmap(iconSize));
+        KDC::GuiUtility::getIconWithColor(":/client/resources/icons/actions/icon-folder-empty.svg", color).pixmap(iconSize));
 
-    QString printablePath = "/" + path;
+    QFileInfo fInfo(path);
```
luc-guyot-infomaniak, line 194:
```suggestion
    const QFileInfo fInfo(path);
```

src/gui/abstractfileitemwidget.cpp
```diff
@@ -177,18 +183,22 @@ void AbstractFileItemWidget::setFileTypeIcon(const QString &ressourcePath) {
...
+    QString printablePath;
+    if (!fInfo.isAbsolute()) printablePath = "/";
+    printablePath += path;
```
luc-guyot-infomaniak, line 197: I believe there are multiple places in the code where we build a `printablePath` (look for the string `+ ...`). It would be nice to have helper function that would handle them all.

src/gui/errortabwidget.cpp
```diff
@@ -58,16 +58,15 @@ ErrorTabWidget::ErrorTabWidget(int driveDbId, bool generic, QWidget *parent)
 
     _resolveButton = new QPushButton(this);
     _resolveButton->setObjectName("resolveButton");
-    _resolveButton->setText(tr("Resolve"));
     _resolveButton->setVisible(false);
     MenuWidgetLite *menu = new MenuWidgetLite(_resolveButton, this);
 
-    _resolveConflictsAction = new QAction(tr("Conflicted item(s)"), this);
+    _resolveConflictsAction = new QAction(this);
```
luc-guyot-infomaniak, line 64:
```suggestion
    _resolveConflictsAction = new QAction();
```

src/gui/errortabwidget.cpp
```diff
@@ -58,16 +58,15 @@ ErrorTabWidget::ErrorTabWidget(int driveDbId, bool generic, QWidget *parent)
...
     _resolveConflictsAction->setVisible(false);
     menu->addAction(_resolveConflictsAction);
     connect(_resolveConflictsAction, &QAction::triggered, this, &ErrorTabWidget::onResolveConflictErrors);
 
-    _resolveUnsupportedCharactersAction = new QAction(tr("Item(s) with unsupported characters"), this);
+    _resolveUnsupportedCharactersAction = new QAction(this);
```
luc-guyot-infomaniak, line 69:
```suggestion
    _resolveUnsupportedCharactersAction = new QAction();
```

src/gui/abstractfileitemwidget.cpp
```diff
@@ -97,7 +105,7 @@ AbstractFileItemWidget::AbstractFileItemWidget(QWidget *parent /*= nullptr*/) :
     mainLayout->addLayout(_bottomLayout);
 
     // Shadow
-    QGraphicsDropShadowEffect *effect = new QGraphicsDropShadowEffect(this);
+    auto effect = new QGraphicsDropShadowEffect(this);
```
luc-guyot-infomaniak, line 108:
```suggestion
    auto effect = new QGraphicsDropShadowEffect();
```

src/gui/abstractfileitemwidget.cpp
```diff
@@ -38,10 +38,13 @@ static const int boxVSpacing = 10;
...
 static const QSize iconSize = QSize(15, 15);
 
+static const QColor folderIconColorLight = QColor("#666666");
+static const QColor folderIconColorDark = QColor("#C0C0C0");
+
 AbstractFileItemWidget::AbstractFileItemWidget(QWidget *parent /*= nullptr*/) : QWidget(parent) {
     setContentsMargins(hMargin, vMargin, hMargin, vMargin);
 
-    QVBoxLayout *mainLayout = new QVBoxLayout(this);
+    auto mainLayout = new QVBoxLayout(this);
```
luc-guyot-infomaniak, line 47:
```suggestion
    auto mainLayout = new QVBoxLayout();
```

src/gui/genericerroritemwidget.cpp
```diff
@@ -68,7 +68,7 @@ void GenericErrorItemWidget::init() {
     }
 
     // Right layout
-    QLabel *fileDateLabel = new QLabel(this);
+    auto fileDateLabel = new QLabel(this);
```
luc-guyot-infomaniak, line 71:
```suggestion
    auto fileDateLabel = new QLabel();
```

src/gui/fixconflictingfilesdialog.cpp
```diff
@@ -209,9 +211,9 @@ QString FixConflictingFilesDialog::descriptionText() const {
 void FixConflictingFilesDialog::insertFileItems(const int nbItems) {
     uint64_t max = std::min(_fileListWidget->count() + nbItems, (int)_conflictList.size());
     for (uint64_t i = _fileListWidget->count(); i < max; i++) {
-        FileItemWidget *w = new FileItemWidget(_conflictList[i].destinationPath(), _conflictList[i].nodeType(), this);
+        auto w = new FileItemWidget(_conflictList[i].destinationPath(), _conflictList[i].nodeType(), this);
```
luc-guyot-infomaniak, line 214:
```suggestion
        auto w = new FileItemWidget(_conflictList[i].destinationPath(), _conflictList[i].nodeType());
```

src/gui/errortabwidget.cpp
ClementKunz: Why? I don't think it is a good idea. Providing a parent to a QWidget ensure that the child will be automatically deleted when the parent is deleted.
`The ownership of action is not transferred to this [QWidget](https://doc.qt.io/qt-6/qwidget.html)` (https://doc.qt.io/qt-6/qwidget.html#addAction)

src/gui/errortabwidget.cpp
ClementKunz: same

src/gui/abstractfileitemwidget.cpp
ClementKunz: same

src/gui/abstractfileitemwidget.cpp
ClementKunz: same

src/gui/genericerroritemwidget.cpp
ClementKunz: same

src/gui/fixconflictingfilesdialog.cpp
ClementKunz: Same, but there may be a memory leak because of the instantation of the `QListWidgetItem`. Should be fixed now

src/gui/errortabwidget.cpp
luc-guyot-infomaniak: You are absolutely right, thanks for the doc check: it is indeed true that ownership is not transferred to `menu` through `addAction`, so that the actions need a parent at the time they are created. My mistake.

In the context of my other comments (actions excepted), the widget `this` is often not the actual parent of the instantiated widget or layout. The parent is usually set later by some `Qt` setter. So,  it often happens that we set first an incorrect parent and then we set the correct one. I would prefer to skip the first step when it is sensible.

src/gui/abstractfileitemwidget.cpp
luc-guyot-infomaniak: No, it's not. See the `setLayout` documentation: https://doc.qt.io/qt-6/qwidget.html#setLayout.

src/gui/abstractfileitemwidget.cpp
ChristopheLarchier: What is done elsewhere is to use a Q_PROPERTY.
See AddDriveSmartSyncWidget:
Q_PROPERTY(QColor logo_color READ logoColor WRITE setLogoColor)

src/gui/errortabwidget.cpp
ClementKunz: Yes, you are totally right, it is the case at least for all `QLayout`. I'll fix it, thx.

src/gui/abstractfileitemwidget.cpp
ClementKunz: Done

src/gui/errortabwidget.cpp
ClementKunz: Done

src/gui/abstractfileitemwidget.cpp
ClementKunz: Done

luc-guyot-infomaniak (approved):

PR #127, opened 2024-05-21
[KDESKTOP-871] Local drive deletion followed by synchronisation deletion causes a crash in debug

luc-guyot-infomaniak:
This pull-request addresses the Jira ticket [DESKTOP-871](kdesktop-871-handle-local-drive-deletion-gracefully): a crash du to an assertion failure is fixed by adapting the assertion condition.

The fix consists only of a trivial one-line change, see comment attached to this request. 
All the other changes are improvements made on the way, so as to reduce code redundancy


src/server/appserver.cpp
```diff
@@ -411,7 +402,7 @@ void AppServer::stopSyncTask(int syncDbId) {
     ASSERT(_syncPalMap[syncDbId].use_count() == 1)
     _syncPalMap.erase(syncDbId);
 
-    ASSERT(_vfsMap[syncDbId].use_count() == 1)
+    ASSERT(_vfsMap[syncDbId].use_count() <= 1)  // `use_count` can be zero when the local drive has been removed.
```
luc-guyot-infomaniak, line 405: This prevents a crash in debug mode, under the conditions describe the Jira ticket: the local drive is removed and the user remove the corresponding synchronization too.

ChristopheLarchier (approved):

ClementKunz (approved):

PR #131, opened 2024-05-24
Kdesktop 885 add tests with and without env var for get set rights on windows

herve-er:


src/libcommonserver/io/iohelper.cpp
```diff
@@ -98,6 +98,9 @@ IoError IoHelper::posixError2ioError(int error) noexcept {
             return IoErrorInvalidArgument;
         case ENAMETOOLONG:
             return IoErrorFileNameTooLong;
+#ifdef _WIN32
+        case ESRCH:
```
herve-er, line 102: Strangely std::filesystem::permission return ec ESRCH instead of ENOENT when the file does not exist on Windows.

ClementKunz (approved):

PR #133, opened 2024-05-28
KDESKTOP-893-macOS-Sync-loop-if-there-is-already-a-file-.lnk-synced-locally

ClementKunz:
We should never generate operations on excluded file


src/libsyncengine/update_detection/file_system_observer/computefsoperationworker.cpp
```diff
@@ -239,7 +239,8 @@ ExitCode ComputeFSOperationWorker::exploreDbTree(std::unordered_set<NodeId> &loc
                 if (!snapshot->exists(nodeId) || movedIntoUnsyncedFolder) {
                     if (!pathInDeletedFolder(dbPath)) {
                         // Check that the file/directory really does not exist on replica
-                        ExitCode exitCode = checkIfOkToDelete(side, dbPath, nodeId);
+                        bool isExcluded = false;
+                        ExitCode exitCode = checkIfOkToDelete(side, dbPath, nodeId, isExcluded);
```
luc-guyot-infomaniak, line 243:
```suggestion
                        const ExitCode exitCode = checkIfOkToDelete(side, dbPath, nodeId, isExcluded);
```

luc-guyot-infomaniak:
I presume that this PR targets the `main` branch. Doesn't it?

ClementKunz:
> I presume that this PR targets the `main` branch. Doesn't it?

Not for now. If we need to provide a hot fix, we could simply cherry pick those commits

luc-guyot-infomaniak (approved):

PR #134, opened 2024-05-28
[KDESKTOP-895] Add Server tests

MaximePouce:


.github/workflows/macos.yml
```diff
@@ -29,18 +27,14 @@ jobs:
...
-        working-directory: ${{env.bin-directory}}
-        continue-on-error: true
+      - name: Grant tests script execute permission
+        run : chmod +x ./infomaniak-build-tools/run-tests.sh
+      - name: Execute tests
+        run : ./infomaniak-build-tools/run-tests.sh
 
       - name: Get the path to kDrive-logdir
         run : echo "logdir_path="$(find /private/var/folders/ -name kDrive-logdir 2>/dev/null)"" >> $GITHUB_ENV
+        if  : success() || failure()
```
luc-guyot-infomaniak, line 37: So, it runs in any case? Does it?

.github/workflows/windows.yml
```diff
@@ -44,13 +42,8 @@ jobs:
           powershell ./infomaniak-build-tools/windows/build-drive.ps1 -ci
         shell: cmd
 
-      - name: Execute common tests
-        run: ./kDrive_test_common.exe
-        working-directory: ${{env.bin-directory}}
-
-      - name: Execute parms tests
-        run: ./kDrive_test_parms.exe
-        working-directory: ${{env.bin-directory}}
```
luc-guyot-infomaniak, line 53: No need for granting execute permission?

.github/workflows/macos.yml
MaximePouce: It would run whether the build and tests succeeded or failed, but not if the workflow was cancelled

.github/workflows/windows.yml
MaximePouce: Not needed for Windows

PR #135, opened 2024-05-28
Fix init value for logUploadState

herve-er:
Fix init value for logUploadState, remove throw when fail to get logUploadStatus and add sentryEvent in case of failed of CommonUtility::stringToAppStateValue.

### DIFF
--- a/src/libcommon/utility/utility.cpp
+++ b/src/libcommon/utility/utility.cpp
@@ -235,28 +235,43 @@ QString CommonUtility::escape(const QString &in) {
 }
 
 bool CommonUtility::stringToAppStateValue(const std::string &stringFrom, AppStateValue &appStateValueTo) {
+    bool res = true;
+    std::string appStateValueType = "Unknown";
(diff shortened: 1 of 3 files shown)


src/libcommon/utility/utility.cpp
```diff
@@ -235,28 +235,43 @@ QString CommonUtility::escape(const QString &in) {
...
     } else if (std::holds_alternative<LogUploadState>(appStateValueTo)) {
+        appStateValueType = "LogUploadState";
         try {
             appStateValueTo = static_cast<LogUploadState>(std::stoi(stringFrom));
         } catch (const std::invalid_argument &) {
-            return false;
+            res = false;
         }
     } else if (std::holds_alternative<int64_t>(appStateValueTo)) {
+        appStateValueType = "int64_t";
```
ClementKunz, line 258: I would create a utility fonction that return the string (based of the std::variant index) instead of hardcoding those string here. For example see 'Utility::logLevel2Str'

ClementKunz (approved):

PR #136, opened 2024-05-28
[KDESKTOP-898] Fix CI Packaging

MaximePouce:


infomaniak-build-tools/macos/build-drive.sh
MaximePouce, line 1: This change is related to the CMakeLists change for MacOS, building the kDrive server app in the install directory instead of bin

infomaniak-build-tools/macos/build-drive.sh
```diff
@@ -97,7 +97,7 @@ cmake \
 make
 
 # Generate Debug files
-dsymutil ./bin/kDrive.app/Contents/MacOS/kDrive -o ../kDrive.dSYM
+dsymutil ./install/kDrive.app/Contents/MacOS/kDrive -o ../kDrive.dSYM
```
MaximePouce, line 100: This change is related to the CMakeLists change for MacOS, building the kDrive server app in the install directory instead of bin

ClementKunz (approved):

PR #155, opened 2024-06-05
KDESKTOP-930 Check for getRights method performance at startup and if…

herve-er:
# Summary

On Windows, we are experiencing significant slowdowns for some users due to the methods used to set/get rights via the Windows API on Windows Server. This issue stems from the API making network calls to the Active Directory manager with each request.

# Details

### Identified Issue
The slowdown is caused by the Windows API functions making network calls to the Active Directory manager.

### Performance Measurements
- **Windows 11 without Active Directory:** 1ms < response time < 4ms
- **Windows 11 with Active Directory:** 70ms < response time < 110ms

### Current Workaround
The current solution involves instructing clients to set an environment variable to bypass the recommended Windows API methods, instead using standard methods. However, the standard methods rely solely on the read-only flag, which often does not accurately reflect the actual rights applied to the file/folder. Despite this limitation, it remains the better option given the constraints.

### Proposed Improvement
This pull request aims to eliminate the need for manual environment variable configuration. The application will automatically select the most appropriate method for rights management, improving overall performance and user experience.


src/libcommonserver/utility/utility_win.cpp
```diff
@@ -112,15 +112,50 @@ static bool initTrusteeWithUserSID() {
 }
 
 static bool init_private() {
+    const std::string useGetRightsFallbackMethod = CommonUtility::envVarValue("KDRIVE_USE_GETRIGHTS_FALLBACK_METHOD");
+    if (!useGetRightsFallbackMethod.empty()) {
+        LOG_DEBUG(Log::instance()->getLogger(), "KDRIVE_USE_GETRIGHTS_FALLBACK_METHOD env is set, using fallback method");
+        IoHelper::_getAndSetRightsMethod = 1;
```
luc-guyot-infomaniak, line 118: It is surprising to see a private variable of the class `IoHelper` to be set in the body of a private function of `KDC::Utility`.

We can use instead a public setter or do this initialization within the `IoHelper` class.

src/libcommonserver/utility/utility_win.cpp
```diff
@@ -112,15 +112,50 @@ static bool initTrusteeWithUserSID() {
...
+    for (int i = 0; i < 10; i++) {
+        if (!IoHelper::getRights(tmpDir, read, write, execute, ioError)) {
+            LOGW_WARN(Log::instance()->getLogger(), "Error in IoHelper::getRights: " << Utility::formatIoError(tmpDir, ioError));
+            return false;
+        }
+    }
+    auto end = std::chrono::high_resolution_clock::now();
+    auto duration = std::chrono::duration_cast<std::chrono::milliseconds>(end - start).count();
+
+    /* Averrage times for 10 calls to getRights with windows API:
```
luc-guyot-infomaniak, line 150:
```suggestion
    /* Average time spent in getRights with windows API:
```
and then divide the subsequent time estimates by 10.

src/libcommonserver/utility/utility_win.cpp
```diff
@@ -112,15 +112,50 @@ static bool initTrusteeWithUserSID() {
...
+     *    -> Windows 11 without Active Directory: 10ms < time < 40ms
+     *    -> Windows 11 with Active Directory: 700ms < time < 1100ms
+     */
+    LOG_DEBUG(Log::instance()->getLogger(), "Time to get rights x10: " << duration << "ms");
+
+    if (duration > 60) {
```
ClementKunz, line 156: Is 60ms a sufficiently high threshold? Did you have the chance to check this value on other machine?
Also create a static variable somewhere with an explicit name

src/libcommonserver/utility/utility_win.cpp
herve-er: I moved the initialization from utility to IoHelper. it was an artefacts from before the creation of the IoHelper class.

herve-er: [43d8601](https://github.com/Infomaniak/desktop-kDrive/pull/155/commits/43d8601f32d5a94a22f1da7e4fcab280de83f82a)

src/libcommonserver/utility/utility_win.cpp
herve-er: Done 👍🏾 ([43d8601](https://github.com/Infomaniak/desktop-kDrive/pull/155/commits/43d8601f32d5a94a22f1da7e4fcab280de83f82a))

ClementKunz (approved):

PR #159, opened 2024-06-06
Add archiving of old log at startup and then check every 1h

herve-er:
This PR aim to fix a bug in the current app where server log are not archived.


src/libcommonserver/io/iohelper.cpp
```diff
@@ -514,6 +514,20 @@ bool IoHelper::tempDirectoryPath(SyncPath &directoryPath, IoError &ioError) noex
 }
 
 bool IoHelper::logDirectoryPath(SyncPath &directoryPath, IoError &ioError) noexcept {
+    //!\ Don't use IoHelper::logger() here, as it Log::instance() may not be initialized yet. /!\_
```
ClementKunz, line 517:
```suggestion
    //!\ Don't use IoHelper::logger() here, as its Log::instance() may not be initialized yet. /!\
```

src/libcommonserver/log/customrollingfileappender.cpp
```diff
@@ -286,30 +288,45 @@ void CustomRollingFileAppender::rollover(bool alreadyLocked) {
...
-        IoError ioError = IoErrorSuccess;
-        SyncPath logDirPath;
-        if (!IoHelper::logDirectoryPath(logDirPath, ioError)) {
-            throw std::runtime_error("Error in CustomRollingFileAppender: failed to get the log directory path.");
+    _lastExpireCheck = std::chrono::system_clock::now();
+    // Archive Previous Log Files and delete expired files
+    IoError ioError = IoErrorSuccess;
+    SyncPath logDirPath;
+    if (!IoHelper::logDirectoryPath(logDirPath, ioError) || ioError != IoErrorSuccess) {
+        throw std::runtime_error("Error in CustomRollingFileAppender: failed to get the log directory path.");
```
ClementKunz, line 296: Is this properly catched? If not, do we really the app to crash in such a case?

test/libcommonserver/log/testlog.cpp
```diff
@@ -51,4 +47,78 @@ void TestLog::testLog() {
...
+    DirectoryEntry entry;
+    while (dirIt.next(entry, endOfDirectory, ioError) && !endOfDirectory && ioError == IoErrorSuccess) {
+        if (entry.path().filename().string() == Log::instance()->getLogFilePath().filename().string()) {
+            continue;
+        }
+        IoHelper::deleteDirectory(entry.path(), ioError);
+    }
+
+    testLargeLogRolling();
+    testExpiredLogFiles();
```
ClementKunz, line 66: Wouldn't it better to run those tests separetly?
And therefor, moving the clearing of the log directory in a dedicated function to be called at the begining of each test

src/libcommonserver/io/iohelper_win.cpp
```diff
@@ -68,7 +68,9 @@ IoError dWordError2ioError(DWORD error) noexcept {
         case ERROR_INVALID_NAME:
             return IoErrorNoSuchFileOrDirectory;
         default:
-            LOG_WARN(Log::instance()->getLogger(), "Unknown IO error - error=" << error);
+            if (Log::isSet()) {
+                LOG_WARN(Log::instance()->getLogger(), "Unknown IO error - error=" << error);
```
luc-guyot-infomaniak, line 72:
```suggestion
                LOG_WARN(Log::instance()->getLogger(), "Unhandled DWORD error - error=" << error);
```

src/libcommonserver/log/customrollingfileappender.cpp
```diff
@@ -286,30 +288,45 @@ void CustomRollingFileAppender::rollover(bool alreadyLocked) {
...
-    if (_expire > 0) {
-        _lastExpireCheck = std::chrono::system_clock::now();
-        QDateTime now = QDateTime::currentDateTime();
-
-        IoError ioError = IoErrorSuccess;
-        SyncPath logDirPath;
-        if (!IoHelper::logDirectoryPath(logDirPath, ioError)) {
-            throw std::runtime_error("Error in CustomRollingFileAppender: failed to get the log directory path.");
+    _lastExpireCheck = std::chrono::system_clock::now();
+    // Archive Previous Log Files and delete expired files
```
luc-guyot-infomaniak, line 292:
```suggestion
    // Archive previous log files and delete expired files
```

src/libcommonserver/log/customrollingfileappender.cpp
```diff
@@ -286,30 +288,45 @@ void CustomRollingFileAppender::rollover(bool alreadyLocked) {
...
-        QDir logDir(SyncName2QStr(logDirPath.native()));
-        QStringList filters;
-        filters << SyncName2QStr(Str("*") + Utility::logFileName() + Str("*"));
-        QStringList logFileNameList = logDir.entryList(filters, QDir::Files);
-        for (const QString &logFileName : logFileNameList) {
-            QFileInfo fileInfo(logDir, logFileName);
-            QDateTime expireDateTime = fileInfo.lastModified().addSecs(_expire * 3600);
+        // Delete expired files
+        if (_expire > 0 && entry.path().string().find(APPLICATION_NAME) != std::string::npos) {
+            auto now = std::chrono::system_clock::now();
```
luc-guyot-infomaniak, line 310:
```suggestion
            const auto now = std::chrono::system_clock::now();
```
Same comment for the next two variables.

src/libcommonserver/log/customrollingfileappender.cpp
```diff
@@ -286,30 +288,45 @@ void CustomRollingFileAppender::rollover(bool alreadyLocked) {
...
+            auto expireDateTime = lastModified + std::chrono::seconds(_expire);
             if (expireDateTime < now) {
-                log4cplus::file_remove(fileInfo.filePath().toStdWString());
+                log4cplus::file_remove(Utility::s2ws(entry.path().string()));
                 continue;
             }
         }
+
+        // Compress previous log sessions
+        if (SyncPath currentLogName = DirectoryEntry(filename).path().filename().replace_extension("");
```
luc-guyot-infomaniak, line 320:
```suggestion
        if (const SyncPath currentLogName = DirectoryEntry(filename).path().filename().replace_extension("");
```

src/libcommonserver/log/customrollingfileappender.cpp
```diff
@@ -286,30 +288,45 @@ void CustomRollingFileAppender::rollover(bool alreadyLocked) {
...
+            entry.path().filename().string().find(".gz") == std::string::npos &&
+            entry.path().string().find(currentLogName.string()) == std::string::npos) {
+            if (CommonUtility::compressFile(QString::fromStdString(entry.path().string()),
```
luc-guyot-infomaniak, line 323: Suggestion: overload `CommonUtility::compressFile` so that it can use `std::string`s as arguments.

test/libcommonserver/io/testio.cpp
```diff
@@ -134,8 +134,10 @@ void TestIo::testLogDirectoryPath() {
             ec = std::make_error_code(std::errc::not_enough_memory);
             return SyncPath{};
         });
-
+        auto logInstance = Log::_instance;
+        Log::_instance.reset();
         CPPUNIT_ASSERT(!_testObj->logDirectoryPath(logDirPath, ioError));
+        Log::_instance = logInstance;
```
luc-guyot-infomaniak, line 140: It would be nice if you could explain why this tweak is required.

src/libcommonserver/io/iohelper.cpp
herve-er: The line can't end with a backslash, it is why i put the underscore at the end.

herve-er: [57fffe2](https://github.com/Infomaniak/desktop-kDrive/pull/159/commits/57fffe2c590b05105eb666ffda35c14fc6e3c640)

src/libcommonserver/log/customrollingfileappender.cpp
herve-er: [2f69544](https://github.com/Infomaniak/desktop-kDrive/pull/159/commits/2f69544f6a274ab3994e29ce9a1e9d287e15910d)

test/libcommonserver/log/testlog.cpp
```diff
@@ -51,4 +47,78 @@ void TestLog::testLog() {
...
+
+void TestLog::testLargeLogRolling(void) {
+    CPPUNIT_ASSERT_EQUAL(1, countFilesInDirectory(_logDir));
+    // Generate a log larger than the max log file size.
+    std::string testLog = "Test info log/";
+    for (int i = 0; i < 1000; i++) {
+        testLog += "Test info log/";
+    }
+    uint64_t currentSize = 0;
+    std::cout << std::endl
```
luc-guyot-infomaniak, line 77: I am afraid that these debug print statements are going to clutter the unit tests output.

test/libcommonserver/log/testlog.cpp
```diff
@@ -51,4 +47,78 @@ void TestLog::testLog() {
...
+    Utility::msleep(4000);  // Wait for the first log file to expire
+    static_cast<CustomRollingFileAppender*>(rfAppenderPtr.get())
+        ->setExpire(5);                   // Force the check of expired files at the next log
+    LOG_DEBUG(_logger, "Log to trigger the appender");  // Generate a log to trigger the appender
+    CPPUNIT_ASSERT_EQUAL(1, countFilesInDirectory(_logDir));
+}
+
+int TestLog::countFilesInDirectory(const SyncPath& directory) const {
+    bool endOfDirectory = false;
+    IoError ioErorr = IoErrorSuccess;
```
luc-guyot-infomaniak, line 111:
```suggestion
    IoError ioError = IoErrorSuccess;
```

test/libcommonserver/log/testlog.cpp
herve-er: image
As the process can be long 30s<t<60s (especially on windows) I think it can be a good think to have a visualization of the advancement?

test/libcommonserver/io/testio.cpp
herve-er:
```
As IoHelper::logDirectoryPath() use the path returned by Log::instance()->getLogFilePath().parent_path() if Log::_instance is not null,
we need to reset Log::_instance to ensure that we are in the default state where the function generates the log directory path from the temp directory path.
```
I added it as a doc string. ([c0c683a](https://github.com/Infomaniak/desktop-kDrive/pull/159/commits/c0c683a96c206345afa604c7fa658a30a597abd9))

test/libcommonserver/log/testlog.cpp
luc-guyot-infomaniak: We may discuss this as a desktop dev team topic. I am worried of the impact on the readability of the unit tests output. When errors happen in unit tests, we identify them more easily if the output is kept simple. Besides, a test that lasts up to one minute does not look like a unit test anymore.

herve-er: I understand your concern regarding readability. To mitigate this impact, I propose an alternative approach: [commit 09737d1](https://github.com/Infomaniak/desktop-kDrive/pull/159/commits/09737d15dc2d63b0df49cabac27ee750a5acb4ae).

image

This solution is less invasive and ensures that the output is completely removed at the end of the test, maintaining the original readability state.

image

Could this approach be a good comromise?

herve-er: The unit test itself runs instantaneously. However, the preparation phase is time-consuming because it requires "spamming" with `LOG_DEBUG` calls until the maximum logfile size is reached. This process is relatively quick on Linux and macOS, but it becomes significantly more time-consuming on Windows.

luc-guyot-infomaniak: Regarding your proposal for reducing the noise in the unit tests output: thumbs up! As long as the final output is kept short and readable, I won't complain.

Regarding, the issue with the unit test preparation: would it be possible to keep the iteration count as low as possible? Two long debug lines?

herve-er: After a quick test, it seems that after 12000 characters there isn't any significant improvement by increasing the size of the log output.

herve-er: (Currently each line is 14000 characters long)

luc-guyot-infomaniak (approved):

luc-guyot-infomaniak (approved):

PR #161, opened 2024-06-06
Fix session issues

ChristopheLarchier:


src/libsyncengine/jobs/network/abstractnetworkjob.cpp
```diff
@@ -328,15 +345,22 @@ bool AbstractNetworkJob::sendRequest(Poco::Net::HTTPSClientSession &session, con
     // Send request, retrieve an open stream
     std::vector<std::reference_wrapper<std::ostream>> stream;
     try {
-        stream.push_back(session.sendRequest(req));
-        if (ioOrLogicalErrorOccurred(stream[0].get())) {
-            int err = session.socket().getError();
-            return processSocketError(session, "sendRequest failed failed", jobId(), err, Poco::Error::getMessage(err));
+        const std::lock_guard<std::recursive_mutex> lock(_mutexSession);
```
luc-guyot-infomaniak, line 348: It would be nice if the block below (or a subset) could be hidden inside a re-usable method: this block seems to be repeated several times later on.

luc-guyot-infomaniak (commented):
Maybe it's possible to re-activate one more unit test after this wonderful finding?

src/libsyncengine/jobs/network/abstractnetworkjob.cpp
ChristopheLarchier: Done

ClementKunz (approved):

PR #164, opened 2024-06-11
KDESKTOP-940-Add-test-for-issue-938

ClementKunz:
Test added for the issue where the app is looping because of an already synchronized .lnk file on macOS


src/libsyncengine/requests/exclusiontemplatecache.cpp
```diff
@@ -145,26 +145,26 @@ ExitCode ExclusionTemplateCache::update(bool def, const std::vector<ExclusionTem
     return ExitCodeOk;
 }
 
-bool ExclusionTemplateCache::checkIfIsExcluded(const SyncPath &basePath, const SyncPath &relativePath, bool &isWarning,
```
luc-guyot-infomaniak, line 148: The original name is certainly both awful and painful to type. However, it remains closer to what the function actually does: it checks, among other things, whether a file is excluded; the `isExcluded` containing the specific exclusion information.

The issue with your proposal is that the caller may think that `isExcluded` returns the exclusion status of the file under scrutiny. It actually returns of the function call: `true` if the function was able to retrieve the desired information, false if some unexpected error was encountered.

test/libsyncengine/update_detection/file_system_observer/testcomputefsoperationworker.h
```diff
@@ -26,15 +26,32 @@ namespace KDC {
...
+        /**
+         * No operation should be generated in this test.
+         */
+        void testNoOps();
+        /**
+         * Multiple operations of different types should be generated in this test.
+         */
+        void testMultipleOps();
+        /**
+         * Spécific test for issue https://infomaniak.atlassian.net/browse/KDESKTOP-893.
```
luc-guyot-infomaniak, line 48:
```suggestion
         * Specific test for the issue https://infomaniak.atlassian.net/browse/KDESKTOP-893.
```

luc-guyot-infomaniak: Very cool that you mentioned the issue!

test/libsyncengine/update_detection/file_system_observer/testcomputefsoperationworker.h
```diff
@@ -26,15 +26,32 @@ namespace KDC {
...
+         * Sync is looping because a .lnk file was already synchronized with an earlier app's version.
```
luc-guyot-infomaniak, line 49:
```suggestion
         * Sync is looping because a `.lnk` file was already synchronized with an earlier app's version.
```

test/libsyncengine/update_detection/file_system_observer/testcomputefsoperationworker.h
```diff
@@ -26,15 +26,32 @@ namespace KDC {
...
+         * A delete FS operation was generated on remote replica but could not be propagated since the file still exist on remote
```
luc-guyot-infomaniak, line 50:
```suggestion
         * A Delete FS operation was generated on remote replica but could not be propagated since the file still exists on this 
```

src/libsyncengine/requests/exclusiontemplatecache.cpp
ClementKunz: Mmmh yes you are right. It was an attempt to uniformize the function names inside this class, especially because there was also a function named 'isEcludedTemplate'.
I will put back the 'checkIf...' part but see if it is possible to uniformize a bit and make it less "awful and painful" :)

ClementKunz: Done. Not perfect but we could refactor deeply later

luc-guyot-infomaniak (approved):

PR #173, opened 2024-06-14
Install shared libraries in the MacOSX release bundle

luc-guyot-infomaniak:


herve-er:
Merge without waiting for requirements to be met (bypass branch protections):  We currently encounter issues with sonar reporting. The code base is unchanged, sonar as nothing to analyse here.

PR #175, opened 2024-06-17
KDESKTOP-966-Bump-Xcode-Clang

herve-er:
# Summary

Some `std::` functions, such as `std::format`, are not available in the current version of libc++.

# Description

Certain standard functions from the C++ library (e.g., `std::format`) are missing in the current version of libc++ we are using. This issue is specific to MacOS and Ubuntu environments.

# Environment

## MacOS

The missing functions seem to be implemented starting from Xcode 15.3. More information can be found in the [C++ Language Support section on the Apple Developer site](https://developer.apple.com/documentation/xcode/c++-language-support).

## Ubuntu

These functions are supported starting from CLang 17. Reference: [Compiler support for C++20 on cppreference.com](https://en.cppreference.com/w/cpp/compiler_support).

## Windows

No changes are required.

# Proposed Solution

- Upgrade the development environment on MacOS to Xcode 15.3 or later.
- Upgrade CLang to version ~~17~~ 18 or later on Ubuntu.

# Additional Information

No changes are necessary for Windows environments.

### DIFF
--- a/infomaniak-build-tools/linux/Readme.md
+++ b/infomaniak-build-tools/linux/Readme.md
@@ -35,18 +35,37 @@ cd desktop-kDrive && git submodule update --init --recursive
 These requirements will only apply if you want to build from a Linux device  
 Currently, the build for Linux release is created from a podman container
 
-You will need cmake and gcc-c++ to compile the libraries and the kDrive project  
+You will need cmake and clang to compile the libraries and the kDrive project  
 This documentation was made for Ubuntu 22.04 LTS
 
 ## Packages :
 
-If not already installed, you will need **git**, **cmake** and **clang** packages.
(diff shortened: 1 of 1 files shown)


herve-er:
Linux CI is now running on CLang 18. 🚀

herve-er:
Clang has been updated to include support for nearly all C++20 features (almost all according to [C++ compiler support - cppreference.com](https://en.cppreference.com/w/cpp/compiler_support)  ).

On macOS, we currently use the latest version of Xcode. However, we're limited by macOS 10.15, which has been deprecated since September 12, 2022. Features like std::format and possibly other C++20 utilities require at least macOS 13.3. Since macOS 10.15 will be deprecated around November 1, 2024, it would be beneficial to update our supported macOS version at that time.

ClementKunz (approved):

PR #179, opened 2024-06-18
[KDESKTOP-969] Add unit tests to prevent a regression in UpdateTreeWorker::updateTmpNode

luc-guyot-infomaniak:


src/libsyncengine/update_detection/update_detector/updatetreeworker.cpp
```diff
@@ -152,38 +171,25 @@ ExitCode UpdateTreeWorker::step3DeleteDirectory() {
...
+            if (!parentNode) {
                 SyncPath newPath;
-                ExitCode exitCode = getNewPathAfterMove(deleteOp->path(), newPath);
-                if (exitCode != ExitCodeOk) {
-                    if (exitCode == ExitCodeDbError) {
-                        LOG_SYNCPAL_WARN(_logger, "Error in UpdateTreeWorker::getNewPathAfterMove");
-                        return exitCode;
-                    }
+                if (const auto newPathExitCode = getNewPathAfterMove(deleteOp->path(), newPath);
+                    newPathExitCode == ExitCodeDbError) {
```
ClementKunz, line 182: This is not directly related to your PR but what happen if the 'newPathExitCode' is an error different from 'ExitCodeDbError' ('ExitCodeDataError' for exemple)?

src/libsyncengine/update_detection/update_detector/updatetreeworker.cpp.orig
ClementKunz, line 1: What is this file? A backup?

test/libsyncengine/update_detection/update_detector/testupdatetreeworker.cpp
```diff
@@ -278,6 +278,80 @@ void TestUpdateTreeWorker::testUtilsFunctions() {
...
+void TestUpdateTreeWorker::testHandleCreateOperationsWithSamePath() {
+    setUpUpdateTree();
+
+    // Regular case: success
+    _operationSet->insertOp(std::make_shared<FSOperation>(OperationTypeDelete, "id4111", NodeTypeFile, 1654798667, 1654798667,
+                                                          12345, "Dir 4/Dir 4.1/Dir 4.1.1/File 4.1.1.1"));
+    _operationSet->insertOp(std::make_shared<FSOperation>(OperationTypeDelete, "id51", NodeTypeFile, 1654788552, 1654788552,
+                                                          12345, "Dir 5/File 5.1"));
+    _operationSet->insertOp(std::make_shared<FSOperation>(OperationTypeCreate, "id511", NodeTypeFile, 1654798336, 1654798336,
+                                                          12345, "Dir 5/File 5.1"));
```
ClementKunz, line 330: I think everywhere else in this file, 3 digit in the code means 3 level deep in the tree (a file with ID 'id511' should have a path such as "Dir 5/Dir 5.1/File 5.1.1"). Can we imagine a different ID in order to make it easier to read?

src/libsyncengine/update_detection/update_detector/updatetreeworker.cpp
luc-guyot-infomaniak: Thanks for noticing this. I am under the impression that it would be better to return immediately the error code if this code is not `ExitCodeOk.` This is what I implement with the next commit.

test/libsyncengine/update_detection/update_detector/testupdatetreeworker.cpp
luc-guyot-infomaniak: You are right, this is not consistent. It originates from a copy-paste, so there are other instances of this mismatch in the same file. I have made a proposal in the next commit to fix all such issues.

ClementKunz (approved):

PR #183, opened 2024-06-20
Kdesktop 984 implementation of app self restarter causing issues with (un)installer

herve-er:
The implementation of the new App self-restarter is problematic for the installer/uninstaller. The installer/uninstaller needs to kill the app, but the self-restarter causes the app to restart immediately, which is breaking the installer/uninstall process.

## Steps to Reproduce:
1. Launch kDrive App (a version with the self-restarter ie: branch develop)
2. Run the installer or uninstaller.
3. Let it attempt to kill the app.
4. Observe the app restarting immediately due to the self-restarter.

## Expected Result:

The app should remain closed when the  installer or uninstaller kills it, allowing the update process to complete successfully.

## Task List:

- [x] Fix on Windows
- [x] Check if the issue is present on Linux 
    - [x] Fix on Linux (Not concerned)
- [x] Check if the issue is present on macOS (it is)
    - [x] Fix on macOS (@ClementKunz)


herve-er:
@ClementKunz will provide a fix for the mac uninstaller.

ChristopheLarchier (approved):

PR #187, opened 2024-06-21
KDESKTOP-988-Crash-on-Windows-invalid-comparator

ClementKunz:
The crash was introduced following a small refactoring suggested by SonarCloud.
It crashed while processing the event on files retrieved from the back through the '/listing/continue' request

### DIFF
--- a/src/libsyncengine/update_detection/file_system_observer/remotefilesystemobserverworker.cpp
+++ b/src/libsyncengine/update_detection/file_system_observer/remotefilesystemobserverworker.cpp
@@ -482,7 +482,7 @@ ExitCode RemoteFileSystemObserverWorker::sendLongPoll(bool &changes) {
 ExitCode RemoteFileSystemObserverWorker::processActions(Poco::JSON::Array::Ptr actionArray) {
     if (!actionArray) return ExitCodeOk;
 
-    std::set<NodeId, std::equal_to<>> movedItems;
+    std::set<NodeId, std::less<>> movedItems;
 
     for (auto it = actionArray->begin(); it != actionArray->end(); ++it) {
         if (stopAsked()) {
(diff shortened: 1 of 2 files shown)


src/libsyncengine/update_detection/file_system_observer/remotefilesystemobserverworker.cpp
```diff
@@ -482,7 +482,7 @@ ExitCode RemoteFileSystemObserverWorker::sendLongPoll(bool &changes) {
 ExitCode RemoteFileSystemObserverWorker::processActions(Poco::JSON::Array::Ptr actionArray) {
     if (!actionArray) return ExitCodeOk;
 
-    std::set<NodeId, std::equal_to<>> movedItems;
+    std::set<NodeId, std::less<>> movedItems;
```
herve-er, line 485: As `std::less` is the default comparator, could we simply use `std:: set<NodeId>`? Unless you see any advantage to explicitly precise the comparator in this case?

src/libsyncengine/update_detection/file_system_observer/remotefilesystemobserverworker.cpp
```diff
@@ -591,7 +591,7 @@ ExitCode RemoteFileSystemObserverWorker::extractActionInfo(const Poco::JSON::Obj
 }
 
 ExitCode RemoteFileSystemObserverWorker::processAction(const SyncName &usedName, const ActionInfo &actionInfo,
-                                                       std::set<NodeId, std::equal_to<>> &movedItems) {
+                                                       std::set<NodeId, std::less<>> &movedItems) {
```
herve-er, line 594: idem

src/libsyncengine/update_detection/file_system_observer/remotefilesystemobserverworker.cpp
herve-er: image

ClementKunz: 'std::less<Key>' is the default while Sonar suggest to use 'std:less<>'.

> std::less<> is transparent, std::less<anything> is not

https://rules.sonarsource.com/cpp/RSPEC-6021/

herve-er (approved):

PR #192, opened 2024-06-26
Kdesktop 944 delete job unit test on can run function

ChristopheLarchier:


src/libcommonserver/io/iohelper_mac.mm
```diff
@@ -153,13 +154,13 @@ IoError nsError2ioError(NSError *nsError) noexcept {
...
     delete[] buffer;
 
+    // Read target
     Boolean isStale;
     CFURLRef targetUrl =
         CFURLCreateByResolvingBookmarkData(nil, bookmarkRef, kCFBookmarkResolutionWithoutUIMask, nil, nil, &isStale, &error);
     CFRelease(bookmarkRef);
     if (targetUrl == nil) {
         if (error) {
-            ioError = nsError2ioError((NSError *)error);
```
ClementKunz, line 162: So we ignore this error now?

src/libsyncengine/jobs/network/upload_session/uploadsession.cpp
```diff
@@ -453,16 +456,15 @@ bool UploadSession::cancelSession() {
...
-    std::unordered_map<UniqueId, std::shared_ptr<UploadSessionChunkJob>>::iterator it = _ongoingChunkJobs.begin();
-    for (; it != _ongoingChunkJobs.end(); it++) {
-        if (it->second->sessionToken() == _sessionToken) {
-            it->second->abort();
+    for (auto &[jobId, job] : _ongoingChunkJobs) {
+        if (job.get() && job->sessionToken() == _sessionToken) {
+            LOG_INFO(_logger, "Aborting chunk job " << jobId);
+            job->abort();
         }
     }
```
ClementKunz, line 464: No mutex here? In the meantime, some items could have been removed from the map in `uploadChunkCallback`

test/libsyncengine/jobs/network/testnetworkjobs.cpp
```diff
@@ -153,11 +167,11 @@ void TestNetworkJobs::testCopyToDir() {
     CPPUNIT_ASSERT(createTestDir());
 
     SyncName filename = Str("testCopyToDir_") + Str2SyncName(CommonUtility::generateRandomStringAlphaNum()) + Str(".txt");
-    CopyToDirectoryJob job(_driveDbId, testFileRemoteId, _dirId, filename);
+    CopyToDirectoryJob job(_driveDbId, testFileRemoteId, _dummyRemoteDirId, filename);
```
ClementKunz, line 170: The method `createTestDir()` already create a temporary directory and store its ID in `_dirID`. In `tearDown`, this temporary directory is automatically deleted.
Do we really need all this `_dummy*` variables?

src/libcommonserver/io/iohelper_mac.mm
ChristopheLarchier: Yes. When the target path of the alias cannot be read, an empty value is sent to the back, it's not an issue as it is not used later to create the alias (see DownloadJob::createLink).
UploadJob::setQueryParameters has been modified to support sending an empty target path.

src/libsyncengine/jobs/network/upload_session/uploadsession.cpp
ChristopheLarchier: Done

test/libsyncengine/jobs/network/testnetworkjobs.cpp
ChristopheLarchier: I have removed:
        NodeId _dirId;
And added:
        NodeId _dummyLocalDirId;
        NodeId _dummyRemoteDirId;

ClementKunz (approved):

ClementKunz:
The tests has been broken because of a  change in the test user. Will be fixed in https://github.com/Infomaniak/desktop-kDrive/pull/195

PR #194, opened 2024-06-26
KDESKTOP-1006-Re-activate-tests-on-PlateformInconsistencyChecker

ClementKunz:


ChristopheLarchier (dismissed):

src/server/appserver.cpp
```diff
@@ -3094,13 +3094,7 @@ void AppServer::clearSyncNodes() {
     // Clear node tables
     for (const Sync &sync : syncList) {
         SyncPath dbPath = sync.dbPath();
-        std::shared_ptr<SyncDb> syncDbPtr = std::make_shared<SyncDb>(dbPath.string(), _theme->version().toStdString());
-
-#ifdef __APPLE__
-        // Fix the names on local replica if necessary
-        SyncPal::fixFileNamesWithColon(syncDbPtr, sync.localPath());
```
luc-guyot-infomaniak, line 3101: This is a harmless change?

ClementKunz: This is arguable but I choose to remove it because:
1- This function contains a call to `fixNameForbiddenChars` which is something we don't do anymore during the sync and I wanted to remove.
2-`fixFileNamesWithColon` is a function called to migrate the content of the DB from version 3.4.2 and lower. Since these DB did not exists before version 3.4.0, only version 3.4.0 and 3.4.1 are concerned by this migration and no user should remain in those versions.

I should have described more why I removed this part in the PR, sorry

luc-guyot-infomaniak (dismissed):

ChristopheLarchier (approved):

luc-guyot-infomaniak (approved):

PR #199, opened 2024-07-03
[KDESKTOP-986] Handles CSV parsing error of the full listing appropriately

luc-guyot-infomaniak:
This pull-request addresses [KDESKTOP-986](https://infomaniak.atlassian.net/browse/KDESKTOP-986).

The fix consists in setting `error` with `true` via `getItem` when a fatal parsing error of the CSV full listing is detected.
This way the parsing is interrupted as soon as a fatal parsing error is encountered.

The CSV parsing functions are now covered by unit tests.

[KDESKTOP-986]: https://infomaniak.atlassian.net/browse/KDESKTOP-986?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/libsyncengine/jobs/network/csvfullfilelistwithcursorjob.cpp
```diff
@@ -103,14 +203,14 @@ bool CsvFullFileListWithCursorJob::getItem(SnapshotItem &item, bool &error, bool
...
+        // A file name surrounded by double quotes can have a line return in it. If so, read next line and continue parsing
         if (readingDoubleQuotedValue) {
             tmp.push_back('\n');
             readNextLine = true;
-            std::getline(_ss, line);
+            std::getline(ss, line);
             if (line.empty()) {
                 LOGW_WARN(_logger, L"Invalid line");
-                ignore = true;
+                error = true;
```
luc-guyot-infomaniak, line 213: This is the fix.

ClementKunz: Do you think you can display an explicit messgae to the user explaining the problem. With perhaps the file name (or at least a part of it) the causing the parsing issue.

luc-guyot-infomaniak: Done in the next commit.

ClementKunz: Sorry, what I meant was adding an error in the error list explaining why the sync stopped explicitly. Probably calling the `SyncPal::addError` method somewhere.
Do you think it is possible?

src/gui/parametersdialog.cpp
```diff
@@ -576,9 +576,18 @@ QString ParametersDialog::getSyncPalErrorText(QString fctCode, ExitCode exitCode
             break;
         case ExitCodeOk:
         case ExitCodeNeedRestart:
-        case ExitCodeInconsistencyError:
+        case ExitCodeLogicError:
+            if (exitCause == ExitCauseFullListParsingError) {
+                return tr("The parsing of the CSV full listing of the remote drive items failed (error %1).<br>"
```
ClementKunz, line 581: Perhaps too technical for an end user.
Maybe only "Special characters such as double quotes, backslashes or line returns can cause kDrive failures."?

ChristopheLarchier: Agree with the above comment.
+ Move "case ExitCodeInconsistencyError" before "case ExitCodeOk".

luc-guyot-infomaniak: > Move "case ExitCodeInconsistencyError" before "case ExitCodeOk".
But this is going to shift all the exit codes. Isn't it?

luc-guyot-infomaniak: Done.

test/libsyncengine/test.cpp
```diff
@@ -41,8 +42,10 @@ namespace KDC {
 CPPUNIT_TEST_SUITE_REGISTRATION(TestExclusionTemplateCache);
 CPPUNIT_TEST_SUITE_REGISTRATION(TestLocalJobs);
 CPPUNIT_TEST_SUITE_REGISTRATION(TestNetworkJobs);
+CPPUNIT_TEST_SUITE_REGISTRATION(TestSnapshotItemHandler);
 CPPUNIT_TEST_SUITE_REGISTRATION(TestJobManager);
-// CPPUNIT_TEST_SUITE_REGISTRATION(TestSyncDb);
+//  CPPUNIT_TEST_SUITE_REGISTRATION(TestSyncDb);
+
```
herve-er, line 48:
```suggestion
```

ClementKunz (approved):

PR #200, opened 2024-07-04
Sentry message added when a local delete is propagated

ChristopheLarchier:
This pull request sends a `Sentry` message when a local Delete operation is propagated and the item exists but with another id.


src/libcommonserver/io/iohelper.cpp
```diff
@@ -591,6 +591,15 @@ bool IoHelper::checkIfPathExistsWithSameNodeId(const SyncPath &path, const NodeI
             LOGW_WARN(logger(), L"Error in IoHelper::getNodeId for path=" << Path2WStr(path).c_str());
         }
         exists = (nodeId == tmpNodeId);
+
+#ifdef NDEBUG
```
ClementKunz, line 595: Shouldn't we send the sentry only from the the DeleteJob, along side with the debug message "still exist on local replica. Aborting current sync and restart."? Because `checkIfPathExistsWithSameNodeId` is called in others places and we don't want too many false positive sentry messages.

ChristopheLarchier: To discuss...
Whatever the solution there will be false positives.
We can add a 2nd Sentry in the DeleteJob.

luc-guyot-infomaniak: Seconding Clément, I would let the `Sentry` message be as close as possible to the potential issue it aims at warning on. 
The reason is that, without additional context, `exists = false` is just a normal situation for the function `getNodeId`.

src/libcommonserver/io/iohelper.cpp
```diff
@@ -591,6 +591,15 @@ bool IoHelper::checkIfPathExistsWithSameNodeId(const SyncPath &path, const NodeI
             LOGW_WARN(logger(), L"Error in IoHelper::getNodeId for path=" << Path2WStr(path).c_str());
```
luc-guyot-infomaniak, line 591: Out of scope, but nice to have:
```suggestion
            LOGW_WARN(logger(), L"Error in IoHelper::getNodeId for " << Utility::formatSyncPath(path).c_str());
```

ChristopheLarchier: Done

src/libsyncengine/jobs/network/deletejob.cpp
```diff
@@ -95,6 +97,17 @@ bool DeleteJob::canRun() {
         _exitCode = ExitCodeDataError;  // Data error so the snapshots will be re-created
         _exitCause = ExitCauseUnexpectedFileSystemEvent;
         return false;
+    } else if (_localItemId != otherNodeId) {
+        LOGW_DEBUG(_logger, L"Item: " << Utility::formatSyncPath(_absoluteLocalFilepath).c_str()
+                                      << L" exist on local replica with another ID (" << Utility::s2ws(_localItemId).c_str()
```
luc-guyot-infomaniak, line 102:
```suggestion
                                      << L" exists on local replica with another ID (" << Utility::s2ws(_localItemId).c_str()
```

ChristopheLarchier: Fixed

ClementKunz (approved):

PR #201, opened 2024-07-04
KDESKTOP-1031-Update-Sparkle-and-remove-entitlements

ClementKunz:
/!\ You will need to upgrade your Sparkle version to 2.6.4 once this PR is merge


admin/osx/sign_app.sh
```diff
@@ -26,7 +26,7 @@ team_identifier="$3"
 app_domain="$4"
 
 codesign -s "$identity" --force --verbose=4 --options=runtime "$src_app/Contents/Frameworks/Sparkle.framework/Versions/B/XPCServices/Installer.xpc"
-codesign -s "$identity" --force --verbose=4 --options=runtime --entitlements "$HOME/Sparkle/Entitlements/Downloader.entitlements" "$src_app/Contents/Frameworks/Sparkle.framework/Versions/B/XPCServices/Downloader.xpc"
```
luc-guyot-infomaniak, line 29: This will continue to work independently of the version of Sparkle? I mean: should I upgrade my Sparkle installation?

ChristopheLarchier (approved):

admin/osx/sign_app.sh
ClementKunz: Actually, since we removed the use of entitlements, it should work anyway with older versions of Sparkle that contain these entitlements. But you should migrate to at least version 2.6.2 (https://github.com/sparkle-project/Sparkle/releases/tag/2.6.2) because it contains an important security fix

luc-guyot-infomaniak (approved):

PR #202, opened 2024-07-04
Kdesktop 1014 reactivate tests on the job manager

ChristopheLarchier:


src/libsyncengine/jobs/network/upload_session/uploadsession.cpp
```diff
@@ -128,7 +128,7 @@ void UploadSession::runJob() {
 void UploadSession::uploadChunkCallback(UniqueId jobId) {
     const std::scoped_lock lock(_mutex);
     auto jobInfo = _ongoingChunkJobs.extract(jobId);
-    if (jobInfo.mapped()) {
+    if (!jobInfo.empty()) {
         if (jobInfo.mapped()->hasHttpError() || jobInfo.mapped()->exitCode() != ExitCodeOk) {
```
ClementKunz, line 132: Can `jobInfo.mapped()` be null at this point?

src/libsyncengine/propagation/executor/executorworker.cpp
```diff
@@ -1559,6 +1561,8 @@ bool ExecutorWorker::enoughLocalSpace(SyncOpPtr syncOp) {
 }
 
 void ExecutorWorker::waitForAllJobsToFinish(bool &hasError) {
+    const std::scoped_lock lock(_mutex);
+
```
ClementKunz, line 1565: Is there a risk of dead lock here since we lock again inside `deleteFinishedAsyncJobs`?

ClementKunz: Actually, `std::scoped_lock` embed a deadlock avoidance algorithm so it is probably ok. I am not sure however about the order of lock/unlock.

ClementKunz: My bad, `_mutex` is a recursive mutex. Ok for me

src/libsyncengine/jobs/network/upload_session/uploadsession.cpp
ChristopheLarchier: Fixed

PR #203, opened 2024-07-05
Kdesktop 1035 add unit test for kdesktop 754

herve-er:
This PR aims to provide a unit test for PR 754. It also includes some bug fixes revealed during the implementation of the tests.


luc-guyot-infomaniak (dismissed):

test/libsyncengine/propagation/executor/executorworker/testexecutorworker.cpp
```diff
@@ -0,0 +1,152 @@
...
+     */
+
+    // Setup DB
+    DbNodeId dbNodeIdA;
+    DbNodeId dbNodeIdAA;
+    DbNodeId dbNodeIdAB;
+    DbNodeId dbNodeIdAAA;
+
+    bool constraintError = false;
+    DbNode dbNodeA(0, _syncPal->_syncDb->rootNode().nodeId(), Str("A"), Str("A"), "lA", "rA", defaultTime, defaultTime,
```
luc-guyot-infomaniak, line 60:
```suggestion
    DbNode dbNodeA(0, _syncPal->syncDb()->rootNode().nodeId(), Str("A"), Str("A"), "lA", "rA", defaultTime, defaultTime,
```

test/libsyncengine/propagation/executor/executorworker/testexecutorworker.cpp
```diff
@@ -0,0 +1,152 @@
...
+                   defaultTime, NodeType::NodeTypeDirectory, 0, std::nullopt);
+    _syncPal->_syncDb->insertNode(dbNodeA, dbNodeIdA, constraintError);
```
luc-guyot-infomaniak, line 62:
```suggestion
    _syncPal->syncDb()->insertNode(dbNodeA, dbNodeIdA, constraintError);
```
You should only use the public interface, whenever possible.

test/libsyncengine/syncpal/testsyncpal.cpp
```diff
@@ -94,8 +94,70 @@ void TestSyncPal::tearDown() {
...
+}
+
+void TestSyncPal::testUpdateTree() {
+    auto updateTree = _syncPal->updateTree(ReplicaSideLocal);
+    CPPUNIT_ASSERT_EQUAL(_syncPal->_localUpdateTree, updateTree);
+
+    updateTree = _syncPal->updateTree(ReplicaSideRemote);
+    CPPUNIT_ASSERT_EQUAL(_syncPal->_remoteUpdateTree, updateTree);
+
+    updateTree = _syncPal->updateTree(ReplicaSideUnknown);
```
luc-guyot-infomaniak, line 108: No need to test accessors. I would only test this special case.

test/libsyncengine/syncpal/testsyncpal.cpp
```diff
@@ -94,8 +94,70 @@ void TestSyncPal::tearDown() {
...
+    CPPUNIT_ASSERT_EQUAL(std::shared_ptr<UpdateTree>(nullptr), updateTree);
+}
+
+void TestSyncPal::testSnapshot() {
+    auto snapshot = _syncPal->snapshot(ReplicaSideLocal);
```
luc-guyot-infomaniak, line 113: Idem.

test/libsyncengine/syncpal/testsyncpal.cpp
```diff
@@ -94,8 +94,70 @@ void TestSyncPal::tearDown() {
...
+
+    snapshot = _syncPal->snapshot(ReplicaSideRemote, true);
+    CPPUNIT_ASSERT_EQUAL(_syncPal->_remoteSnapshotCopy, snapshot);
+
+    snapshot = _syncPal->snapshot(ReplicaSideUnknown, true);
+    CPPUNIT_ASSERT_EQUAL(std::shared_ptr<Snapshot>(nullptr), snapshot);
+}
+
+void TestSyncPal::testOperationSet() {
+    auto operationSet = _syncPal->operationSet(ReplicaSideLocal);
```
luc-guyot-infomaniak, line 133: Idem.

test/libsyncengine/syncpal/testsyncpal.h
```diff
@@ -25,26 +26,31 @@ namespace KDC {
...
 
-    protected:
-        void testAll();
-        void testConflictQueue();
-
     private:
         std::shared_ptr<SyncPal> _syncPal;
         std::shared_ptr<ParmsDb> _parmsDb;
-
+        std::shared_ptr<LocalTemporaryDirectory> _localTempDir;
```
luc-guyot-infomaniak, line 42: Why is it better to hold a shared pointer instead of an object?

test/libsyncengine/update_detection/file_system_observer/testfsoperationset.cpp
```diff
@@ -0,0 +1,523 @@
...
+    for (OperationType type : _operationTypes) {
+        for (int i = 0; i < 3; i++) {
+            auto op = std::make_shared<FSOperation>(type, nodeId + nodeIdSuffix);
+            nodeIdSuffix++;
+            fsOperationSet.insertOp(op);
+        }
+    }
+
+    FSOperationSet fsOperationSetCopy(fsOperationSet);
+    CPPUNIT_ASSERT_EQUAL(fsOperationSet._ops.size(), fsOperationSetCopy._ops.size());
```
luc-guyot-infomaniak, line 508: Use only the public interface of `FSOperationSet`. You can extend it, if need be.

test/libsyncengine/update_detection/file_system_observer/testfsoperationset.cpp
```diff
@@ -0,0 +1,523 @@
...
+    FSOperationSet fsOperationSet;
+    NodeId nodeId = "nodeId";
+    std::vector<UniqueId> opsIds;
+    for (OperationType type : _operationTypes) {
+        auto op = std::make_shared<FSOperation>(type, nodeId);
+        fsOperationSet.insertOp(op);
+        opsIds.push_back(op->id());
+    }
+
+    // Test with a not existing node
```
luc-guyot-infomaniak, line 83:
```suggestion
    // Test with a non existing node
```

test/libsyncengine/update_detection/file_system_observer/testfsoperationset.cpp
```diff
@@ -0,0 +1,523 @@
...
+    std::unordered_set<UniqueId> ops;
+    ops.insert(1);
+    CPPUNIT_ASSERT(!fsOperationSet.getOpsByNodeId("notExistingNodeId", ops));
```
luc-guyot-infomaniak, line 86:
```suggestion
    CPPUNIT_ASSERT(!fsOperationSet.getOpsByNodeId("nonExistingNodeId", ops));
```

test/libsyncengine/update_detection/file_system_observer/testfsoperationset.cpp
```diff
@@ -0,0 +1,523 @@
...
+        for (int i = 1; i <= 2; i++) {
+            auto op = std::make_shared<FSOperation>(type, nodeId + nodeIdSuffix);
+            nodeIdSuffix++;
+            fsOperationSet.insertOp(op);
+        }
+    }
+
+    // Test clear
+    fsOperationSet.clear();
+    CPPUNIT_ASSERT_EQUAL(size_t(0), fsOperationSet._ops.size());
```
luc-guyot-infomaniak, line 135: Use only the public interface of `FSOperationSet`. Extend it if need be.

test/libsyncengine/update_detection/file_system_observer/testfsoperationset.cpp
```diff
@@ -0,0 +1,523 @@
...
+        fsOperationSet.insertOp(op);
+    }
+
+    for (auto type : _operationTypes) {
+        FSOpPtr res;
+        CPPUNIT_ASSERT(fsOperationSet.findOp(nodeId, type, res));
+        CPPUNIT_ASSERT(res->operationType() == type);
+    }
+
+    // Test findOp with a not existing operation
```
luc-guyot-infomaniak, line 196:
```suggestion
    // Test findOp with a non existing operation.
```

test/libsyncengine/update_detection/file_system_observer/testfsoperationset.cpp
```diff
@@ -0,0 +1,523 @@
...
+    fsOperationSet.clear();
+    FSOpPtr res;
+    fsOperationSet.insertOp(std::make_shared<FSOperation>(OperationTypeCreate, nodeId));
+    CPPUNIT_ASSERT(!fsOperationSet.findOp("notExistingNodeId", OperationTypeCreate, res));
```
luc-guyot-infomaniak, line 200:
```suggestion
    CPPUNIT_ASSERT(!fsOperationSet.findOp("nonExistingNodeId", OperationTypeCreate, res));
```

test/libsyncengine/update_detection/file_system_observer/testfsoperationset.cpp
```diff
@@ -0,0 +1,523 @@
...
+
+    fsOperationSet.insertOp(std::make_shared<FSOperation>(OperationTypeCreate, nodeId));
+    CPPUNIT_ASSERT(!fsOperationSet.findOp(nodeId, OperationTypeMove, res));
+}
+
+void TestFsOperationSet::testOperatorEqual() {
+    FSOperationSet fsOperationSet1;
+    FSOperationSet fsOperationSet2;
+
+    // Test operator = with an empty set
```
luc-guyot-infomaniak, line 210:
```suggestion
    // Test operator the assignment operator '=' with an empty set
```

src/libsyncengine/syncpal/syncpal.h
```diff
@@ -218,14 +218,11 @@ class SYNCENGINE_EXPORT SyncPal : public std::enable_shared_from_this<SyncPal> {
...
 
-        std::shared_ptr<UpdateTree> getUpdateTree(ReplicaSide side) {
-            return side == ReplicaSideLocal ? _localUpdateTree : _remoteUpdateTree;
-        }
-
         //! Makes copies of real-time snapshots to be used by synchronization workers.
         void copySnapshots();
 
     private:
+        friend class TestSyncPal;
```
ClementKunz, line 225: Already defined below

src/libsyncengine/update_detection/file_system_observer/fsoperationset.cpp
```diff
@@ -17,62 +17,57 @@
...
-            _mutex.unlock();
-        }
         return true;
     }
-
-    if (lockMutex) {
-        _mutex.unlock();
-    }
     return false;
 }
```
ClementKunz, line 40: Perhaps we should use recursive mutex instead, therefore we could keep only 1 `getOp` method.

luc-guyot-infomaniak (commented):
A recurrent remark about the use of private members and methods by friend classes. To be discussed.

test/libsyncengine/reconciliation/conflict_resolver/testconflictresolverworker.cpp
```diff
@@ -726,12 +726,62 @@ void TestConflictResolverWorker::testMoveMoveCycle() {
...
+    std::shared_ptr<Node> lNodeAAA = _syncPal->_localUpdateTree->getNodeById("lAAA");
+    std::shared_ptr<Node> lNodeAB = _syncPal->_localUpdateTree->getNodeById("lAB");
+    lNodeAAA->setChangeEvents(OperationTypeMove);
+    lNodeAAA->setMoveOriginParentDbId(lNodeAA->idb());
+    lNodeAAA->setMoveOrigin("A/AA/AAA");
+    lNodeAAA->setParentNode(lNodeAB);
+    lNodeAA->deleteChildren(lNodeAAA);
+    lNodeAB->insertChildren(lNodeAAA);
+
+    // Simulate move of node AB to AA, on remote replica
```
ClementKunz, line 763: Are you sure this creates a cycle? UpdateTreeWorker::getNewPathAfterMove

src/libsyncengine/update_detection/file_system_observer/fsoperationset.cpp
herve-er: done 👍🏾

src/libsyncengine/update_detection/file_system_observer/computefsoperationworker.cpp
```diff
@@ -584,8 +584,8 @@ ExitCode ComputeFSOperationWorker::checkFileIntegrity(const DbNode &dbNode) {
             return ExitCodeOk;
         }
 
-        const bool localSnapshotIsLink = _syncPal->_localSnapshotCopy->isLink(dbNode.nodeIdLocal().value());
-        if (localSnapshotIsLink) {
+        if (const bool localSnapshotIsLink = _syncPal->snapshot(ReplicaSideLocal, true)->isLink(dbNode.nodeIdLocal().value());
```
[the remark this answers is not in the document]
herve-er: This call is in the following `if` statement : ```if (... && dbNode.nodeIdLocal().has_value() && ...) {``` line 574 preventing for a call on an empty `nodeIdLocal`.

src/libsyncengine/syncpal/syncpal.cpp
```diff
@@ -964,6 +964,10 @@ ExitCode SyncPal::updateSyncNode() {
 }
 
 std::shared_ptr<Snapshot> SyncPal::snapshot(ReplicaSide side, bool copy) {
+    if (side == ReplicaSideUnknown) {
+        LOG_ERROR(_logger, "Call to SyncPal::snapshot with 'ReplicaSideUnknown').");
```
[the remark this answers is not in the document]
herve-er: The three strings are distinct (function names). Are you considering removing the function name from the log since we already have the line information?

test/libsyncengine/reconciliation/conflict_resolver/testconflictresolverworker.cpp
herve-er: Typo in the comment AB is move to AAA not AA

herve-er, line 763:
```suggestion
    // Simulate move of node AB to AAA, on remote replica
```

herve-er: https://github.com/Infomaniak/desktop-kDrive/pull/203/files#r1677889007

src/libsyncengine/syncpal/syncpal.cpp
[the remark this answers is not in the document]
herve-er: As this is a recurrent comment, what would you think of a dedicated namespace `TypesUtility` like this:

```
namespace TypesUtility {

constexpr const char *unknownStr = "No string value defined for this enum value :";

std::string enum2str(AppStateKey val) noexcept {
    switch (val) {
        using enum KDC::AppStateKey;
        case LastServerSelfRestartDate:
            return "LastServerSelfRestartDate";
        case LastClientSelfRestartDate:
            return "LastClientSelfRestartDate";
        case LastSuccessfulLogUploadDate:
            return "LastSuccessfulLogUploadDate";
        case LastLogUploadArchivePath:
            return "LastLogUploadArchivePath";
        case LogUploadState:
            return "LogUploadState";
        case LogUploadPercent:
            return "LogUploadPercent";
        case Unknown:
            return "Unknown";
        default:
            return sts::string(unknownStr) + std::to_string(static_cast<int>(val));
    }
}

std::string enum2str(ReplicaSide val) noexcept {
    switch (val) {
        case ReplicaSideUnknown:
            return "Unknown";
        case ReplicaSideLocal:
            return "Local";
        case ReplicaSideRemote:
            return "Remote";
        default:
            return sts::string(unknownStr) + std::to_string(static_cast<int>(val));
    }
}

...
...
...
}  // namespace TypesUtility
```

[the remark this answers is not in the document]
luc-guyot-infomaniak: Yes, that would be nice. I would advocate for a short and systematic `toStr(MyEnumType enum)`, and also `toWStr((MyEnumType enum)` when appropriate.

test/libsyncengine/syncpal/testsyncpal.h
luc-guyot-infomaniak: I didn't realize that there was a risk of temp dir name collisions across tests (we use the time stamp, but the truncation is around 1 minute). I believe we should make the temporary directory name unique for each test. Otherwise some tests may share the same temporary directory, with unpredictable side effects.

luc-guyot-infomaniak (commented):
Major point: the creation and deletion of temporary directories in unit tests.

src/libsyncengine/update_detection/file_system_observer/fsoperationset.cpp
```diff
@@ -17,62 +17,57 @@
...
 bool FSOperationSet::getOpsByNodeId(const NodeId &nodeId, std::unordered_set<UniqueId> &ops) {
-    const std::lock_guard<std::mutex> lock(_mutex);
-    auto it = _opsByNodeId.find(nodeId);
-    if (it != _opsByNodeId.end()) {
+    const std::scoped_lock lock(_mutex);
+    if (auto it = _opsByNodeId.find(nodeId); it != _opsByNodeId.end()) {
         ops = it->second;
         return true;
     }
+    ops.clear();
```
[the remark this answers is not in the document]
herve-er: As `ops.clear()` is unnecessary when we find the nodeId (which occurs most of the time), I have excluded it from the if statement for optimization (to avoid an unnecessary call).

test/libsyncengine/syncpal/testsyncpal.h
herve-er: Following our discussion, a new instance of the testFixture is indeed created for each test. I have removed the use of shared_ptr and improved the behavior of localTemporaryDirectory to ensure that two tests do not use the same directory. [7d95284](https://github.com/Infomaniak/desktop-kDrive/pull/203/commits/7d952842d3a24c69e60e9330d937db0149bb9ca8)

src/libsyncengine/syncpal/syncpal.cpp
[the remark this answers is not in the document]
herve-er: I created a new Jira tcket with the only goal to adress this issue: https://infomaniak.atlassian.net/browse/KDESKTOP-1054

src/libsyncengine/update_detection/file_system_observer/fsoperationset.cpp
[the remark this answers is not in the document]
ClementKunz: The underlining question here is : is it the responsibility of the called function to initialize (clear in this context) the variable passed by reference? Or is it the responsibility of the caller?

Anyway, we could also never clear `ops` but return a boolean value, `true` if found, `false` if not. That way, the caller can discard the returned value if not found.

test/libsyncengine/propagation/executor/executorworker/testexecutorworker.h
```diff
@@ -0,0 +1,50 @@
...
+    public:
+        void setUp() override;
+        void tearDown() override;
+
+    private:
+        void testAffectedUpdateTree();
+        void testTargetUpdateTree();
+
+        log4cplus::Logger _logger;
+        std::shared_ptr <ExecutorWorker> _testObj;
```
ClementKunz, line 46: Since there is already an ExecutorWorker in SyncPal, those 2 variables are probably not both useful. Either, use the `_executorWorker` attribute of SyncPal or, if possible, do not set up a SyncPal at all and only create an ExecutorWorker

src/libsyncengine/update_detection/file_system_observer/fsoperationset.cpp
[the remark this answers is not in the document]
herve-er: > is it the responsibility of the called function to initialize (clear in this context) the variable passed by reference? Or is it the responsibility of the caller?

In my opinion, it is the responsibility of the called function to initialize (clear in this context) the variable passed by reference. This approach ensures that the result is consistent and reliable, as the called function has full control over the state of the variable. 
Additionally, it makes the code easier to maintain and debug, as there is only one place where the variable critticaly need to be initialized, rather than multiple places throughout the codebase.
(found an issue recently with UpdateInfo::parseString acting randomly because of this).

> Anyway, we could also never clear ops but return a boolean value, true if found, false if not. That way, the caller can discard the returned value if not found.

Most of the time, the result goes to a for loop, by assuring here that ops contains only the desired element discharge the caller for checking an other other boolean (when needed, it still can check the return value of the function wich correspond to found or not found).  It will just exit the loop immediatly. 
But it could maybe make the call function more explicit on how it behave?

> Actually, wouldn't it be simpler to let the function return a set of IDs?

It could be a solution yes. I think the time complexity would be the same as anyway we already do a copy

test/libsyncengine/propagation/executor/executorworker/testexecutorworker.h
herve-er: I'm trying to apply your suggestion, but i'm faccing the following issue:
The `_executorWorker` attribute of `SyncPal` is initialize only after the call to `SyncPal::Start()` wich need to set a user, drive and a syncro wich seems to be a little exagerated just for the `ExecutorWorker` tests. 

Only create an `ExecutorWorker` is not viable as the tested function rely at some point on the `_syncPal` memeber of `ExecutorWorker`.

ClementKunz: Ok, so perhaps this change should be the topic of a dedicated PR where we refactor deeper the way the tests are implemented.
We can resolve this thread

src/libsyncengine/update_detection/file_system_observer/fsoperationset.cpp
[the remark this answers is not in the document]
ClementKunz: > In my opinion, it is the responsibility of the called function to initialize (clear in this context) the variable passed by reference. This approach ensures that the result is consistent and reliable, as the called function has full control over the state of the variable.

Ok for me. As long as it is clear in the code. So IMHO we can keep it like that or apply Luc's suggestion

ChristopheLarchier (dismissed):

test/libsyncengine/test.cpp
```diff
@@ -41,17 +42,17 @@
...
 CPPUNIT_TEST_SUITE_REGISTRATION(TestNetworkJobs);
 CPPUNIT_TEST_SUITE_REGISTRATION(TestJobManager);
-
 CPPUNIT_TEST_SUITE_REGISTRATION(TestSnapshot);
+CPPUNIT_TEST_SUITE_REGISTRATION(TestFsOperationSet);
 // CPPUNIT_TEST_SUITE_REGISTRATION(TestLocalFileSystemObserverWorker);
 CPPUNIT_TEST_SUITE_REGISTRATION(TestSnapshotItemHandler);
-CPPUNIT_TEST_SUITE_REGISTRATION(TestRemoteFileSystemObserverWorker);
-// CPPUNIT_TEST_SUITE_REGISTRATION(TestComputeFSOperationWorker);
+//CPPUNIT_TEST_SUITE_REGISTRATION(TestRemoteFileSystemObserverWorker);
```
ClementKunz, line 54: This test should be activated

src/libsyncengine/update_detection/file_system_observer/fsoperationset.cpp
```diff
@@ -84,52 +80,60 @@ void FSOperationSet::clear() {
...
+            res = opPtr;
+            return true;
         }
     }
     return false;
 }
 
-FSOperationSet &FSOperationSet::operator=(const FSOperationSet &other) {
+ReplicaSide FSOperationSet::side() const {
+    const std::scoped_lock lock(_mutex);
```
ClementKunz, line 131: This lock is useless, variable `_side` is never changed

src/libsyncengine/propagation/executor/executorworker.h
```diff
@@ -147,6 +147,7 @@ class ExecutorWorker : public OperationProcessor {
 
         bool _snapshotToInvalidate = false;
 
+        friend class TestExecutor;
```
ClementKunz, line 150: Useless?

src/libsyncengine/update_detection/file_system_observer/fsoperationset.cpp
herve-er: [b3b1d7c](https://github.com/Infomaniak/desktop-kDrive/pull/203/commits/b3b1d7c8c92629981ee79b3407a15cc9e77051d7)

src/libsyncengine/propagation/executor/executorworker.h
herve-er: [64d2da1](https://github.com/Infomaniak/desktop-kDrive/pull/203/commits/64d2da18f83f1e3a9bcaf55f4348c28daae5096c)

PR #206, opened 2024-07-08
[KDESKTOP-990] Re-enable LFSO unit tests

luc-guyot-infomaniak:


src/libcommonserver/io/iohelper_win.cpp
```diff
@@ -202,8 +212,9 @@ bool IoHelper::getFileStat(const SyncPath &path, FileStat *buf, IoError &ioError
                 continue;
             }
 
-            LOGW_WARN(logger(), L"Error in CreateFileW: " << Utility::formatSyncPath(path.parent_path()).c_str());
             ioError = dWordError2ioError(dwError);
+            LOGW_WARN(logger(), L"Error in CreateFileW: " << Utility::formatIoError(path.parent_path(), ioError).c_str());
+
```
ClementKunz, line 217: Should we really log a warning message, while the error might be "expected"?

ClementKunz: Let's keep it that way for now

ClementKunz (approved):

PR #234, opened 2024-07-31
Fix vfs socket bug.

herve-er:
Summary: Fix issue with random crashes during app download step related to QLocalSocket internal buffer management.

Description: We have identified an issue where the application crashes randomly during the download step, initially thought to be related to large sync operations.

Upon investigation, the error occurs within the vfsForceStatus call, specifically when the socket->write method is invoked within SocketListener::sendMessage. The root cause appears to be inconsistencies in the QLocalSocket's internal buffer, which occasionally becomes empty or changes size unpredictably.

During downloads, the write method is called in a dedicated thread, while some QLocalSocket private callback rely on connect and are  handled by the event loop (in the main thread). This asynchronous handling can lead to a situation where one download callback interferes with another, corrupting the download buffer.

Solution: To mitigate this issue, we propose emitting a signal from the download thread instead of directly invoking the function. This approach ensures that all operations are processed sequentially by the application's event loop in the main thread, preventing interference between multiple download instances and other file explorer events.

### DIFF
--- a/src/server/appserver.h
+++ b/src/server/appserver.h
@@ -132,7 +132,8 @@ class AppServer : public SharedTools::QtSingleApplication {
         void parseOptions(const QStringList &);
         void initLogging() noexcept(false);
         void setupProxy();
-        void handleCrashRecovery(bool &shouldQuit);  // Sets `shouldQuit` with true if the crash recovery is successful, false if the application should exit.
+        void handleCrashRecovery(bool &shouldQuit);  // Sets `shouldQuit` with true if the crash recovery is successful, false if
+                                                     // the application should exit.
         bool serverCrashedRecently(int seconds = 60 /*Allow one server self restart per minute (default)*/);
         bool clientCrashedRecently(int second = 60 /*Allow one client self restart per minute (default)*/);
 
@@ -248,6 +249,9 @@ class AppServer : public SharedTools::QtSingleApplication {
         void onRequestReceived(int id, RequestNum num, const QByteArray &params);
         void onRestartClientReceived();
         void onMessageReceivedFromAnotherProcess(const QString &message, QObject *);
+
+    signals:
+        void socketApiExecuteCommandDirect(const QString commandLine);
 };
 
 
--- a/src/server/socketapi.cpp
+++ b/src/server/socketapi.cpp
@@ -149,8 +149,7 @@ SocketApi::~SocketApi() {
     _listeners.clear();
 }
 
-void SocketApi::executeCommandDirect(const char *commandLine) {
-    QString commandLineStr(commandLine);
+void SocketApi::executeCommandDirect(const QString commandLineStr) {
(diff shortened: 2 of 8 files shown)


ChristopheLarchier:
I'm not a fan of this solution (Qt signal in AppServer).

src/server/appserver.h
```diff
@@ -248,6 +249,9 @@ class AppServer : public SharedTools::QtSingleApplication {
         void onRequestReceived(int id, RequestNum num, const QByteArray &params);
         void onRestartClientReceived();
         void onMessageReceivedFromAnotherProcess(const QString &message, QObject *);
+
+    signals:
+        void socketApiExecuteCommandDirect(const QString commandLine);
```
luc-guyot-infomaniak, line 254:
```suggestion
        void socketApiExecuteCommandDirect(const QString &commandLine);
```

src/server/socketapi.cpp
```diff
@@ -149,8 +149,7 @@ SocketApi::~SocketApi() {
     _listeners.clear();
 }
 
-void SocketApi::executeCommandDirect(const char *commandLine) {
-    QString commandLineStr(commandLine);
+void SocketApi::executeCommandDirect(const QString commandLineStr) {
```
luc-guyot-infomaniak, line 152:
```suggestion
void SocketApi::executeCommandDirect(const QString &commandLineStr) {
```

src/server/socketapi.h
```diff
@@ -86,11 +86,12 @@ class SocketApi : public QObject {
             _getPublicLinkUrl = getPublicLinkUrl;
         }
 
-        void executeCommandDirect(const char *commandLine);
         void unregisterSync(int syncDbId);
         void registerSync(int syncDbId);
 
         static bool syncForPath(const std::filesystem::path &path, KDC::Sync &sync);
+    public slots:
+        void executeCommandDirect(const QString commandLine);
```
luc-guyot-infomaniak, line 94:
```suggestion
        void executeCommandDirect(const QString &commandLine);

```

src/server/socketlistener.cpp
```diff
@@ -34,9 +34,15 @@ bool BloomFilter::isHashMaybeStored(uint hash) const {
...
 }
 
-SocketListener::SocketListener(QIODevice *socket) : socket(socket) {}
+SocketListener::SocketListener(QIODevice *socket) : socket(socket) {
+    _threadId = std::this_thread::get_id();
+}
 
 void SocketListener::sendMessage(const QString &message, bool doWait) const {
+    if (_threadId != std::this_thread::get_id()) {
+        LOGW_ERROR(KDC::Log::instance()->getLogger(), L"SocketListener::sendMessage should only be called from the main thread");
```
luc-guyot-infomaniak, line 43: Can this happen under normal circumstances? (If not, we may consider an assertion failure instead.)
If this situation happens for a user, how should the application respond?

ChristopheLarchier (approved):

PR #236, opened 2024-08-02
Kdesktop 1096 crash in socket listener send message

ChristopheLarchier:


src/server/socketapi.cpp
```diff
@@ -214,6 +214,7 @@ void SocketApi::slotNewConnection() {
         return;
     }
 
+    QMutexLocker lock(&_sendMessageMutex);
```
herve-er, line 217: Unfortunately, I'm not sure this will solve our issue. When we call the QLocalSocket `write` method, the first chunk of data is handled by the calling thread, but the next ones will be handled by the main thread (event loop). I do think this is where our issue is coming from.
The thread clash occur when a thread is sending it's first chunk of data and at the same time, the main thread is sending  chunk from previous call to write.
Had your sync been successful with this approach?

herve-er (changes requested):

PR #242, opened 2024-08-08
Allow lastModTime < 0 in csv parsing.

herve-er:


src/libsyncengine/jobs/network/csvfullfilelistwithcursorjob.cpp
```diff
@@ -120,13 +120,6 @@ bool SnapshotItemHandler::updateSnapshotItem(const std::string &str, CsvIndex in
                 logError(L"setLastModified", L"out_of_range", str, exc);
                 return false;
             }
-
-            if (item.lastModified() < 0) {
```
luc-guyot-infomaniak, line 124: *Question:*  I am wondering if it would be useful to log something though.

In any case, I would add a comment saying that negative values are purposely supported.

luc-guyot-infomaniak (commented):
Question about logging or not logging.

ChristopheLarchier (approved):

PR #249, opened 2024-08-09
Handling inconsistencies in the update tree

ChristopheLarchier:


src/libsyncengine/update_detection/update_detector/node.cpp
```diff
@@ -187,4 +220,18 @@ SyncPath Node::getPath() const {
     return path;
 }
 
+bool Node::isParentValid(const std::shared_ptr<Node> parentNode) const {
+    for (auto &child : _childrenById) {
+        if (child.second == parentNode) {
+            return false;
+        }
+
+        if (!child.second->isParentValid(parentNode)) {
```
luc-guyot-infomaniak, line 229: If every parent node is set via `setParent` (advised), then the recursive call is superfluous.

luc-guyot-infomaniak (commented):
Suggestion: constrain the node API to use always `setParent` to set a parent node and then remove recursion in `isParentValid`.

src/libsyncengine/update_detection/update_detector/node.cpp
```diff
@@ -91,6 +91,21 @@ void Node::setName(const SyncName &name) {
     _name = Utility::normalizedSyncName(name);
 }
 
+bool Node::setParentNode(const std::shared_ptr<Node> &parentNode) {
```
luc-guyot-infomaniak, line 94: We should check that this method is also called in `Node` constructors which are provided with a parent node.

luc-guyot-infomaniak (commented):
A remark about the `Node` constructors.

src/libsyncengine/update_detection/update_detector/node.cpp
ClementKunz: For now, I have only just removed the Node constructor that allow to create a Node without parent. Let's re discuss this modification

src/libsyncengine/update_detection/update_detector/node.cpp
ClementKunz: I did not find a way to avoid the recursion. However, I change the function to climb up the update tree instead of going down and checking every child.

src/libsyncengine/update_detection/update_detector/node.cpp
ClementKunz: Edit: Finally, I rewrite the `isParentValid` method to climb up the tree and testing only direct parents, instead of going down the tree and testing every child

src/libsyncengine/update_detection/update_detector/node.cpp
luc-guyot-infomaniak: Changed in the latest commits: the method is now iterative.

luc-guyot-infomaniak (approved):

PR #261, opened 2024-08-20
[DESKTOP-1098] Ignore the local path information of a DbNode

luc-guyot-infomaniak:
The changes of this pull-request  removes all unnecessary logic about the local path information of a `DbNode` as it is now redundant with the remote path information.

Unit tests check that the modified `path` method. works as expected.

Some on-the-way cleanup changes accompany the removal of database local names and paths from the application logic:
- more constness
- initialized formerly non-initialized variables.
- use `Utility::formatSyncPath` instead of `Path2WStr`.
- replace `snapshot` with `side` when `snapshot` is of type `ReplicaSide`.


luc-guyot-infomaniak:
This pull-request is abandoned because the encoding issues with file names will be solved by storing the actual file system encoded file names in the synchronisation database. See current work on https://infomaniak.atlassian.net/browse/KDESKTOP-1152.

PR #262, opened 2024-08-21
Kdesktop 1105 uniformisation of trash bin in UI

herve-er:


luc-guyot-infomaniak (commented):
This first commit needs to reverted since the LiteSyncExtConnector relies on FileSystem on MacOSX (build error). Otherwise, the rest looks good.

luc-guyot-infomaniak (approved):

PR #263, opened 2024-08-21
Kdesktop 1054  Replace Enum Code with Name in Logs and Sentry

herve-er:
# Description
Currently, in our logs and Sentry, all enum values are represented by their respective codes. This has led to difficulties in identifying the corresponding values, as it requires navigating through the codebase each time an enum value is encountered.

To improve this process, this PR aims to display the name of the enum value instead of its code in both the logs and Sentry. Previously, this was not feasible due to the use of enums with type def instead of enum classes. However, since we have replaced these instances, there should be no issues implementing this change.

Before:
image
image

Now: 
image

It also simplify the code inside LOG macro. For exemple:
image
become: 
image
without any change in the output str.

## Good to know
There is no need to worry about string or wide string, as the enum name will be converted to the right type depending on the stream type. 

The template is only defined for enum classes where enumClassToString(enum class) is available. In other cases, it will raise an error in the log macro, as enum classes hopefully cannot be implicitly converted to int.  (see concepts and requires C++20).

Enum classes can be manually converted to std::string using the method toString(enum class x), if the conversion has been implemented in types.cpp.


src/libcommon/comm.h
```diff
@@ -123,6 +124,172 @@ enum class RequestNum {
     RECONSIDER_SKIPPED_UPDATE,
     UTILITY_QUIT,
 };
+inline std::string enumClassToString(RequestNum e) {
```
luc-guyot-infomaniak, line 127: Why not putting the implementation in `types.cpp`?

src/libcommon/utility/types.cpp
```diff
@@ -0,0 +1,730 @@
...
+namespace KDC {
+namespace typesUtility {
+std::wstring stringToWideString(const std::string &str) {
+    std::wstring wstr;
+    Poco::UnicodeConverter::convert(str, wstr);
+    return wstr;
+}
+}  // namespace typesUtility
+
+std::string enumClassToString(ReplicaSide e) {
```
ClementKunz, line 32: As discussed this morning, it would be great to call those helper functions only `toString`, `toWideString`, ...
And start using this rule everywhere

herve-er: I updated theses functions names here: [6d13250](https://github.com/Infomaniak/desktop-kDrive/pull/263/commits/6d1325018b6096378d1e90df97a0430510b30d4c)

src/libcommon/comm.h
herve-er: The idea was to have only the enum from types.h in types.cpp. In fact, it is possible to add the converters for RequestNum and SignalNum directly in types.h as it is the same lib, but it may be better to keep them close to the enum they are referencing. For example, if an enum is defined in the libcommon server, we would not be able to set the converter in types.cpp as is in libcommon and does not import libcommonserver.

ClementKunz (approved):

src/libcommon/utility/types.h
```diff
@@ -26,6 +26,7 @@
 #include <sstream>
 #include <unordered_set>
 #include <variant>
+#include <qdebug.h>
```
ChristopheLarchier, line 29: Usefull?

herve-er: It is needed to get the same beahviour with side logs, ie:
`qCDebug(lcCommClient()) << "Snd rqst" << id << num;`

line 497:
```
inline QDebug &operator<<(QDebug &os, C e) {
    return os << toStringWithCode(e).c_str();
}
```

src/libcommon/comm.h
```diff
@@ -123,6 +124,172 @@ enum class RequestNum {
     RECONSIDER_SKIPPED_UPDATE,
     UTILITY_QUIT,
 };
+inline std::string toString(RequestNum e) {
+    switch (e) {
+        case RequestNum::LOGIN_REQUESTTOKEN:
```
luc-guyot-infomaniak, line 129: Is there a Jira task for changing the case of enum values like `LOGIN_REQUESTTOKEN`?

herve-er: There wasn't, just created it -> https://infomaniak.atlassian.net/browse/KDESKTOP-1166

ChristopheLarchier (approved):

PR #264, opened 2024-08-22
KDESKTOP-1140-Fix-LFSO-tests

ClementKunz:
`TestLocalFileSystemObserverWorker` now generates its own test files. It does not use files from git repository


herve-er (commented):
Some interrogation about the use of system call instead of IoHelper/std::filesystem methods.

ClementKunz:
> Some interrogations about the use of system call instead of IoHelper/std::filesystem methods.

I cannot think of a good argument in favor of those system calls. I will apply the changes

test/libsyncengine/test.cpp
```diff
@@ -42,31 +42,31 @@
...
+// CPPUNIT_TEST_SUITE_REGISTRATION(TestPlatformInconsistencyCheckerWorker);
+////  CPPUNIT_TEST_SUITE_REGISTRATION(TestConflictFinderWorker);
+// CPPUNIT_TEST_SUITE_REGISTRATION(TestConflictResolverWorker);
+////  CPPUNIT_TEST_SUITE_REGISTRATION(TestOperationGeneratorWorker);
+////  CPPUNIT_TEST_SUITE_REGISTRATION(TestOperationSorterWorker);
+//
+//// CPPUNIT_TEST_SUITE_REGISTRATION(TestOldSyncDb); // Needs a pre 3.3.4 DB
+// CPPUNIT_TEST_SUITE_REGISTRATION(TestExecutorWorker);
+// CPPUNIT_TEST_SUITE_REGISTRATION(TestSyncPal);
+//// CPPUNIT_TEST_SUITE_REGISTRATION(TestIntegration);
```
herve-er, line 69: I guess this is not wanted 😉

PR #267, opened 2024-08-23
[KDESKTOP-1149] Do not delete a local item if the remote counterpart exists

luc-guyot-infomaniak:
This pull-request addresses an issue with the method `LocalDeleteJob::canRun`: the comparison of the local and remote relative paths is incorrect and can cause unwanted files deletions.

The fix consist in comparing 
- a local relative path against a remote relative path for a regular synchronisation,
-  a local relative path against the endings of a remote relative path for an **advanced synchronisation**.


src/libsyncengine/syncpal/syncpal.cpp
```diff
@@ -67,68 +66,69 @@ SyncPal::SyncPal(const SyncPath &syncDbPath, const std::string &version, bool ha
     createSharedObjects();
 }
 
-SyncPal::SyncPal(int syncDbId, const std::string &version) : _logger(Log::instance()->getLogger()), _syncDbId(syncDbId) {
+SyncPal::SyncPal(const int syncDbId_, const std::string &version) : _logger(Log::instance()->getLogger()) {
```
ClementKunz, line 69: The '_' at the end of the parameter's name is not the usual naming convention that we use. Is it voluntary?

ClementKunz (approved):

src/libsyncengine/syncpal/syncpal.cpp
luc-guyot-infomaniak: This the way I usually resolve the name conflict between the input parameter and the method name (accessor). 
Would you suggest something else?

ClementKunz: No, this is fine for me then 👍

PR #273, opened 2024-08-27
Kdesktop 1171 implement user information collection for non add error sentries

herve-er:
# Description

Currently, Sentry events that do not originate from an "add error" do not include user information. This limitation prevents these events from being sorted and grouped by user, making it difficult to determine whether a large number of issues affect a diverse user base or a smaller subset of users experiencing recurring problems.

### DIFF
--- a/src/server/appserver.cpp
+++ b/src/server/appserver.cpp
@@ -309,6 +309,8 @@ AppServer::AppServer(int &argc, char **argv)
         return;
     }
 
+    // Configure sentry users for better identification
+    setDefaultSentryUsers();
 
     // Start syncs
     QTimer::singleShot(0, [=]() { startSyncPals(); });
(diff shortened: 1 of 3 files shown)


src/server/appserver.cpp
```diff
@@ -467,6 +469,57 @@ void AppServer::logExtendedLogActivationMessage(bool isExtendedLogEnabled) noexc
...
+    std::string userName = "No user in db";
+    std::string userEmail = "No user in db";
+    std::string allUsersIds = "No user in db";
+
+    if (!userList.empty()) {
+        userId = std::to_string(userList[0].userId());
+        userName = userList[0].name();
+        userEmail = userList[0].email();
+        for (const auto &user : userList) {
+            if (user.userId() != userList[0].userId()) {
```
luc-guyot-infomaniak, line 486: If the list can contain duplicate entries, shouldn't we use a set to remove duplicates?

src/server/appserver.cpp
```diff
@@ -467,6 +469,57 @@ void AppServer::logExtendedLogActivationMessage(bool isExtendedLogEnabled) noexc
...
+                allUsersIds += " | ";
```
luc-guyot-infomaniak, line 487: Consider `std::accumulate` for such concatenation operations.

luc-guyot-infomaniak (commented):
Comments about duplicate user management and a secure mechanism to set Sentry user back to default.

PR #287, opened 2024-09-06
KDESKTOP-1192-Files-downloaded-at-creation-in-LiteSync-mode

ClementKunz:
All files created on remote while the app was running were detected as link. Therefore, the app would download the file content even in LiteSync mode.

### DIFF
--- a/src/libsyncengine/update_detection/file_system_observer/remotefilesystemobserverworker.cpp
+++ b/src/libsyncengine/update_detection/file_system_observer/remotefilesystemobserverworker.cpp
@@ -44,8 +44,8 @@
 namespace KDC {
 
 RemoteFileSystemObserverWorker::RemoteFileSystemObserverWorker(std::shared_ptr<SyncPal> syncPal, const std::string &name,
-                                                               const std::string &shortName)
-    : FileSystemObserverWorker(syncPal, name, shortName, ReplicaSide::Remote), _driveDbId(syncPal->driveDbId()) {}
+                                                               const std::string &shortName) :
+    FileSystemObserverWorker(syncPal, name, shortName, ReplicaSide::Remote), _driveDbId(syncPal->driveDbId()) {}
 
 RemoteFileSystemObserverWorker::~RemoteFileSystemObserverWorker() {
     LOG_SYNCPAL_DEBUG(_logger, "~RemoteFileSystemObserverWorker");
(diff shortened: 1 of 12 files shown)


src/libsyncengine/update_detection/file_system_observer/remotefilesystemobserverworker.cpp
```diff
@@ -565,6 +565,12 @@ ExitCode RemoteFileSystemObserverWorker::extractActionInfo(const Poco::JSON::Obj
         }
     }
 
+    std::string tmp;
+    if (!JsonParserUtility::extractValue(actionObj, symbolicLinkKey, tmp, false)) {
+        return ExitCode::BackError;
+    }
+    actionInfo.isLink = !tmp.empty();
```
ClementKunz, line 572: This is the main change.

src/libsyncengine/update_detection/file_system_observer/remotefilesystemobserverworker.cpp
```diff
@@ -578,7 +584,7 @@ ExitCode RemoteFileSystemObserverWorker::extractActionInfo(const Poco::JSON::Obj
 ExitCode RemoteFileSystemObserverWorker::processAction(const SyncName &usedName, const ActionInfo &actionInfo,
                                                        std::set<NodeId, std::less<>> &movedItems) {
     SnapshotItem item(actionInfo.nodeId, actionInfo.parentNodeId, usedName, actionInfo.createdAt, actionInfo.modtime,
-                      actionInfo.type, actionInfo.size, actionInfo.canWrite);
+                      actionInfo.type, actionInfo.size, actionInfo.isLink, actionInfo.canWrite, true);
```
ClementKunz, line 587: And this is where the variable `isLink` was always set to `true` (at least set to the value of `canWrite`)

luc-guyot-infomaniak: Suggestion: define a constructor of `SnapshotItem` which task an object of type `ActionInfo` as argument.

luc-guyot-infomaniak (approved):

src/libsyncengine/update_detection/file_system_observer/remotefilesystemobserverworker.cpp
ClementKunz: Done, but not exactly as suggested. Since `ActionInfo` only exist in `RemoteFileSystemObserverWorker` I have preferred to include a `SnapshotItem` object directly inside `ActionInfo`

luc-guyot-infomaniak (approved):
Looks good to me.

PR #290, opened 2024-09-10
Fix Translation issue

herve-er:
In the log upload windows, the large folder text is not translated.

### DIFF
--- a/src/gui/debuggingdialog.h
+++ b/src/gui/debuggingdialog.h
@@ -63,6 +63,7 @@ class DebuggingDialog : public CustomDialog {
         bool _deleteLogs = false;
         bool _needToSave = false;
         bool _sendArchivedLogs = true;
+        QString _heavyLogLabelStr;
 
         void initUI();
         void initLogUploadLayout();
--- a/src/gui/debuggingdialog.cpp
+++ b/src/gui/debuggingdialog.cpp
@@ -46,10 +46,6 @@ static const int recordDebuggingBoxVMargin = 20;
 static const int debugLevelLabelBoxVMargin = 10;
 static const int debugLevelSelectBoxVMargin = 20;
 static const QString debuggingFolderLink = "debuggingFolderLink";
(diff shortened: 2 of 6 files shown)


src/gui/debuggingdialog.h
```diff
@@ -63,6 +63,7 @@ class DebuggingDialog : public CustomDialog {
         bool _deleteLogs = false;
         bool _needToSave = false;
         bool _sendArchivedLogs = true;
+        QString _heavyLogLabelStr;
```
luc-guyot-infomaniak, line 66: Would `static const` here  cause a new problem? What was the cause of the original problem?

luc-guyot-infomaniak (commented):
A question about qualifier of the newly introduced variable.

ClementKunz (approved):

src/gui/debuggingdialog.h
herve-er: Using a `static const`, means the string gets initialized when the app starts up. At that moment, `QTranslator` sets `heavyLogLabelStr` to English text because there's no language set yet. This leads to the text staying in English, even if you pick another language, because cause it is not being updated after that.

To solve this, we can call `QObject::tr` when we open the modal. 
At best, I can make it const and initialize it in the constructor, but using static const isn't working in this case.

src/gui/debuggingdialog.h
```diff
@@ -63,7 +63,12 @@ class DebuggingDialog : public CustomDialog {
         bool _deleteLogs = false;
         bool _needToSave = false;
         bool _sendArchivedLogs = true;
-
+        /* _heavyLogLabelStr can't be a static const because of tr needing to be called at each instantiation to apply the
+         *  current languag*
```
luc-guyot-infomaniak, line 67:
```suggestion
         *  current language*
```

luc-guyot-infomaniak (dismissed):

luc-guyot-infomaniak (dismissed):

ChristopheLarchier (approved):

PR #291, opened 2024-09-10
Kdesktop 1186 add unit test for kdesktop 1172 sentry wrapper

herve-er:
Add unit test for Sentry handler and adresse some issues.


src/libcommon/log/sentry/sentryhandler.h
```diff
@@ -136,5 +139,8 @@ class SentryHandler {
         std::unordered_map<std::string, SentryEvent, StringHash, std::equal_to<>> _events;
         SentryConfidentialityLevel _globalConfidentialityLevel = SentryConfidentialityLevel::Anonymous; // Default value
         SentryConfidentialityLevel _lastConfidentialityLevel = SentryConfidentialityLevel::None;
+        const int _sentryMaxCaptureCountBeforeRateLimit = 10; // Number of capture before rate limiting an event
```
luc-guyot-infomaniak, line 142: `static` class members ?

test/libcommon/log/sentry/testsentryhandler.cpp
```diff
@@ -0,0 +1,119 @@
...
+    CPPUNIT_ASSERT_EQUAL(2, mockSentryHandler.sentryUploadedEventCount());
+    mockSentryHandler.captureMessage(SentryLevel::Info, "Test",
+                                     "Test message"); // Rate limit reached, should be sent (first event on rate limit is sent).
+    CPPUNIT_ASSERT_EQUAL(3, mockSentryHandler.sentryUploadedEventCount());
+
+    for (int i = 3; i <= 4; i++) { // Done twice to check that the timer is well reset.
+        mockSentryHandler.captureMessage(SentryLevel::Info, "Test", "Test message"); // Rate limit reached, should not be sent
+        CPPUNIT_ASSERT_EQUAL(i, mockSentryHandler.sentryUploadedEventCount());
+
+        std::this_thread::sleep_for(std::chrono::seconds(1)); // Wait less than upload rate limit (2s in test)
```
luc-guyot-infomaniak, line 49: Can we lower the rate limit the minimum so as to speed up this test as much as possible?

test/libcommon/log/sentry/testsentryhandler.cpp
```diff
@@ -0,0 +1,119 @@
...
+
+        std::this_thread::sleep_for(std::chrono::seconds(1)); // Wait 1s more, the upload rate limit should be passed
+
+        mockSentryHandler.captureMessage(SentryLevel::Info, "Test",
+                                         "Test message"); // Should be sent
+        CPPUNIT_ASSERT_EQUAL(i + 1, mockSentryHandler.sentryUploadedEventCount());
+    }
+
+    std::this_thread::sleep_for(
+            std::chrono::seconds(3)); // Wait 4s witout capturing any event, it should be reset all rate limits
```
luc-guyot-infomaniak, line 62:
```suggestion
            std::chrono::seconds(3)); // Wait 4s without capturing any event. This should reset all rate limits.
```

luc-guyot-infomaniak: Same remark: can we decrease the delays?

src/libcommon/log/sentry/sentryhandler.cpp
```diff
@@ -115,10 +112,17 @@ void SentryHandler::captureMessage(SentryLevel level, const std::string &title,
...
-    sentry_capture_event(
-        sentry_value_new_message_event(static_cast<sentry_level_t>(event.level), event.title.c_str(), event.message.c_str()));
+    sendEventToSentry(event.level, event.title, event.message);
 }
 
+void SentryHandler::sendEventToSentry(const SentryLevel level, const std::string &title, const std::string &message) const {
+    sentry_capture_event(sentry_value_new_message_event(static_cast<sentry_level_t>(level), title.c_str(), message.c_str()));
+}
+
+SentryHandler::SentryHandler(int maxCaptureCountBeforeRateLimit, int minUploadIntervalOnRateLimit) :
```
luc-guyot-infomaniak, line 122: Because this constructor is only used in unit tests, I would define it in the mock class instead. (This improves intent clarity; this way no one is tempted to use it outside of this context).

src/libcommon/log/sentry/sentryhandler.h
luc-guyot-infomaniak: Now I understand that these constants are changed in the derived class for unit testing.

luc-guyot-infomaniak (commented):
Main comment: the place where the new constructor should be defined.

test/libcommon/log/sentry/testsentryhandler.cpp
herve-er: Here is a commit where I set the rate limit to 1s: [0da0908](https://github.com/Infomaniak/desktop-kDrive/pull/291/commits/0da0908d780918bef512ca89763594501931c988)

src/libcommon/log/sentry/sentryhandler.cpp
luc-guyot-infomaniak: > Doing this means _maxCaptureCountBeforeRateLimit and minUploadIntervalOnRateLimit will need to be protected members, which will trigger Sonar. 

I didn't know this would upset Sonar. My feeling is that allowing data members to be only `protected` is less confusing or noisy on the implementation side than a constructor dedicated to unit tests only. What would Sonar Cloud think about making the test class class friend?

herve-er: Here are the details of why Sonar considers this a critical maintainability issue: https://rules.sonarsource.com/cpp/RSPEC-3656/?search=protected. 
Not tested yet, but I don't think Sonar would flag this, as it is already the case for some other classes (like `SyncPal`, for example). 

However, regardless of what I mentioned in my previous comment 👀, I am unsure whether it's better to use a `friend` declaration or to add a protected setter. It doesn't seem unreasonable to me to allow a derived class to modify these settings of the base class.

herve-er: Here it is 👍🏾: [2d8d492](https://github.com/Infomaniak/desktop-kDrive/pull/291/commits/2d8d492c2daa421185a7a41775c4b5b8a44d8b47)

PS: 
> looks better than the public alternative constructor

In the previous implementation the test specific constructor was protected

luc-guyot-infomaniak: Oh, yes, sorry. What troubled me most was the visibility in the base class of a constructor that should be used only in unit tests. Sorry if I wasn't clear.

luc-guyot-infomaniak (approved):

ClementKunz (approved):

PR #295, opened 2024-09-13
Fix crash in logging

ChristopheLarchier:


luc-guyot-infomaniak (commented):
Suggestion: define a single logging helper function and re-use 3 times. This reduces the risk of partially propagated modifications the next time someone modifies the log messages.

PR #296, opened 2024-09-13
Implement CustomLogWStream wich doesn't allow non wstring in LOGW_XXX

herve-er:
Currently, using LOGW_XXXX macros with non-wchar_t types leads to crashes that are difficult to diagnose and are not caught during compilation. A solution is required to prevent the misuse of these macros with incorrect types at compile-time.Prevent the use of LOGW_XXXX with non-wchar types

### DIFF
--- a/src/libcommonserver/log/log.h
+++ b/src/libcommonserver/log/log.h
@@ -42,7 +42,7 @@ namespace KDC {
 
 #define LOGW_DEBUG(logger, logEvent)                                                                      \
     {                                                                                                     \
-        std::wstringstream stream;                                                                        \
+        Log::CustomLogWStream stream;                                                                       \
         stream << logEvent;                                                                               \
         sentry_value_t crumb = sentry_value_new_breadcrumb(nullptr, Utility::ws2s(stream.str()).c_str()); \
         sentry_value_set_by_key(crumb, "level", sentry_value_new_string("debug"));                        \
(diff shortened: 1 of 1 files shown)


src/libcommonserver/log/log.h
```diff
@@ -42,7 +43,7 @@ namespace KDC {
 
 #define LOGW_DEBUG(logger, logEvent)                                                                      \
     {                                                                                                     \
-        std::wstringstream stream;                                                                        \
+        CustomLogWStream stream;                                                                       \
```
luc-guyot-infomaniak, line 46: This is a very nice improvement and we should definitely merge it soon.

**Just one caveat**: if we decide to remove the Sentry instructions (which makes sense to me; debug or info logs are not supposed to be sent to Sentry), then we don't need to instantiate a stream anymore. (Without a stream instance of the proper type, we cannot check string types at compile time anymore.)
 
I am also willing to go a step further, shortly afterwards: we can **remove all invocations of logging macros** in our code. 
Here is what I envision:
- Define a `LogStream` and a `LogWStream` object. Each object stores (composition) a stream of the correct type and a reference to the logger in use.
- Define `operator<<(LogStream& logStream, const std::string &str)` and `operator<<(LogWStream& logStream, const std::wstring &str)` as expected.
- Define an object `EndLine` and  `operator<<(LogStream& logStream, const EndLine &endl)` so that it calls the appropriate `LOG4CPLUS` macro, passing the internal stream to the logger; this stream has been filled with all the message segments accumulated up to now.
**Example.**
```
LogWStream(_logger) << L"This wide string will be checked at compile time" << SyncName2WStr(someName) << EndLine();
```

herve-er: Let's include this in the existing Jira card about refactoring the log: https://infomaniak.atlassian.net/jira/software/c/projects/KDESKTOP/boards/341/backlog?selectedIssue=KDESKTOP-775&text=log

luc-guyot-infomaniak (approved):

PR #301, opened 2024-09-18
Adresse maxPathLength issue on Windows.

herve-er:
# Description:
We encountered a case where the sync fails with the error "Node not found in DB". After investigating, the issue arises during the consolidation of temporary nodes because some nodes are not found in the database.

This happens when a user syncs a file with a short name located in a folder whose path length is between 247 and 259 characters.

The folder gets excluded because its size exceeds MAX_PATH_LENGTH_WIN_SHORT - 12 (see [Microsoft Documentation](https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation?tabs=registry) for more details on the -12), but the path of an item within this folder is less than MAX_PATH_LENGTH_WIN_SHORT, leading to a missing node.

# Steps to Reproduce:

Sync a file with a short name in a folder with a path length between 247 and 259 characters.

Observe that the folder is excluded due to its length exceeding MAX_PATH_LENGTH_WIN_SHORT - 12.

The file inside the folder is synced, causing the node to be missing and resulting in the error.

# Solution
We decided to set the same maxPathLenght for file and folder.


src/libcommon/utility/utility.cpp
```diff
@@ -652,23 +652,19 @@ size_t CommonUtility::maxPathLength() {
             _maxPathWin = MAX_PATH_LENGTH_WIN_SHORT;
         }
     }
-    return _maxPathWin;
-
+    // For folders in short path mode, it is MAX_PATH - 12
+    // (https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation?tabs=registry)
+    // We decided to apply this rule for files too. We could else encounter issues.
+    // If the path of a folder is > MAX_PATH - 12 and the path of a file in this folder is between MAX_PATH - 12 and MAX_PATH.
```
luc-guyot-infomaniak, line 658:
```suggestion
    // If the path length of a folder is > MAX_PATH - 12 and the path length of a file in this folder is between MAX_PATH - 12 and MAX_PATH.
```

luc-guyot-infomaniak: I don't understand how it is possible that `C:\Dir` can be longer than `C:\Dir\file.txt`.

src/libcommon/utility/utility.cpp
```diff
@@ -652,23 +652,19 @@ size_t CommonUtility::maxPathLength() {
...
+    // It would lead to a synced file in a folder that is not synced (excluded because of its path length).
```
luc-guyot-infomaniak, line 659:
```suggestion
    // It would lead to a synced file in a folder that is not synced (hence excluded because of its path length).
```

src/libcommon/utility/utility.cpp
herve-er: **Example Case:**

Let's consider `MAX_PATH` = 24. This means `FOLDER_MAX_PATH` = `MAX_PATH` - 12 = 12.

Now, imagine we have a folder named `C:\test1\test2`, which is 14 characters long. Since this length is greater than `FOLDER_MAX_PATH`, this folder is excluded.

Assuming there's a file named `foo.txt` in this folder, its full path would be `C:\test1\test2\foo.txt`, which is 22 characters long. This length is less than `MAX_PATH`, so the file isn't excluded and will be processed.

The problem arises when we try to synchronize `foo.txt` in a folder that doesn't exist in the subsequent step because we've excluded `C:\test1\test2`

src/libsyncengine/reconciliation/platform_inconsistency_checker/platforminconsistencycheckerutility.h
```diff
@@ -42,7 +42,7 @@ class PlatformInconsistencyCheckerUtility {
 #endif
         bool checkNameSize(const SyncName &name);
         bool checkReservedNames(const SyncName &name);
-        bool checkPathLength(size_t pathSize, NodeType type);
+        bool checkPathLength(size_t pathSize);
```
luc-guyot-infomaniak, line 45: What about `const ` for the method?

src/libcommon/utility/utility.cpp
luc-guyot-infomaniak: Thanks a lot, that's very clear !

ClementKunz (approved):

luc-guyot-infomaniak (approved):

PR #318, opened 2024-09-25
Crash in a LOGW due to an access to a pointer that can be null

herve-er:


src/libsyncengine/propagation/executor/executorworker.cpp
```diff
@@ -290,15 +290,17 @@ void ExecutorWorker::handleCreateOp(SyncOpPtr syncOp, std::shared_ptr<AbstractJo
 
             // Remove from update tree
             if (!affectedUpdateTree(syncOp)->deleteNode(syncOp->affectedNode())) {
-                LOGW_SYNCPAL_WARN(_logger, L"Error in UpdateTree::deleteNode: node name="
+                LOGW_SYNCPAL_WARN(_logger, L"Error in UpdateTree::deleteNode: affectedNode name="
                                                    << SyncName2WStr(syncOp->affectedNode()->name()).c_str());
```
luc-guyot-infomaniak, line 294: Should we be more careful about dereferencing `syncOp->affectedNode()`? Or is this paranoïa ?

src/libsyncengine/propagation/executor/executorworker.cpp
```diff
@@ -290,15 +290,17 @@ void ExecutorWorker::handleCreateOp(SyncOpPtr syncOp, std::shared_ptr<AbstractJo
...
                 hasError = true;
                 return;
             }
 
             if (!targetUpdateTree(syncOp)->deleteNode(syncOp->correspondingNode())) {
```
luc-guyot-infomaniak, line 299: Is there a well-understood scenario that leads to a `nullptr`? Do we understand why it is possible?
Should we handle a `nullptr` differently, e.g., by not trying to delete it in the first place?

src/libsyncengine/propagation/executor/executorworker.cpp
```diff
@@ -290,15 +290,17 @@ void ExecutorWorker::handleCreateOp(SyncOpPtr syncOp, std::shared_ptr<AbstractJo
...
-                LOGW_SYNCPAL_WARN(_logger, L"Error in UpdateTree::deleteNode: node name="
-                                                   << SyncName2WStr(syncOp->correspondingNode()->name()).c_str());
+                std::wstring nodeName = syncOp->correspondingNode() ? SyncName2WStr(syncOp->correspondingNode()->name())
```
luc-guyot-infomaniak, line 300: This could be isolated in a tiny helper method (logging commands are really verbose!).

ClementKunz (dismissed):

src/libsyncengine/propagation/executor/executorworker.cpp
luc-guyot-infomaniak: Thanks. No need to change anything if there is no good reason to do so.

PR #320, opened 2024-09-25
Kdesktop 1248 crash in node delete children function

ChristopheLarchier:


src/libsyncengine/jobs/jobmanager.cpp
```diff
@@ -202,14 +206,18 @@ void JobManager::run() noexcept {
...
+            }
         } else {
-            LOG_DEBUG(Log::instance()->getLogger(),
-                      "Starting job " << nextJob.first->jobId() << " with priority " << nextJob.second);
+            LOG_DEBUG(Log::instance()->getLogger(), "Starting job " << jobId << " with priority " << nextJob.second);
             Poco::ThreadPool::defaultPool().startWithPriority(nextJob.second, *nextJob.first);
-            _runningJobs.insert(nextJob.first->jobId());
+            if (_runningJobs.contains(jobId)) {
+                _runningJobs.insert(jobId);
+            }
```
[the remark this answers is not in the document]
ChristopheLarchier: Done

src/libsyncengine/jobs/jobmanager.cpp
```diff
@@ -202,14 +206,16 @@ void JobManager::run() noexcept {
 void JobManager::startJob(std::pair<std::shared_ptr<AbstractJob>, Poco::Thread::Priority> nextJob) {
     const std::scoped_lock lock(_mutex);
     try {
+        UniqueId jobId = nextJob.first->jobId();
```
luc-guyot-infomaniak, line 209:
```suggestion
       const UniqueId jobId = nextJob.first->jobId();
```

src/libsyncengine/jobs/network/API_v2/abstracttokennetworkjob.cpp
```diff
@@ -97,6 +97,18 @@ void AbstractTokenNetworkJob::updateLoginByUserDbId(const Login &login, int user
     }
 }
 
+void AbstractTokenNetworkJob::clearCacheForUser(int userDbId) {
```
luc-guyot-infomaniak, line 100: As the intent seems to avoid exception, let's declare these method `no except`.
It might be useful to:
- `assert(false)` if the condition is not `true`,
- or log something as an error.

luc-guyot-infomaniak (commented):
As a general comment: if the goal is to make sure no exceptions escape the body of a number of functions, let us mark these functions as `no except`. 

There might be a performance issue depending on the map types: `contains` followed by `erase` performs two search operations. For `std::unordered_map` with keys of a fixed size, we get two times a constant. But for keys of variable sizes (`e.g., std::string`) and `std::map` the penalty is larger.

Possible approach:

```
size_t tryToErase(MapType &someMap, const KeyType &key) noexcept {
   if (auto it = someMap.find(key) != someMap.cend()) {
       someMap.erase(it); // noexcept, constant time.
      return 1;
} 
return 0;
}
```

src/libsyncengine/jobs/jobmanager.cpp
ChristopheLarchier: Done

ChristopheLarchier:
> As a general comment: if the goal is to make sure no exceptions escape the body of a number of functions, let us mark these functions as `no except`.
> 
> There might be a performance issue depending on the map types: `contains` followed by `erase` performs two search operations. For `std::unordered_map` with keys of a fixed size, we get two times a constant. But for keys of variable sizes (`e.g., std::string`) and `std::map` the penalty is larger.
> 
> Possible approach:
> 
> ```
> size_t tryToErase(MapType &someMap, const KeyType &key) noexcept {
>    if (auto it = someMap.find(key) != someMap.cend()) {
>        someMap.erase(it); // noexcept, constant time.
>       return 1;
> } 
> return 0;
> }
> ```

Done

src/libsyncengine/jobs/network/API_v2/abstracttokennetworkjob.h
```diff
@@ -43,8 +43,8 @@ class AbstractTokenNetworkJob : public AbstractNetworkJob {
 
         static void updateLoginByUserDbId(const Login &login, int userDbId);
 
-        inline static void clearCacheForUser(int userDbId) { _userToApiKeyMap.erase(userDbId); }
-        inline static void clearCacheForDrive(int driveDbId) { _driveToApiKeyMap.erase(driveDbId); }
+        static void clearCacheForUser(int userDbId);
```
luc-guyot-infomaniak, line 46: No except?

src/libsyncengine/jobs/network/API_v2/abstracttokennetworkjob.cpp
ChristopheLarchier: Done

ChristopheLarchier: Done

PR #321, opened 2024-09-25
Kdesktop 1251 Some sentry are not readable

herve-er:


src/libcommon/log/customlogwstream.h
```diff
@@ -26,19 +26,43 @@ class CustomLogWStream : private std::wstringstream {
...
+            std::wstringstream::operator<<(ul);
+            return *this;
+        }
+        CustomLogWStream &operator<<(double d) {
+            std::wstringstream::operator<<(d);
+            return *this;
+        }
         CustomLogWStream &operator<<(const std::wstring &str) {
-            std::wstringstream::operator<<(str.c_str());
+            CustomLogWStream::operator<<(str.c_str());
```
luc-guyot-infomaniak, line 65: Is it useful or required to call `c_str()`?

herve-er: As for CustomLogWStream &operator<<(const wchar_t *str), the operator is defined outside of std::wstringstream. Here is a commit where I used the same workaround for both CustomLogWStream &operator<<(const wchar_t *str) and CustomLogWStream &operator<<(const std::wstring &str).

herve-er: [893c0da](https://github.com/Infomaniak/desktop-kDrive/pull/321/commits/893c0da73b65eade834cfb5c0fb4eba7b42ab79f)

src/libcommon/log/customlogwstream.h
```diff
@@ -26,19 +26,45 @@ class CustomLogWStream : private std::wstringstream {
...
-            // inheritance
             static_cast<std::wstringstream &>(*this) << str;
             return *this;
         }
+        CustomLogWStream &operator<<(const std::wstring &str) {
+            static_cast<std::wstringstream &>(*this) << str;
+            return *this;
+        }
+
+        CustomLogWStream &operator<<(bool b) {
```
luc-guyot-infomaniak, line 42: Maybe for later: using `std::boolalpha` would print `true` and `false` instead of `1` and `0`.

PR #322, opened 2024-09-25
Locks read access to the terminated jobs queue to avoid race conditions

luc-guyot-infomaniak:


luc-guyot-infomaniak:
Closing because it does not seem to address the actual problem.

**Assumption.** The `Executor` is destroyed before every job is terminated. A residual job can then trigger the callback of a dead `Executor` which causes a crash.

PR #338, opened 2024-10-07
Kdesktop 1275 sync on fat system are broken

herve-er:
The `checkIfPathExist` function has been optimized for better performance. However, there is a known issue when calling this function with a path on a FAT32 disk in Visual Studio versions before 17.3 Preview 2. In such cases, the function returns invalid arguments due to a bug in Visual Studio (msvc). More details about this bug can be found in the following link: [std::filesystem::is\_symlink is broken on FAT](https://developercommunity.visualstudio.com/t/std::filesystem::is_symlink-is-broken-on/1638272).

This issue has a high impact as all sync operations on FAT32 disks are currently not working.

- [x] Manually run  & check results of  IoHelper tests on a FAT32/exFAT disk
- [x] Fix CheckIfPathExists on FAT32 Disk


src/libcommonserver/io/iohelper.cpp
```diff
@@ -598,6 +603,18 @@ bool IoHelper::checkIfPathExists(const SyncPath &path, bool &exists, IoError &io
         ioError = IoError::Success;
         return true;
     }
+#ifdef _WIN32 // TODO: Remove this block when we are migrating the release process to Visual Studio 2022
```
luc-guyot-infomaniak, line 606:
```suggestion
#ifdef _WIN32 // TODO: Remove this block when migrating the release process to Visual Studio 2022.
```

src/libcommonserver/io/iohelper.cpp
```diff
@@ -598,6 +603,18 @@ bool IoHelper::checkIfPathExists(const SyncPath &path, bool &exists, IoError &io
...
+    // Prior to Visual Studio 2022, std::filesystem::symlink_status would return a missleading InvalidArgument if the path is
+    // found but located on a FAT32 disk. If the file is not found, it works as expected. This behavior is fixed when compiling
+    // with VS2022. https://developercommunity.visualstudio.com/t/std::filesystem::is_symlink-is-broken-on/1638272
```
luc-guyot-infomaniak, line 609:
```suggestion
    // with VS2022, see https://developercommunity.visualstudio.com/t/std::filesystem::is_symlink-is-broken-on/1638272
```

src/libcommonserver/io/iohelper.cpp
```diff
@@ -598,6 +603,18 @@ bool IoHelper::checkIfPathExists(const SyncPath &path, bool &exists, IoError &io
...
+    if (ioError == IoError::InvalidArgument) {
+        if (Utility::isNtfs(path.parent_path()) ||
+            path.string().length() > CommonUtility::maxPathLength()) { // These are the normal cases for a InvalidArgument
```
luc-guyot-infomaniak, line 612:
```suggestion
            path.string().length() > CommonUtility::maxPathLength()) { // These are the normal cases for an InvalidArgument
```

luc-guyot-infomaniak: Are there other natural conditions associated to `InvalidArgument`? The length of file name? Special characters?

src/libcommonserver/io/iohelper.cpp
```diff
@@ -284,8 +285,12 @@ bool IoHelper::getItemType(const SyncPath &path, ItemType &itemType) noexcept {
     const bool isSymlink = _isSymlink(path, ec);
 
     itemType.ioError = stdError2ioError(ec);
+#ifndef WIN32
     const bool fsSupportsSymlinks =
             itemType.ioError != IoError::InvalidArgument; // If true, we assume that the file system in use does support symlinks.
+#else
+    const bool fsSupportsSymlinks = Utility::isNtfs(path.parent_path());
```
luc-guyot-infomaniak, line 292: What is the plan in terms of unit tests? (Another PR?) The fact that we are using static methods in helper classes make them difficult to test through mocking.

src/libcommonserver/io/iohelper.cpp
```diff
@@ -284,8 +285,12 @@ bool IoHelper::getItemType(const SyncPath &path, ItemType &itemType) noexcept {
     const bool isSymlink = _isSymlink(path, ec);
 
     itemType.ioError = stdError2ioError(ec);
+#ifndef WIN32
```
ClementKunz, line 288: You did not use the same macro as below (WIN32 vs _WIN32). Is there a particular reason£? If not, it would be great to uniformize.
Also avoid, when it is possible, the negation. Use 'ifdef' instead on 'ifndef'

src/libcommonserver/io/iohelper.cpp
herve-er: Here is a revised implementation to avoid overrding `InvalidArgument` errors. [5125d61](https://github.com/Infomaniak/desktop-kDrive/pull/338/commits/5125d61f0b912e0ddf4ec793fa166b1a8b2ffc1e)

src/libcommonserver/io/iohelper.cpp
herve-er: No particular reason, here are the corrections:
- For _WIN32 [ff0ce59](https://github.com/Infomaniak/desktop-kDrive/pull/338/commits/ff0ce5938bce3875ef902d238be29ca77bf5b8a8) 
- Use of ifdef instead of ifndef [d120287](https://github.com/Infomaniak/desktop-kDrive/pull/338/commits/d12028727708e33c9975d589839df07ae9ccc015)

src/libcommonserver/io/iohelper.cpp
```diff
@@ -598,6 +603,20 @@ bool IoHelper::checkIfPathExists(const SyncPath &path, bool &exists, IoError &io
         ioError = IoError::Success;
         return true;
     }
+#ifdef _WIN32 // TODO: Remove this block when migrating the release process to Visual Studio 2022.
+    // Prior to Visual Studio 2022, std::filesystem::symlink_status would return a missleading InvalidArgument if the path is
```
luc-guyot-infomaniak, line 607:
```suggestion
    // Prior to Visual Studio 2022, std::filesystem::symlink_status would return a misleading InvalidArgument if the path is
```

luc-guyot-infomaniak (approved):

src/libcommonserver/io/iohelper.cpp
```diff
@@ -284,8 +285,12 @@ bool IoHelper::getItemType(const SyncPath &path, ItemType &itemType) noexcept {
     const bool isSymlink = _isSymlink(path, ec);
 
     itemType.ioError = stdError2ioError(ec);
+#ifdef _WIN32
+    const bool fsSupportsSymlinks = Utility::isNtfs(path.parent_path());
```
ChristopheLarchier, line 289: Use remove_filename() instead of parent_path() ?

src/libcommonserver/io/iohelper.cpp
```diff
@@ -598,6 +603,20 @@ bool IoHelper::checkIfPathExists(const SyncPath &path, bool &exists, IoError &io
         ioError = IoError::Success;
         return true;
     }
+#ifdef _WIN32 // TODO: Remove this block when migrating the release process to Visual Studio 2022.
+    // Prior to Visual Studio 2022, std::filesystem::symlink_status would return a missleading InvalidArgument if the path is
+    // found but located on a FAT32 disk. If the file is not found, it works as expected. This behavior is fixed when compiling
+    // with VS2022, see https://developercommunity.visualstudio.com/t/std::filesystem::is_symlink-is-broken-on/1638272
+    if (ioError == IoError::InvalidArgument && !Utility::isNtfs(path)) {
```
ChristopheLarchier, line 610: path.remove_filename() ?

herve-er: `isNtfs `only use `SyncPath::root_name()` to determine the filesystem, there is no need to remove the filename.
```
bool Utility::isNtfs(const SyncPath &dirPath) {
    auto it = rootFsTypeMap.find(dirPath.root_name().string());
    if (it == rootFsTypeMap.end()) {
        std::string fsType = Utility::fileSystemName(dirPath);
        auto val = rootFsTypeMap.insert({dirPath.root_name().string(), fsType == NTFS});
        if (!val.second) {
            // Failed to insert into map
            return false;
        }
        it = val.first;
    }
    return it->second;
}
```

src/libcommonserver/io/iohelper.cpp
herve-er: I should just provide `path` to `Utility::isNtfs`. In a first time, I used` parent_path()` because of `Utility:: isNtfs (const SyncPath &dirPath)`  definition which suggest it is expecting a directory. But as it internally only relies on `SyncPath::root_name()` there is no need to remove the file name.
Could be interesting to change `Utility:: isNtfs (const SyncPath &dirPath)` declaration.

herve-er: Here is the commit with the modifications: [b3028ae](https://github.com/Infomaniak/desktop-kDrive/pull/338/commits/b3028ae8c60abc4c121048dd2e4fc1f155486763)

ChristopheLarchier (approved):

PR #345, opened 2024-10-09
Kdesktop 1260 Fix a crash when we fail to create a VFS.

herve-er:


src/server/appserver.cpp
```diff
@@ -3667,8 +3670,8 @@ ExitCode AppServer::createAndStartVfs(const Sync &sync, ExitCause &exitCause) no
         QString error;
         _vfsMap[sync.dbId()] = KDC::createVfsFromPlugin(sync.virtualFileMode(), vfsSetupParams, error);
```
ChristopheLarchier, line 3671: I would rather do it like this:
        auto vfsPtr = KDC::createVfsFromPlugin(sync.virtualFileMode(), vfsSetupParams, error);
        if (vfsPtr) {
             _vfsMap[sync.dbId()] = vfsPtr;
        } else { ...

src/server/appserver.cpp
```diff
@@ -2928,6 +2932,9 @@ ExitCode AppServer::startSyncs(User &user, ExitCause &exitCause) {
                         }
                     }
                 }
+                // Clear old errors for this sync
```
ChristopheLarchier, line 2935: Why move this code here and not leave it in initSyncPal?

PR #360, opened 2024-10-25
KDESKTOP-1244-Sync-loop-when-trying-to-upload-a-file-folder-to-a-directory-for-which-the-user-does-not-have-rights

ClementKunz:
When a file was created in a local folder with read-only access right, the newly created file was not correctly excluded.
Moreover, the access rights were evaluated in the Executor by reading them directly in the remote snapshot, which should never be the case.
This PR addresses these issues by trying to upload the file anyway and blacklisting it if the backend returns an error (403).


src/libsyncengine/propagation/executor/executorworker.cpp
```diff
@@ -1629,7 +1629,7 @@ bool ExecutorWorker::hasRight(SyncOpPtr syncOp, bool &exists) {
                     return false;
                 }
 
-                return _syncPal->_remoteSnapshot->canWrite(*newCorrespondingParentNode->id());
+                return true;
```
luc-guyot-infomaniak, line 1632: The methods impacted by your changes are essentially consuming a `SyncOp` object. Therefore they should be easily unit-testable. However, because of the logging operations and macros, the methods require more data to be actually tested. A small refactoring could help. Do you think it's manageable given our timeline?

ClementKunz: Done in 6a06e7fa011a0bb5cf5f7e4c76b32f03e8583b6a

luc-guyot-infomaniak (commented):
Thanks for the tests ! Ready to approve once the Windows build is green again.

luc-guyot-infomaniak (approved):

ChristopheLarchier (approved):

ClementKunz:
Following https://github.com/Infomaniak/desktop-kDrive/pull/367, this PR in not relevant anymore

PR #364, opened 2024-10-30
KDESKTOP-1325-kStore-Send-user-ID-when-checking-for-new-update

ClementKunz:
Add user IDs as query parameter to the GetAppVersion request.
Also log the app ID and all user IDs at start up.

### DIFF
--- a/src/libsyncengine/jobs/network/getappversionjob.cpp
+++ b/src/libsyncengine/jobs/network/getappversionjob.cpp
@@ -40,7 +40,8 @@ static const std::string buildVersionKey = "build_version";
 static const std::string buildMinOsVersionKey = "build_min_os_version";
 static const std::string downloadUrlKey = "download_link";
 
-GetAppVersionJob::GetAppVersionJob(const Platform platform, const std::string &appID) : _platform(platform), _appId(appID) {
+GetAppVersionJob::GetAppVersionJob(const Platform platform, const std::string &appID, const std::list<int> &userIdList) :
+    _platform(platform), _appId(appID), _userIdList(userIdList) {
     _httpMethod = Poco::Net::HTTPRequest::HTTP_GET;
 }
 
@@ -67,6 +68,12 @@ std::string GetAppVersionJob::getSpecificUrl() {
 std::string GetAppVersionJob::getContentType(bool &canceled) {
     canceled = false;
     return {};
+}
+
+void GetAppVersionJob::setQueryParameters(Poco::URI &uri, bool &canceled) {
+    for (const auto id: _userIdList) {
+        uri.addQueryParameter("user_ids[]", std::to_string(id));
+    }
 }
 
 bool GetAppVersionJob::handleError(std::istream &, const Poco::URI &uri) {
--- a/src/libsyncengine/jobs/network/getappversionjob.h
+++ b/src/libsyncengine/jobs/network/getappversionjob.h
@@ -25,7 +25,7 @@
 namespace KDC {
 class GetAppVersionJob : public AbstractNetworkJob {
     public:
-        GetAppVersionJob(Platform platform, const std::string &appID);
+        GetAppVersionJob(Platform platform, const std::string &appID, const std::list<int> &userIdList);
         ~GetAppVersionJob() override = default;
 
         const VersionInfo &getVersionInfo(const DistributionChannel channel) { return _versionInfo[channel]; }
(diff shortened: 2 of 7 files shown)


src/libsyncengine/jobs/network/getappversionjob.cpp
```diff
@@ -69,6 +70,12 @@ std::string GetAppVersionJob::getContentType(bool &canceled) {
     return {};
 }
 
+void GetAppVersionJob::setQueryParameters(Poco::URI &uri, bool &canceled) {
+    for (const auto id: _userIdList) {
```
luc-guyot-infomaniak, line 74:
```suggestion
    for (const auto &id: _userIdList) {
```

test/server/updater/testupdatechecker.cpp
```diff
@@ -34,7 +34,7 @@ static const std::string smallVersionJsonUpdateStr =
 class GetAppVersionJobTest final : public GetAppVersionJob {
     public:
         GetAppVersionJobTest(const Platform platform, const std::string &appID, const bool updateShouldBeAvailable) :
-            GetAppVersionJob(platform, appID), _updateShouldBeAvailable(updateShouldBeAvailable) {}
+            GetAppVersionJob(platform, appID, {}), _updateShouldBeAvailable(updateShouldBeAvailable) {}
```
luc-guyot-infomaniak, line 37: Would it be more natural to set the user id list with at least one element?

src/libsyncengine/jobs/network/getappversionjob.h
```diff
@@ -40,16 +40,15 @@ class GetAppVersionJob : public AbstractNetworkJob {
...
-        }
+        void setQueryParameters(Poco::URI &uri, bool &canceled) override;
+        void setData(bool & /*canceled*/) override { /* no body parameters */ }
         bool handleError(std::istream &is, const Poco::URI &uri) override;
 
         [[nodiscard]] DistributionChannel toDistributionChannel(const std::string &val) const;
 
         const Platform _platform{Platform::Unknown};
         const std::string _appId;
+        const std::list<int> _userIdList;
```
luc-guyot-infomaniak, line 51: **Suggestion.** Use `std::vector` because the interface is simpler and accessing its element is usually faster. Plus, this is what we already use for the user list.

test/server/updater/testupdatechecker.cpp
ClementKunz: I can change it, but it will not impact the test result. Moreover, no user ID is a valid case. For example, first installation where the user has not logged in yet

src/libsyncengine/jobs/network/getappversionjob.h
ClementKunz: Agreed! Done in c6700de4a937df1ac5a2c4300b373e5c7e1a31c1

src/libsyncengine/jobs/network/getappversionjob.cpp
ClementKunz: Done in c6700de4a937df1ac5a2c4300b373e5c7e1a31c1

test/server/updater/testupdatechecker.cpp
luc-guyot-infomaniak: It was not clear to me that an empty list isa valid use case; thanks for the clarification.

 Would it be possible then to use a version of the `GetAppVersionJob` constructor that does take the list as argument? 

The reason of my initial comment: the unit tests can also serve the purpose of code documentation. They ideally reflect actual use cases.

ClementKunz: Done in 95eeadc19f2f1300de7f391be3d7c5ddabd88413

src/server/appserver.cpp
```diff
@@ -3022,6 +3025,25 @@ bool AppServer::initLogging() noexcept {
...
+    }
+    const auto &appUid = std::get<std::string>(appStateValue);
+    LOG_INFO(Log::instance()->getLogger(), "App ID: " << appUid.c_str());
+
+    // Log user IDs
+    std::vector<User> userList;
+    if (!ParmsDb::instance()->selectAllUsers(userList)) {
+        LOG_WARN(Log::instance()->getLogger(), "Error in ParmsDb::selectAllUsers");
+    }
+    std::list<int> userIdList;
```
ChristopheLarchier, line 3042: Unused variable?

src/server/appserver.cpp
```diff
@@ -3022,6 +3025,25 @@ bool AppServer::initLogging() noexcept {
...
+    if (bool found = false; !ParmsDb::instance()->selectAppState(AppStateKey::AppUid, appStateValue, found) || !found) {
+        LOG_WARN(Log::instance()->getLogger(), "Error in ParmsDb::selectAppState");
+    }
+    const auto &appUid = std::get<std::string>(appStateValue);
+    LOG_INFO(Log::instance()->getLogger(), "App ID: " << appUid.c_str());
+
+    // Log user IDs
+    std::vector<User> userList;
+    if (!ParmsDb::instance()->selectAllUsers(userList)) {
+        LOG_WARN(Log::instance()->getLogger(), "Error in ParmsDb::selectAllUsers");
```
ChristopheLarchier, line 3040: LOG_INFO?

src/server/appserver.cpp
ClementKunz: Indeed thx

src/server/appserver.cpp
```diff
@@ -3022,6 +3025,24 @@ bool AppServer::initLogging() noexcept {
                                .c_str());
     return true;
 }
+void AppServer::logUsefulInformation() const {
+    // Log app ID
```
herve-er, line 3029: It might be interesting to consider moving other useful information here (such as extended logs).

ClementKunz: Done in d12235a68d1f80b5fdbc84aee029baafac34ebcd

herve-er (approved):

ChristopheLarchier (approved):

PR #366, opened 2024-10-31
[KDESKTOP-1262] Do not force file status after an upload session abortion

luc-guyot-infomaniak:
This pull request aims at addressing [KDESKTOP-1262](https://infomaniak.atlassian.net/browse/KDESKTOP-1262):

- Reset the `vfsForceStatus` with `nullptr` after a drive upload session abortion.
- For MacOSX only, reset the extended attribute of the local file that failed to be uploaded.

[KDESKTOP-1262]: https://infomaniak.atlassian.net/browse/KDESKTOP-1262?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/libsyncengine/jobs/network/API_v2/upload_session/driveuploadsession.cpp
```diff
@@ -109,4 +111,26 @@ bool DriveUploadSession::handleCancelJobResult(const std::shared_ptr<UploadSessi
...
+#ifdef __APPLE__
+    // Removing all extended attributes so that the file that failed to be uploaded is not identified as a dehydrated placeholder
+    // by the application when restarting the synchronization.
+
+    const SyncPath &localPath = getFilePath();
+    static const std::vector<const char *> attributes = {EXT_ATTR_STATUS, EXT_ATTR_PIN_STATE};
+
+    for (const auto attribute: attributes) {
+        if (auto ioError = IoError::Success;
+            !IoHelper::removeXAttr(localPath, EXT_ATTR_STATUS, ioError) || ioError != IoError::NoSuchFileOrDirectory) {
```
luc-guyot-infomaniak, line 128: **Questions.**
- Is a missing permission an actual problem for the user?
- In case of unexpected failure, should the exit code be changed?

src/libsyncengine/jobs/network/API_v2/upload_session/driveuploadsession.cpp
```diff
@@ -109,4 +111,26 @@ bool DriveUploadSession::handleCancelJobResult(const std::shared_ptr<UploadSessi
...
+    setVfsForceStatusCallback(nullptr);
+
+#ifdef __APPLE__
+    // Removing all extended attributes so that the file that failed to be uploaded is not identified as a dehydrated placeholder
+    // by the application when restarting the synchronization.
+
+    const SyncPath &localPath = getFilePath();
+    static const std::vector<const char *> extendedAttributes = {EXT_ATTR_STATUS, EXT_ATTR_PIN_STATE};
+
+    for (const auto attribute: extendedAttributes) {
```
luc-guyot-infomaniak, line 126: **Questions.**
- Should we remove all attributes or just `EXT_ATTR_STATUS`? (What is the definition of a placeholder for the application?)
- Is a missing permission an actual problem for the user?
- In case of unexpected failure, should the exit code be changed?

ClementKunz: A file is a placeholder if it has the `EXT_ATTR_STATUS` extended attribute (see `LiteSyncExtConnector::vfsGetStatus`). So it should be enough to remove just this one. However, there is no point to keep only the pin state.
Moreover, the definition of a placeholder based solely on the status is also arguable. Perhaps basing it on the pinstate instead could help us display the correct status in Finder without risking to allow the deshydration of a not yet synced item.

ClementKunz: > In case of unexpected failure, should the exit code be changed?

Maybe again ^^ but what can we do if the access rights have been removed 🥲

src/libsyncengine/jobs/network/API_v2/upload_session/driveuploadsession.cpp
```diff
@@ -109,4 +111,27 @@ bool DriveUploadSession::handleCancelJobResult(const std::shared_ptr<UploadSessi
...
```
ClementKunz, line 126: This should be in a utility function outside this class. Especially because `normal`upload (meaning direct uploads, not upload session) are also concerned.

extensions/MacOSX/kDriveLiteSync/Extension/main.m
```diff
@@ -16,8 +16,8 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 
+#import "../../../../src/libcommonserver/io/fileAttributes.h"
```
ChristopheLarchier, line 19: Move the common declarations in vfs.h ?
Or create a new dynamic library "libcommonserverext" with the common code.

src/libsyncengine/jobs/network/API_v2/upload_session/driveuploadsession.cpp
```diff
@@ -109,4 +111,27 @@ bool DriveUploadSession::handleCancelJobResult(const std::shared_ptr<UploadSessi
...
+void DriveUploadSession::abort() {
+    AbstractUploadSession::abort();
+    setVfsForceStatusCallback(nullptr);
+
+#ifdef __APPLE__
+    // Removing all extended attributes so that the file that failed to be uploaded is not identified as a dehydrated placeholder
+    // by the application when restarting the synchronization.
+
+    const SyncPath &localPath = getFilePath();
+    static const std::vector<const char *> extendedAttributes = {EXT_ATTR_STATUS, EXT_ATTR_PIN_STATE};
```
ChristopheLarchier, line 124: Use the new function IoHelper::removeLiteSyncXAttrs

PR #374, opened 2024-11-05
Bug fix: setProgress  should not exit with an error when the item progress is not managed.

herve-er:
## Description 
Currently, calling ProgressInfo::setProgress on an item that is not in the _currentItems list returns an error. It should instead behave as if !shouldCountProgress and return successfully.

This is problematic in the case of omitted operations, which will systematically generate an error. This error prevent us to correctly set the vfs and status of the file.

### DIFF
--- a/src/libsyncengine/progress/progressinfo.cpp
+++ b/src/libsyncengine/progress/progressinfo.cpp
@@ -119,7 +119,9 @@ bool ProgressInfo::setProgress(const SyncPath &path, const int64_t completed) {
 
     const auto it = _currentItems.find(normalizedPath);
     if (it == _currentItems.end() || it->second.empty()) {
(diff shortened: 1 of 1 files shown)


ClementKunz (approved):

src/libsyncengine/progress/progressinfo.cpp
```diff
@@ -140,7 +142,9 @@ bool ProgressInfo::setProgressComplete(const SyncPath &path, const SyncFileStatu
 
     const auto it = _currentItems.find(normalizedPath);
```
luc-guyot-infomaniak, line 143: The two blocks looks identical.  Refactoring to remove redundancy would let us write a fix in a single place.

PR #382, opened 2024-11-08
KDESKTOP-1360-Issues-with-Sparkle-updater

ClementKunz:
Once the update has been accepted by the user, the package is downloaded by Sparkle and the user needs to start the update. If another update windows pops up before the validation, the next updates will fail.

Fix : Do not start a new update process if one is already ongoing

### DIFF
--- a/src/server/updater/sparkleupdater.mm
+++ b/src/server/updater/sparkleupdater.mm
@@ -192,6 +192,7 @@ void SparkleUpdater::onUpdateFound() {
         return;
     }
     if ([d->updater sessionInProgress]) {
+        // An update window is already opened or installation is in progress, no need to start a new one
         LOG_INFO(KDC::Log::instance()->getLogger(), "A session is already in progress...");
         return;
     }
--- a/src/server/updater/sparkleupdater.mm
+++ b/src/server/updater/sparkleupdater.mm
@@ -192,8 +192,7 @@ void SparkleUpdater::onUpdateFound() {
         return;
     }
     if ([d->updater sessionInProgress]) {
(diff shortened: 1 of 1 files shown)


src/server/updater/sparkleupdater.mm
```diff
@@ -187,7 +187,14 @@ - (BOOL)supportsGentleScheduledUpdateReminders {
 }
 
 void SparkleUpdater::onUpdateFound() {
-    if (isVersionSkipped(versionInfo().fullVersion())) return;
+    if (isVersionSkipped(versionInfo().fullVersion())) {
+        LOG_INFO(KDC::Log::instance()->getLogger(), "Version " << versionInfo().fullVersion().c_str() << " is skipped.");
+        return;
+    }
+    if ([d->updater sessionInProgress]) {
+        LOG_INFO(KDC::Log::instance()->getLogger(), "A session is already in progress...");
```
luc-guyot-infomaniak, line 195: Maybe a word about the fact the application is already **updating** itself?

ClementKunz: Done d7a9ea3030ec854e949b671630d16732b74a9ccc
We are waiting for user input or for the end of an ongoing installation

luc-guyot-infomaniak: Sorry, I was not precise enough: I thought the log message itself could  mention the ongoing update process, to ease off debugging.

ClementKunz: Done. I have removed the comment, no need to say the same thng twice

ChristopheLarchier (approved):

PR #390, opened 2024-11-13
Fix MoveToTrashFailed behaviour

herve-er:
# Descritpion
This PR changes the behavior of the "Move to Trash" option for file deletion:

When the "Move to Trash" option is enabled, files are typically moved to the trash.
New behavior: If moving the file to the trash fails, the file will be permanently deleted.
This PR also includes a disclaimer to inform the user that the file may be irreversibly deleted if the move to trash operation fails.

# UI/UX Modifications:
## Settings -> Preferences:
MoveToTrash_settings

### With "Move to Trash" enabled:
MoveToTrash_conflicts

### With "Move to Trash" disabled:
warning_delete_conflicts


herve-er:
Waiting FAQ

ChristopheLarchier (approved):

PR #400, opened 2024-11-21
Resolve Concurrent Access Issue in TerminatedJobsQueue

herve-er:
#### **Overview**
This PR addresses a critical issue in the `ExecutorWorker` class where concurrent access to the `TerminatedJobsQueue` caused a crash. It occurred during the execution of the `deleteFinishedAsyncJobs` method when multiple threads attempted to access `_terminatedJobs` without proper synchronization.

#### **Root Cause**
The root cause was identified as unsafe access to `_terminatedJobs`, a shared resource accessed concurrently by multiple threads. This led to race conditions and crashes, particularly when one thread was modifying the queue while another was reading or checking its contents (TerminatedJobsQueue::front()).

#### **Key Changes**
1. **Made `TerminatedJobsQueue` Thread-Safe:**
   - Converted `TerminatedJobsQueue` into a thread-safe container by inheriting from `std::recursive_mutex`.
   - Used `std::scoped_lock` for locking within all public methods of the queue (`push`, `pop`, `front`, `empty`) to ensure safe access across threads.

2. **Updated `deleteFinishedAsyncJobs` Logic:**
   - Added a `std::scoped_lock` within the `deleteFinishedAsyncJobs` method to ensure thread-safe access to `_terminatedJobs` and prevent that no item is inserted between a call to `TerminatedJobsQueue::first()` and `TerminatedJobsQueue::pop()`.

3. **Test Coverage:**
   - Implemented a new test, `testTerminatedJobsQueue`, in the `TestExecutorWorker` suite to verify that `TerminatedJobsQueue` operations (`push`, `pop`, `front`, `empty`) are thread-safe.

4. **Refactored `TimeOutChecker`:**
   - Introduced a new utility class, `TimeChecker`, to replace `TimeOutChecker` for cleaner and more precise timeout checks in test cases.

### DIFF
--- a/test/test_utility/timechecker.h
+++ b/test/test_utility/timechecker.h
@@ -1,0 +1,49 @@
+/*
+ * Infomaniak kDrive - Desktop
+ * Copyright (C) 2023-2024 Infomaniak Network SA
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+class TimeChecker {
+    public:
+        explicit TimeChecker(bool start = false) {
+            if (start) this->start();
+        }
+        void start() { _time = std::chrono::steady_clock::now(); }
+        void stop() {
+            auto end = std::chrono::steady_clock::now();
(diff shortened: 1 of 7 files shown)


ChristopheLarchier (approved):

test/test_utility/timechecker.h
```diff
@@ -0,0 +1,49 @@
...
+        explicit TimeChecker(bool start = false) {
+            if (start) this->start();
+        }
+        void start() { _time = std::chrono::steady_clock::now(); }
+        void stop() {
+            auto end = std::chrono::steady_clock::now();
+            _diff = std::chrono::duration_cast<std::chrono::milliseconds>(end - _time).count();
+        }
+        bool lessOrEqualThan(long long value) {
+            if (_diff > value) std::cout << "TimeChecker::lessThan: " << _diff << " >= " << value << std::endl;
```
luc-guyot-infomaniak, line 32: Would it be possible to pass logger argument to the constructor and forward output comments to this logger.

Printing debug information to `std::cout` clutters the unit tests outputs.

herve-er: This line is only printed when the test fails. 
CPPUNIT_ASSERT_LESS(expected, actual) print the same line.

luc-guyot-infomaniak (approved):

PR #411, opened 2024-11-27
Enhance FileAccessError on a parent detection.

herve-er:
If an application (e.g., Microsoft Money) creates a locked file, the localfilesystemobserver will not be able to read its attributes and will call handleAccessDenied.

Since the item is not present in the snapshot, handleAccessDenied will assume that the permission issue originates from the parent directory and will therefore blacklist the parent.

The problem is that in this case, this assumption is incorrect; the issue does not come from the parent directory.

When the parent directory is later unblacklisted, most files within it will be synced again, often leading to pseudo-create conflicts.

However, if one of the files in the folder has been modified, it will be blacklisted because it is not a pseudo conflict but a genuine one. This scenario is highly likely, as demonstrated in this specific case: Microsoft Money systematically modifies a file in the folder after generating a locked temporary file.


ChristopheLarchier (approved):

src/libsyncengine/syncpal/syncpal.cpp
```diff
@@ -1459,33 +1460,45 @@ ExitInfo SyncPal::handleAccessDeniedItem(const SyncPath &relativePath, ExitCause
...
 
 
     LOGW_SYNCPAL_DEBUG(_logger, L"Item " << Utility::formatSyncPath(relativePath) << L" (NodeId: " << Utility::s2ws(localNodeId)
                                          << L" is blacklisted temporarily because of a denied access.");
 
-    NodeId correspondingNodeId;
-    correspondingNodeId = snapshot(ReplicaSide::Remote)->itemId(relativePath);
-    if (bool found; correspondingNodeId.empty() &&
-                    !_syncDb->correspondingNodeId(ReplicaSide::Local, localNodeId, correspondingNodeId, found)) {
+    NodeId remoteNodeId;
```
luc-guyot-infomaniak, line 1483:
```suggestion
    NodeId remoteNodeId = snapshot(ReplicaSide::Remote)->itemId(relativePath);
```

src/libsyncengine/syncpal/syncpal.cpp
```diff
@@ -1470,33 +1471,45 @@ ExitInfo SyncPal::handleAccessDeniedItem(const SyncPath &relativePath, ExitCause
...
 
     LOGW_SYNCPAL_DEBUG(_logger, L"Item " << Utility::formatSyncPath(relativePath) << L" (NodeId: " << Utility::s2ws(localNodeId)
                                          << L" is blacklisted temporarily because of a denied access.");
 
-    NodeId correspondingNodeId;
-    correspondingNodeId = snapshot(ReplicaSide::Remote)->itemId(relativePath);
-    if (bool found; correspondingNodeId.empty() &&
-                    !_syncDb->correspondingNodeId(ReplicaSide::Local, localNodeId, correspondingNodeId, found)) {
+    NodeId remoteNodeId = snapshot(ReplicaSide::Remote)->itemId(relativePath);
+    remoteNodeId = snapshot(ReplicaSide::Remote)->itemId(relativePath);
```
ClementKunz, line 1495: Same line twice?

src/libsyncengine/syncpal/syncpal.cpp
```diff
@@ -1470,33 +1471,45 @@ ExitInfo SyncPal::handleAccessDeniedItem(const SyncPath &relativePath, ExitCause
...
     if (localNodeId = snapshot(ReplicaSide::Local)->itemId(relativePath); localNodeId.empty()) {
-        // The file does not exit yet on local file system, or we do not have sufficient right on a parent folder.
-        LOGW_DEBUG(_logger, L"Item " << Utility::formatSyncPath(relativePath)
-                                     << L"is not present local file system, blacklisting the parent item.");
-        return handleAccessDeniedItem(relativePath.parent_path(), cause);
+        SyncPath absolutePath = localPath() / relativePath;
+        if (!IoHelper::getNodeId(absolutePath, localNodeId)) {
+            bool exists = false;
+            IoError ioError = IoError::Success;
+            if (!IoHelper::checkIfPathExists(absolutePath, exists, ioError)) {
```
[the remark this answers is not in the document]
herve-er: `IoHelper::checkIfPathExists` is called only if `IoHelper::getNodeId` failed.

src/libsyncengine/syncpal/syncpal.cpp
herve-er:
```suggestion
    NodeId remoteNodeId = snapshot(ReplicaSide::Remote)->itemId(relativePath);
```

src/libsyncengine/syncpal/syncpal.cpp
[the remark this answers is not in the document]
ClementKunz: Indeed, sorry, I missed the first `!` 👓

PR #421, opened 2024-12-03
Kdesktop 1422 remaining encoding ms office issue 3.6.7

herve-er:


src/libsyncengine/propagation/executor/executorworker.cpp
```diff
@@ -2121,6 +2121,10 @@ ExitInfo ExecutorWorker::propagateEditToDbAndTree(SyncOpPtr syncOp, const NodeId
         return ExitCode::DataError;
     }
 
+    // In case of Delete+Create, the encoding might have changed. Therefor, we update the name anyway.
+    dbNode.setNameLocal(localName);
+    dbNode.setNameRemote(remoteName);
+
```
herve-er, line 2127: @ClementKunz Just a comment to mention that I have tested a Move + Edit, and it is working as expected; the Edit does not 'overwrite' the Move.

ChristopheLarchier (approved):

PR #424, opened 2024-12-06
KDESKTOP-1355-Renaming-a-folder-on-remote-side-should-trigger-an-explore-directory

ClementKunz:
If a directory is renamed on remote side so that an exclusion template does not apply anymore, the folder is correctly re-created on local side but not its content.

Steps to reproduce:

- Create a new exclusion template like that “C”
- Create a new directory on remote side called “C”
- Add something in “C”
- Rename “C” into “B”
====> “B” is correctly created on remote but is empty


src/libsyncengine/update_detection/file_system_observer/remotefilesystemobserverworker.cpp
```diff
@@ -623,11 +623,15 @@ ExitCode RemoteFileSystemObserverWorker::processAction(ActionInfo &actionInfo, s
         case ActionCode::actionCodeMoveIn:
         case ActionCode::actionCodeRestore:
         case ActionCode::actionCodeCreate:
+        case ActionCode::actionCodeRename: {
+            const bool exploreDir = actionInfo.snapshotItem.type() == NodeType::Directory &&
+                                    actionInfo.actionCode != ActionCode::actionCodeCreate &&
```
herve-er, line 628: Don't we still want explore the directory in the case of a `CREATE`
```suggestion
```

herve-er (approved):

PR #427, opened 2024-12-11
Kdesktop 1439 crash when displaying contextual menu in finder

ChristopheLarchier:


src/libsyncengine/syncpal/syncpal.cpp
```diff
@@ -1052,6 +1061,8 @@ ExitCode SyncPal::fileRemoteIdFromLocalPath(const SyncPath &path, NodeId &nodeId
 bool SyncPal::checkIfExistsOnServer(const SyncPath &path, bool &exists) const {
     exists = false;
 
+    if (!_remoteSnapshot) return false;
```
luc-guyot-infomaniak, line 1064: Can this be unit tested in a simple way?

ChristopheLarchier: Done

PR #432, opened 2024-12-12
Create a static library containing constants to be shared with extensions

luc-guyot-infomaniak:


luc-guyot-infomaniak:
Abandonned because of update issues caused by missing GPG commit signatures.

PR #449, opened 2024-12-23
Kdesktop 1463 mac os file content exchange issue

ChristopheLarchier:


test/libcommon/utility/testutility.cpp
```diff
@@ -299,6 +299,45 @@ void TestUtility::testCurrentVersion() {
...
+                err++;
+            }
+        }
+        CPPUNIT_ASSERT(err == 0);
+    }
+
+    {
+        int err = 0;
+        for (int c = 0; c < 100000; c++) {
+            int nb = 3;
```
ClementKunz, line 318: `static const`? Or just write 3 in the for loop

test/libcommon/utility/testutility.cpp
```diff
@@ -299,6 +299,45 @@ void TestUtility::testCurrentVersion() {
     CPPUNIT_ASSERT(std::regex_match(test, std::regex(R"(\d{1,2}\.{1}\d{1,2}\.{1}\d{1,2}\.{1}\d{0,8}$)")));
 }
 
+void TestUtility::testGenerateRandomStringAlphaNum() {
+    {
+        int err = 0;
+        std::set<std::string> results;
+        for (int i = 0; i < 100000; i++) {
+            std::string str = CommonUtility::generateRandomStringAlphaNum();
```
ClementKunz, line 307: The chance are probably insignificant, but since there is no guaranty of uniqueness from method `generateRandomStringAlphaNum`, this test might fail sometimes.

test/libcommon/utility/testutility.cpp
ChristopheLarchier: Done

ClementKunz (approved):

PR #450, opened 2024-12-27
KDESKTOP-1449-macOS-Finder-contextual-menu-Wording-issue

ClementKunz:
For folders, replace “Always make available locally” by “Make available locally”.


src/server/socketapi.cpp
```diff
@@ -1322,10 +1322,14 @@ void SocketApi::processFileList(const QStringList &inFileList, std::list<KDC::Sy
     }
 }
 
-QString SocketApi::vfsPinActionText() {
+QString SocketApi::vfsPinFileActionText() {
```
ChristopheLarchier, line 1325: Is it necessary to have 2 different messages for folders and files?

ClementKunz: Discussed together. Finally, we keep the same text for files and folders but remove the "Always" everywhere

herve-er (dismissed):

ClementKunz:
Looks like the change is already on `develop`. Nothing to do anymore...

PR #455, opened 2025-01-03
KDESKTOP-1451-Windows-Fix-Always-available-locally-folder-status

ClementKunz:
The pin state of a parent folder should be propagated to newly created child


ClementKunz:
Will be handled in another PR

PR #459, opened 2025-01-09
KDESKTOP-1477-Copyright-translation-is-wrong

ClementKunz:


translations/client_de.ts
```diff
@@ -29,7 +29,7 @@
     <message>
         <location filename="../src/gui/aboutdialog.cpp" line="124"/>
         <source>Copyright 2019-%1 Infomaniak Network SA&lt;br&gt;&lt;br&gt;</source>
-        <translation>Copyright 2019-2023 Infomaniak Network SA&lt;br&gt;&lt;br&gt; {2019-%1 ?}</translation>
+        <translation>Copyright 2019-%1 Infomaniak Network SA&lt;br&gt;&lt;br&gt;</translation>
```
luc-guyot-infomaniak, line 32: Regarding the following widget, I believe we need another change: 
copyright_2023

luc-guyot-infomaniak (commented):
Suggesting an additional change

translations/client_de.ts
ClementKunz: Done in 93748b23d707b4f566ae66def8cde05423c65008

ClementKunz:
> Why did some `svg` files change?
Looks like there is the same copyright inside the svg files as in the h/cpp files.

I can remove the changes on svg files if you like but it seemed pertinent to me since there is already an Infomaniak copyright inside

luc-guyot-infomaniak:
> > Why did some `svg` files change?
> > Looks like there is the same copyright inside the svg files as in the h/cpp files.
> 
> I can remove the changes on svg files if you like but it seemed pertinent to me since there is already an Infomaniak copyright inside

I am fine with it.

luc-guyot-infomaniak (approved):

PR #464, opened 2025-01-14
Do not use std::ranges

ClementKunz:
We must not use std::ranges until we are able to bump gcc version


src/gui/preferenceswidget.cpp
```diff
@@ -442,8 +442,13 @@ void PreferencesWidget::clearUndecidedLists() {
 }
 
 bool PreferencesWidget::isStaff() const {
-    constexpr auto isStaffCallback = [](std::pair<int, UserInfoClient> const &item) { return item.second.isStaff(); };
-    return std::ranges::find_if(_gui->userInfoMap(), isStaffCallback) != _gui->userInfoMap().end();
+    // To be used with an later gcc version
+    // constexpr auto isStaffCallback = [](std::pair<int, UserInfoClient> const &item) { return item.second.isStaff(); };
+    // return std::ranges::find_if(_gui->userInfoMap(), isStaffCallback) != _gui->userInfoMap().end();
+    for (const auto &[_, userInfo]: _gui->userInfoMap()) {
```
luc-guyot-infomaniak, line 448: What about `std::any_of`, see e.g.,  https://en.cppreference.com/w/cpp/algorithm/all_any_none_of ?

ClementKunz: Yes b2157ee5df8f454fe87fca765b7af07af38e9ee0

src/gui/preferenceswidget.cpp
```diff
@@ -443,7 +443,9 @@ void PreferencesWidget::clearUndecidedLists() {
 
 bool PreferencesWidget::isStaff() const {
     constexpr auto isStaffCallback = [](std::pair<int, UserInfoClient> const &item) { return item.second.isStaff(); };
-    return std::ranges::find_if(_gui->userInfoMap(), isStaffCallback) != _gui->userInfoMap().end();
+    // To be used with an later gcc version
+    // return std::ranges::find_if(_gui->userInfoMap(), isStaffCallback) != _gui->userInfoMap().end();
```
luc-guyot-infomaniak, line 447:
```suggestion
    // return std::ranges::any_of(_gui->userInfoMap().cbegin(), _gui->userInfoMap().cend(), isStaffCallback);
```

ChristopheLarchier (dismissed):

src/gui/preferenceswidget.cpp
ClementKunz: Why? Is `any_of` more performant?

luc-guyot-infomaniak: I just find `any_of` easier to understand. 

A possible implementation of `any_of` (see https://en.cppreference.com/w/cpp/algorithm/all_any_none_of) is
```
template<class InputIt, class UnaryPred>
constexpr bool any_of(InputIt first, InputIt last, UnaryPred p)
{
    return [std::find_if](http://en.cppreference.com/w/cpp/algorithm/find)(first, last, p) != last;
}
```

So `any_of` is somehow higher level call compared to `find_if`.

ClementKunz: Yes indeed we do not have to check the value of the iterator with `any_of`. Anyway, we need to wait for the bump of gcc version

luc-guyot-infomaniak (approved):

PR #467, opened 2025-01-14
KDESKTOP-1488-Create-ParentDelete-conflict-Item-not-correctly-removed-from-DB

ClementKunz:
In case of Create-ParentDelete conflict, change (Delete operation) was not propagate to DB because of a previous conflict resolution behavior that is used anymore.
Previously, we tried to keep the created file by moving it to the root of the drive. Now, we simply propagate the Delete operation.


luc-guyot-infomaniak (commented):
Any unit tests?

ClementKunz:
> Any unit tests?

Done in ca2d38c8aba0efc0b5abeaa8ff86bbdff4c771dc

ChristopheLarchier (approved):

luc-guyot-infomaniak (approved):

PR #468, opened 2025-01-15
KDESKTOP-1475-Windows-Folder-status-remain-syncing-after-a-remote-move

ClementKunz:


src/libsyncengine/propagation/executor/executorworker.cpp
```diff
@@ -1225,6 +1226,12 @@ ExitInfo ExecutorWorker::generateMoveJob(SyncOpPtr syncOp, bool &ignored, bool &
         relativeOriginLocalFilePath = correspondingNode->getPath();
         absoluteDestLocalFilePath = _syncPal->localPath() / relativeDestLocalFilePath;
         SyncPath absoluteOriginLocalFilePath = _syncPal->localPath() / relativeOriginLocalFilePath;
+
+        bool isPlaceholder = false;
```
luc-guyot-infomaniak, line 1230: At some point, passing a `VfsStatus` object (with a helpful default initialization) to `vfsStatus` should be preferred.

luc-guyot-infomaniak (commented):
Is it manageable to add some unit tests?

src/libsyncengine/propagation/executor/executorworker.cpp
ClementKunz: Done in this [PR](https://github.com/Infomaniak/desktop-kDrive/pull/488).
However, I am really not sure if it is a good thing. This gave rise to a lot of changes that could lead to regression without IMHO improving code readability that much. I would argue that in such cases, where code should be deeply refactored, we re-write everything instead of trying to do small, dangerous changes which does not improve code quality that much.

ClementKunz:
> Is it manageable to add some unit tests?

IMHO this should probably be in the scope of integration tests

ChristopheLarchier (approved):

PR #469, opened 2025-01-15
Configure sentry handler for Linux amd64

ChristopheLarchier:
The crash_handler build does not work on ARM64, you will have to wait until the end of support for the Ubuntu 20.04.


ClementKunz (dismissed):

herve-er (changes requested):
This branch needs to be recreated based on release-3.6.8. Currently, it includes all the changes from develop that we don't want to merge into the release-3.6.8 branch.

PR #473, opened 2025-01-16
Fix snapshot invalidation issue (sync is not restarting).

herve-er:
## PR Description
When synchronization encounters a `DataError`, it typically invalidates the snapshot. However, this operation consumes significant time and resources.  

#### The **3.6.8 version** of the app introduces the following speed optimization:  

In most cases, a `DataError` reflects a temporary issue caused by changes in the local file system that were not yet captured in the local snapshot when synchronization began (due to the frozen state of the snapshot copy used during synchronization). Rebuilding the snapshot from scratch in these cases is unnecessary, as restarting synchronization will use an updated snapshot that incorporates the latest changes.  

To optimize this process, the behavior of `invalidateSnapshot()` has been updated as follows:  
- Calling `invalidateSnapshot()` now increments a counter.  
- The snapshot will only be rebuilt (invlidate) when this counter reaches **3**.  
- The counter is reset each time synchronization completes successfully.  

---

### Bug Fix

A bug was identified in the implementation of this new behavior.  

To determine whether synchronization should restart, the following condition is use:  

```cpp
_syncPal->snapshot(ReplicaSide::Local)->updated() || 
_syncPal->snapshot(ReplicaSide::Remote)->updated() || 
_syncPal->restart()
```

Previously, this condition always returned `true` in cases of snapshot invalidation because rebuilding the snapshot inevitably set the `updated` flag to `true`. However, with the new behavior, where snapshots are only rebuilt once out of three invalidations, the `updated` flag may remain `false` if no changes occurred between the previous synchronization and the invalidation.  

To ensure that synchronization restarts even when the snapshot is not rebuilt, the following change was made:  

- Added `_syncPal->setRestart(true);` in the `invalidateSnapshot()` method when the counter has not yet reached 3.  

With this modification, the above condition will always return `true` after a snapshot invalidation, ensuring synchronization restarts as expected.


ChristopheLarchier (approved):

ClementKunz (dismissed):

src/libsyncengine/update_detection/file_system_observer/filesystemobserverworker.cpp
```diff
@@ -35,6 +35,7 @@ void FileSystemObserverWorker::invalidateSnapshot() {
     if (_invalidateCounter < maxRetryBeforeInvalidation) {
         LOG_SYNCPAL_DEBUG(_logger, _snapshot->side()
                                            << " snapshot is not invalidated. Invalidation count: " << _invalidateCounter);
+        _syncPal->setRestart(true);
```
luc-guyot-infomaniak, line 38: What is the rationale behind this line?

I understand that the synchronisation might have been stopped and should be restarted.
Is it possible to add more context?

herve-er: I updated the PR description: https://github.com/Infomaniak/desktop-kDrive/pull/473#issue-2791988385

luc-guyot-infomaniak: Thanks. 
**Suggestion.** Put ` _syncPal->setRestart(true);` right after line 32, with a comment explaining the consequences of this command.

herve-er: Done here: [182e2bb](https://github.com/Infomaniak/desktop-kDrive/pull/473/commits/182e2bbda90f31840ef640cded9cd3fc65b4d486)

src/libsyncengine/update_detection/file_system_observer/filesystemobserverworker.cpp
```diff
@@ -30,6 +30,8 @@ FileSystemObserverWorker::FileSystemObserverWorker(std::shared_ptr<SyncPal> sync
 
 void FileSystemObserverWorker::invalidateSnapshot() {
     if (!_snapshot->isValid()) return;
+    // The synchronisation will restart, even if there is no change in the file system and if the snapshot is not invalidated.
```
herve-er, line 33:
```suggestion
    // The synchronisation will restart, even if there is no change in the file system and if the snapshot is not actually invalidated.
```

PR #479, opened 2025-01-17
Ignore path starting with one letter and a colon

ClementKunz:
Ignore operation on files whose name start with pattern "X:".


src/libsyncengine/update_detection/file_system_observer/computefsoperationworker.cpp
```diff
@@ -968,9 +968,16 @@ bool ComputeFSOperationWorker::checkIfPathIsInDeletedFolder(const SyncPath &path
 }
 
 void ComputeFSOperationWorker::notifyIgnoredItem(const NodeId &nodeId, const SyncPath &path, const NodeType nodeType) {
-    LOGW_SYNCPAL_INFO(_logger, L"Item (or one of its descendants) has been ignored: " << Utility::formatSyncPath(path));
-    const Error err(_syncPal->syncDbId(), "", nodeId, nodeType, path, ConflictType::None, InconsistencyType::ReservedName);
-    _syncPal->addError(err);
+    if (!path.root_name().empty()) {
+        // Display the error only once per broken path
+        const auto [_, inserted] = _ignoredDirectoryNames.insert(path.root_name());
```
luc-guyot-infomaniak, line 973: This unordered set does not seem to be used anywhere else. Are some changes missing?

src/libsyncengine/update_detection/file_system_observer/snapshot/snapshot.cpp
```diff
@@ -273,14 +272,12 @@ bool Snapshot::path(const NodeId &itemId, SyncPath &path, bool &ignore) const no
...
 
-        // Trick to ignore items with pattern like "X:" in their name on Windows - Begin
-        if (path.parent_path() != tmpParentPath) {
-            ignore = true;
-            return false;
-        }
-        tmpParentPath = path;
-        // Trick to ignore items with pattern like "X:" in their name on Windows - End
+    // Trick to ignore items with pattern like "X:" in their name on Windows
+    if (!path.root_name().empty()) {
```
luc-guyot-infomaniak, line 278: What would happen with a `path` argument value of the form `X:/kDrive`? Unit tests to document the problematic cases?

src/libsyncengine/update_detection/file_system_observer/computefsoperationworker.cpp
ClementKunz: No, just here. We try to insert the root name of the path. If succeeded, meaning it is not yet into the set, we display an error.
This allows to display the error only once, even if there are many files inside the folder that cannot be synced.

ChristopheLarchier, line 973: Is it really OK to display a unique error by path.root_name()?

ClementKunz: Why not? Because if the folders contain thousands of items, there would be thousands of errors generated for the same reason. This allows to avoid the error table to be flooded with too many errors. However, I am opened if you see any better way of doing it.

ChristopheLarchier: The root_name is something like "C:", so the error won't be very significant.

ClementKunz: I have added more comments to explicit how the trick works.
Since we are using relative path here, the root name should always be empty. If it is not, that means there is an invalid name in the path and every item with the same non-empty root name should be ignored.

The flaw with this "trick" is that only 1 error will be displayed is there are 2 folders with broken names starting with the same letter (ex: "E:S1" and "E:S2")

ChristopheLarchier (approved):

extensions/MacOSX/kDriveLiteSync/Extension/xpcService.m
```diff
@@ -106,6 +106,7 @@ - (void)initOpenBlackListSet
                             @"com.apple.QuickLookThumbnailing.extension.ThumbnailExtension-macOS",  // QuickLookThumbnailing
                             @"com.apple.quicklook.thumbnail.ImageExtension",    // Quicklook Thumbnail
                             @"com.apple.quicklook.thumbnail.AudiovisualExtension",    // Quicklook Thumbnail
+                            @"com.apple.quicklook.thumbnail",                   // Quicklook Thumbnail
```
luc-guyot-infomaniak, line 109: It does not seem related to the fix of this PR:

src/libsyncengine/syncpal/isyncworker.cpp
```diff
@@ -30,7 +30,7 @@ ISyncWorker::~ISyncWorker() {
     if (_isRunning) {
         ISyncWorker::stop();
     }
-
+    waitForExit();
```
luc-guyot-infomaniak, line 33: Same remark as above.

src/libsyncengine/update_detection/file_system_observer/computefsoperationworker.cpp
luc-guyot-infomaniak: Thanks. Is it correct that `path` is always a relative path in this context? If so, what about renaming `path`with `relativePath`?

src/libsyncengine/update_detection/file_system_observer/snapshot/snapshot.cpp
```diff
@@ -273,14 +272,14 @@ bool Snapshot::path(const NodeId &itemId, SyncPath &path, bool &ignore) const no
...
+    }
 
-        // Trick to ignore items with pattern like "X:" in their name on Windows - Begin
-        if (path.parent_path() != tmpParentPath) {
-            ignore = true;
-            return false;
-        }
-        tmpParentPath = path;
-        // Trick to ignore items with pattern like "X:" in their name on Windows - End
+    // Trick to ignore items with pattern like "X:" in their name on Windows.
```
luc-guyot-infomaniak, line 277:
```suggestion
    // Trick to ignore items with a pattern like "X:" in their names on Windows.
```

src/libsyncengine/update_detection/file_system_observer/snapshot/snapshot.cpp
```diff
@@ -273,14 +272,14 @@ bool Snapshot::path(const NodeId &itemId, SyncPath &path, bool &ignore) const no
...
+    // Since only relative path are stored in the snapshot, the root name shoudl always be empty.
```
luc-guyot-infomaniak, line 278:
```suggestion
    // Since only relative path are stored in the snapshot, the root name should always be empty.
```

test/libsyncengine/update_detection/file_system_observer/testsnapshot.cpp
```diff
@@ -241,6 +241,42 @@ void TestSnapshot::testPath() {
...
+    for (const auto &name: names) {
+        Snapshot snapshotCopy(ReplicaSide::Local, _dummyRootNode);
+        snapshotCopy = *_snapshot;
+        const auto id = CommonUtility::generateRandomStringAlphaNum();
+        const SnapshotItem item(id, "a", name, testhelpers::defaultTime, testhelpers::defaultTime, NodeType::Directory,
+                                testhelpers::defaultFileSize, false, true, true);
+        snapshotCopy.updateItem(item);
+        SyncPath path;
+        bool ignore = false;
+        // On Windows, if the file name starts with "X:" pattern, the previous element of the path are overrode
```
luc-guyot-infomaniak, line 257:
```suggestion
        // On Windows, if the file name starts with the "X:" pattern, the previous elements of the path are overrode
```

test/libsyncengine/update_detection/file_system_observer/testsnapshot.cpp
```diff
@@ -241,6 +241,42 @@ void TestSnapshot::testPath() {
...
         CPPUNIT_ASSERT(!ignore);
+#endif
+    }
+    // Same test but on the snapshot copy (meaning using paths stored in cache)
+    std::vector<SyncName> names = {Str("E:S"), Str("E:/S")};
+    for (const auto &name: names) {
+        Snapshot snapshotCopy(ReplicaSide::Local, _dummyRootNode);
+        snapshotCopy = *_snapshot;
+        const auto id = CommonUtility::generateRandomStringAlphaNum();
+        const SnapshotItem item(id, "a", name, testhelpers::defaultTime, testhelpers::defaultTime, NodeType::Directory,
```
luc-guyot-infomaniak, line 252: The last three arguments of the `SnapshotItem` constructor can be omitted.

test/libsyncengine/update_detection/file_system_observer/testsnapshot.cpp
```diff
@@ -241,6 +241,42 @@ void TestSnapshot::testPath() {
...
+#else
+        CPPUNIT_ASSERT(snapshotCopy.path(id, path, ignore));
+        CPPUNIT_ASSERT_EQUAL(SyncPath("A") / name, path);
+        CPPUNIT_ASSERT(!ignore);
+#endif
+
+        const auto childId = CommonUtility::generateRandomStringAlphaNum();
+        const auto childName = Str("test");
+        const SnapshotItem childItem(childId, id, childName, testhelpers::defaultTime, testhelpers::defaultTime,
+                                     NodeType::Directory, testhelpers::defaultFileSize, false, true, true);
```
luc-guyot-infomaniak, line 271: The last three arguments of the `SnapshotItem` constructor can be omitted.

ClementKunz: No we cannot, but the source file was misleading indeed. Fixed in 99610d00f4d7aaa4c3de8b1fcfc58fea68d1b350 (in the other PR)

test/libsyncengine/update_detection/file_system_observer/testsnapshot.cpp
ClementKunz: No we cannot, but the source file was misleading indeed. Fixed in 99610d0 (in the other PR)

src/libsyncengine/update_detection/file_system_observer/computefsoperationworker.cpp
ClementKunz: Done in 8628fef099799009388f0273129d06420a0693f1

extensions/MacOSX/kDriveLiteSync/Extension/xpcService.m
ClementKunz: Indeed, that was because the branch was created from branch `release-3.6.8`. I have created another [PR](https://github.com/Infomaniak/desktop-kDrive/pull/491) and I will close this one.

src/libsyncengine/syncpal/isyncworker.cpp
ClementKunz: See comment above

ClementKunz:
To be closed, replaced by https://github.com/Infomaniak/desktop-kDrive/pull/491

PR #480, opened 2025-01-20
Kdesktop 1451 windows fix always available locally folder status

herve-er:
### PR Description

**Main Goal**: To stop forcing the pin state on Windows. (A PR addressing the issue on macOS is on the way).

**Changes**:
- **Executor Worker**:
  - When a placeholder is created, we check its pin state (inherited from the parent) on Windows.
  - If the pin state is pinned, we request the hydration of the file.

- **SyncPal**:
  - At syncpal startup, if a file is pinned but not hydrated, it will be hydrated.
  - If a file is unpinned but hydrated, it will be dehydrated.
  - If the pin state is unspecified, no action will be taken.


src/libcommonserver/vfs/win/vfs_win.cpp
```diff
@@ -497,6 +497,9 @@ ExitInfo VfsWin::isDehydratedPlaceholder(const SyncPath &initFilePathStd, bool &
 }
 
 ExitInfo VfsWin::setPinState(const SyncPath &relativePathStd, PinState state) {
+    // We currently don't have any features that require pinning/unpinning files manually on windows.
+    return ExitCode::Ok;
```
ClementKunz, line 501: Really!? What about "Always make available locally"? This does not set the pin state to "Pinned"?

src/libcommonserver/vfs/win/vfs_win.cpp
```diff
@@ -536,19 +544,27 @@ PinState VfsWin::pinState(const SyncPath &relativePathStd) {
     // VFS_PIN_STATE_INHERIT.
```
ClementKunz, line 544: So you did this TODO, we can remove this comment

src/libsyncengine/propagation/executor/executorworker.cpp
```diff
@@ -189,7 +190,7 @@ void ExecutorWorker::execute() {
                     } else if (syncOp->affectedNode() && syncOp->affectedNode()->inconsistencyType() != InconsistencyType::None) {
                         setProgressComplete(syncOp, SyncFileStatus::Inconsistency);
                     } else {
-                        setProgressComplete(syncOp, SyncFileStatus::Success);
+                        setProgressComplete(syncOp, hydrating ? SyncFileStatus::Syncing : SyncFileStatus::Success);
```
ClementKunz, line 193: Is it necessary to manage the status here. It should be managed by the download job that should be created to hydrate the file.

src/libsyncengine/syncpal/syncpalworker.cpp
```diff
@@ -578,49 +578,45 @@ bool SyncPalWorker::resetVfsFilesStatus() {
...
-                    if (ExitInfo exitInfo = _syncPal->vfsSetPinState(dirIt->path(), PinState::Unspecified); !exitInfo) {
-                        LOGW_SYNCPAL_WARN(_logger,
-                                          L"Error in vfsSetPinState : " << Utility::formatSyncPath(dirIt->path()) << exitInfo);
-                        ok = false;
-                        continue;
-                    }
+            bool hydrationOrDehydrationInProgress = false;
+            const SyncPath relativePath =
+                    CommonUtility::relativePath(_syncPal->localPath(), dirIt->path()); // Get the relative path of the file
+            _syncPal->fileSyncing(ReplicaSide::Local, relativePath, hydrationOrDehydrationInProgress);
```
ClementKunz, line 608: Are we sure that the "syncing" field in DB is meant for that only? Is it reliable? It looks like a shortcut to me

src/libcommonserver/vfs/win/vfs_win.cpp
herve-er: The OS handles this. When the user clicks on "Always make available locally," the pinState changes to pinned (and the app has nothing to do; this will even occur if the app is closed).  

A use case where we might want to manually set a pinState would be, for example, if we had a "favorite folder" similar to the common folder and decided that, by default, the favorite folder should always be available locally. In this case, we would call `setPinState` on the "favorite folder."

src/libsyncengine/syncpal/syncpalworker.cpp
herve-er: Dans le code de la `Vfs`, il est présenté comme suit :  
```cpp  
// Set hydrating indicator (avoid double hydration)  
_setSyncFileSyncing(_vfsSetupParams._syncDbId, fileRelativePath, true);  
```  

Et en vérifiant le code, c'est bien le seul contexte dans lequel il est utilisé :  
- Mis à `true` quand l'hydratation est demandée.  
- Mis à `false` quand l'hydratation ou la déshydratation est terminée ou lorsque l'on reçoit un `cancel hydrate`.

src/libcommonserver/vfs/win/vfs_win.cpp
ClementKunz: Oh ok, I did not understand it correctly. So the pin state can be set to `Pinned`for example, just not using this method

src/libsyncengine/syncpal/syncpalworker.cpp
ClementKunz: Okay, I am not a huge fan of the wording but this is a another problem 😅 we will not migrate the DB just for that

ClementKunz (approved):

PR #481, opened 2025-01-20
Windows - LitesSync - Crash when registering folder

ChristopheLarchier:
Not reproduced.
Checks added to avoid crash.


extensions/windows/cfapi/Vfs/cloudproviderregistrar.cpp
```diff
@@ -48,6 +48,10 @@ std::wstring CloudProviderRegistrar::registerWithShell(ProviderInfo *providerInf
 
     try {
         syncRootID = getSyncRootId(providerInfo);
+        if (syncRootID.empty()) {
```
ClementKunz, line 51: Perhaps we could send a Sentry message when this method fail, otherwise we will probably never understand the root cause of this problem. Since we are in the extension here, the sentry message could be sent from `VfsWin::startImpl` for example.

PR #488, opened 2025-01-23
Refactoring vfs status2

ClementKunz:
Create a struct `VfsStatus` to reduce the number of arguments of vfs methods


src/libsyncengine/jobs/network/API_v2/createdirjob.cpp
```diff
@@ -79,7 +82,8 @@ bool CreateDirJob::handleResponse(std::istream &is) {
             }
         }
 
-        if (!_filePath.empty() && _vfsForceStatus && !_vfsForceStatus(_filePath, false, 100, true)) {
+        VfsStatus vfsStatus(true, true, false, 100);
```
luc-guyot-infomaniak, line 85: You may consider designated initialization (https://www.cppstories.com/2021/designated-init-cpp20/), or alternatively a dedicated method that sets the data members if this method bears a meaningful name and can be re-used in other situations.

PR #503, opened 2025-01-28
The first operation after the app starts is ignored.

herve-er:
The first generated operation has a UniqueId of 0.

The following code in the executorWorker leads to this operation being ignored:
```cpp
UniqueId opId = 0;
_opListMutex.lock();
if (!_opList.empty()) {
    opId = _opList.front();
    _opList.pop_front();
}
_opListMutex.unlock();

if (!opId) break; // <-- /!\ Evaluate to false for the first SyncOperation
```

However, it will be propagated as soon as another change occurs or if the user pauses and restarts the synchronization.


src/libsyncengine/reconciliation/syncoperation.cpp
```diff
@@ -20,7 +20,7 @@
 
 namespace KDC {
 
-UniqueId SyncOperation::_nextId = 0;
+UniqueId SyncOperation::_nextId = 1;
```
luc-guyot-infomaniak, line 23: Why does it solve the issue reported in the description of the PR?

herve-er: The first operation will have the ID 1 instead of 0; therefore, `if (!opId) break;` will not be triggered.  

Another solution would be to change the executor code from:

```cpp
UniqueId opId = 0;
_opListMutex.lock();
if (!_opList.empty()) {
    opId = _opList.front();
    _opList.pop_front();
}
_opListMutex.unlock();

if (!opId) break;
```

to:

```cpp
UniqueId opId = -1;
_opListMutex.lock();
if (!_opList.empty()) {
    opId = _opList.front();
    _opList.pop_front();
}
_opListMutex.unlock();

if (opId < 0) break;
```

However, in my opinion, the implemented solution (starting with ID 1) is safer and more likely to prevent similar issues elsewhere in the codebase in the future.

ChristopheLarchier (approved):

ChristopheLarchier (approved):

PR #504, opened 2025-01-29
KDESKTOP-1515-Sync-loop

ClementKunz:
**Initial state:**

Create 2 files “test.txt” and “test2.txt”

**Steps to reproduce:**

Pause the sync

On remote:
- Edit “test.txt”
- Rename “test2.txt” into “test2_old.txt”
- Rename “test.txt” into “test2.txt”

Start the sync

**Result:**

Sync is looping


ChristopheLarchier (approved):

luc-guyot-infomaniak (commented):
A question regarding the comparison of source and target names of move operations.

src/libsyncengine/propagation/operation_sorter/operationsorterworker.cpp
```diff
@@ -419,11 +419,11 @@ void OperationSorterWorker::fixMoveBeforeMoveOccupied() {
...
 
-            std::shared_ptr<Node> otherMoveNode = moveOp->affectedNode();
-            if (otherMoveNode->parentNode() != nullptr) {
-                std::optional<NodeId> moveDestParentId = otherMoveNode->parentNode()->id();
-                if (sourceParentId == moveDestParentId) {
-                    if (moveNode->name() == otherMoveNode->name()) {
+            if (std::shared_ptr<Node> otherMoveNode = moveOp->affectedNode(); otherMoveNode->parentNode() != nullptr) {
+                if (std::optional<NodeId> moveDestParentId = otherMoveNode->parentNode()->id();
+                    sourceParentId == moveDestParentId) {
+                    if (moveNode->name() == otherMoveNode->moveOrigin()->filename() ||
```
luc-guyot-infomaniak, line 425: Why is the condition symmetric with respect to `moveNode` and `otherMoveNode`? 
If I understand correctly the above lines, we only know that the source directory of `moveOp2` (affects `moveNode`) is also the destination of `movOp` (affects `otherMoveNode`). So I would only check `moveNode->moveOrigin()->filename() == otherMoveNode->name()`. Am I missing something?

ClementKunz: I think you are absolutely correct. Even more, your remark pointed out an error in the algorithm.
I think there is a mistake in the thesis (Algorithmus 10, p.206) and I have modified the proposed algorithm as follows:
```
def fix_move_before_move_occupied(ops, sorted_ops, db, snapshot) :

foreach op in ops where optype(op) == move :
    foreach otherOp in ops where optype(otherOp) == move :
        if op == otherOp :
            continue

        (source, dest) = path(op)
        source_parent_id = id(db, parent(source))

        (other_source, other_dest) = path(otherOp)
        move_dest_parent_id = id(snapshot, parent(other_dest))

        if source_parent_id == move_dest_parent_id :
            if basename(source) == basename(other_dest) :
                if sorted_ops.index(op) > sorted_ops.index(otherOp) :
                    move_first_after_second(otherOp, op, sorted_ops)
```

Finally, I did not activate the unit tests on this particular function (`fixMoveBeforeMoveOccupied`) before. I have now re-activated and refactored them to take into account the specific case described in this issue.

luc-guyot-infomaniak: **Suggestion.** Rename `OperationSorterWorker::moveFirstAfterSecond` with `OperationSorterWorker::swap`.

ClementKunz: The error on the Windows CI pointed out another logic error.
To answer the original comment : Yes, we need to check both conditions (`nodeCurrentName ` vs `otherNodeOriginName` AND `nodeOriginName ` vs `otherNodeCurrentName` ) but the direction of the inversion is different.
I hope this now correct and the implementation clear enough

ClementKunz: > Suggestion. Rename OperationSorterWorker::moveFirstAfterSecond with OperationSorterWorker::swap

I do agree that the name is not very good but `swap` would be incorrect since the operations are not swapped in the list, the first one is removed and added back just after the second one

src/libsyncengine/propagation/operation_sorter/operationsorterworker.cpp
```diff
@@ -389,48 +389,55 @@ void OperationSorterWorker::fixDeleteBeforeCreate() {
...
+
+            NodeId nodeOriginParentId;
+            if (!getIdFromDb(node->side(), nodeOriginPath->parent_path(), nodeOriginParentId)) continue;
+            NodeId otherNodeOriginParentId;
+            if (!getIdFromDb(otherNode->side(), otherNodeOriginPath->parent_path(), otherNodeOriginParentId)) continue;
+
+            if (nodeParentId == otherNodeOriginParentId && nodePath.filename() == otherNodeOriginPath->filename()) {
+                // move only if op is before otherMoveOp
+                moveFirstAfterSecond(op, otherOp);
+                continue;
```
luc-guyot-infomaniak, line 433: Don't we need an `else if` statement to make sure to go through the `erase` instruction below?
Don't we need a second `erase` statement for the first list? 
Is all this safe (removing items while iterating over the lists)?

src/libsyncengine/propagation/operation_sorter/operationsorterworker.cpp
luc-guyot-infomaniak: I thought that the two nested for loops over the move operations were taking care of the symmetry between `move` and `otherMove`.  You rightly pointed out that we erase operation ids from the lists on which we iterate.

- Is it safe? 
- Is it more efficient than two bare nest for loops?  Now that we handle properly the symmetry the most nested clause, instructions are doubled too.

src/libsyncengine/propagation/operation_sorter/operationsorterworker.cpp
```diff
@@ -389,48 +389,55 @@ void OperationSorterWorker::fixDeleteBeforeCreate() {
...
-            std::shared_ptr<Node> otherMoveNode = moveOp->affectedNode();
-            if (otherMoveNode->parentNode() != nullptr) {
-                std::optional<NodeId> moveDestParentId = otherMoveNode->parentNode()->id();
-                if (sourceParentId == moveDestParentId) {
-                    if (moveNode->name() == otherMoveNode->name()) {
-                        // move only if moveOp is before op
-                        moveFirstAfterSecond(moveOp2, moveOp);
-                    }
-                }
+            const auto nodePath = node->getPath();
```
luc-guyot-infomaniak, line 417: I would propose to use `const auto&` (because we might be copying strings) unless we are certain that our compilers have activated the Return Value Optimization (which is likely).

src/libsyncengine/propagation/operation_sorter/operationsorterworker.cpp
ClementKunz: > Don't we need an else if statement to make sure to go through the erase instruction below?
I don't think so, since the `erase` statement is outside the second loop

> Don't we need a second erase statement for the first list?
No, since we remove from list 2 the current item from list 1.

Let's take the operation set : A, B, C.
On first passage in the outer loop, we will check: A vs B, A vs C.
Then A is removed from the set used in the inner loop.
On the second passage in the outer loop, we will check: B vs C.

> Is all this safe (removing items while iterating over the lists)?
We are not doing that. We remove the item from the inner loop set once we are outside the inner loop. The inner and outer loop sets are independent (but initialize with the same values).

src/libsyncengine/propagation/operation_sorter/operationsorterworker.cpp
ClementKunz: > Is it safe?

I think it is, but cannot actually prove it mathematically.

> Is it more efficient than two bare nest for loops? Now that we handle properly the symmetry the most nested clause, instructions are doubled too.

On second thought, it is probably more or less the same in terms of efficiency, since we could remove half of the path computation from each loop. It is probably more readable with two bare nest for loops, as you suggest. It is also how it was indicated in the thesis.

ClementKunz: Done in 8c1a138e5f1ab0a341af268b1e2b194d4e2fa6de

src/libsyncengine/propagation/operation_sorter/operationsorterworker.cpp
ClementKunz: However, all this is not relevant anymore after 8c1a138e5f1ab0a341af268b1e2b194d4e2fa6de

luc-guyot-infomaniak (approved):

ChristopheLarchier (approved):

PR #506, opened 2025-01-30
KDesktop-1535: Set up a custom rate limit on specific Sentry transactions

herve-er:


ClementKunz:
Looks good to me 👍 
However, my only interrogation is about the performance of generating random number on every call of `AbstractPTrace::start()`. Could we imagine a simpler and faster way of rate limiting those calls like, for example, a static counter?
I am not even sure that what I am proposing would be more optimized. It would probably require the computation of a modulus operation or something like that. Just perhaps something to think about

herve-er:
> Looks good to me 👍 However, my only concern is the performance impact of generating a random number on every call to `AbstractPTrace::start()`. Could we consider a simpler and faster way to rate-limit those calls, such as using a static counter? I'm not even sure that this approach would be more optimized—it would likely involve computing a modulus operation or something similar. Just something to think about.  

That was actually my first approach, but after giving it more thought, I realized it could introduce a significant bias in the forwarded metrics. Using a counter would, for example, ensure that the first trace is always sent. In some scenarios, the first traces might consistently be faster (or slower), leading to less accurate metrics than randomly chosen traces.  

I just ran a performance test, and the difference with or without sampling is minimal, less than 7ms for 100,000 traces.  
Since the sampling rate is per trace, implementing it with a static counter would also require a `std::map<PTraceName, int /*counter*/>` and a modulo operation, which would likely be at least as slow as using a random number generator.

src/libcommon/log/sentry/ptraces.h
```diff
@@ -97,12 +97,13 @@ struct Propagation2 : public AbstractPTrace {
...
  destroyed.
  Some scoped performance traces expect to be manually stopped. In this case, the stop() method must be called, else the trace will
- be consider as aborted when the object will be destroyed. Such traces are indicated by a comment in their class definition.
+ be considered as aborted when the object will be destroyed. Such traces are indicated by a comment in their class definition.
  */
 namespace scoped {
 struct LFSOChangeDetected : public AbstractScopedPTrace {
         explicit LFSOChangeDetected(int syncDbId) :
-            AbstractScopedPTrace({"LFSO_ChangeDetected", "Handle one detected changes", PTraceName::LFSOChangeDetected},
+            AbstractScopedPTrace({"LFSO_ChangeDetected", "Handle one detected changes", PTraceName::LFSOChangeDetected,
```
luc-guyot-infomaniak, line 105:
```suggestion
            AbstractScopedPTrace({"LFSO_ChangeDetected", "Handle one detected change", PTraceName::LFSOChangeDetected,
```

src/libcommon/log/sentry/ptraces.h
```diff
@@ -115,8 +116,9 @@ struct LFSOGenerateInitialSnapshot : public AbstractScopedPTrace {
 
 struct RFSOChangeDetected : public AbstractScopedPTrace {
         explicit RFSOChangeDetected(int syncDbId) :
-            AbstractScopedPTrace({"RFSO_ChangeDetected", "Handle one detected changes", PTraceName::RFSOChangeDetected},
-                                 PTraceStatus::Ok, syncDbId) {}
+            AbstractScopedPTrace(
+                    {"RFSO_ChangeDetected", "Handle one detected changes", PTraceName::RFSOChangeDetected, PTraceName::None, 0.1},
```
luc-guyot-infomaniak, line 120:
```suggestion
                    {"RFSO_ChangeDetected", "Handle one detected change", PTraceName::RFSOChangeDetected, PTraceName::None, 0.1},
```

ClementKunz (approved):

src/libcommon/log/sentry/ptracedescriptor.h
```diff
@@ -67,14 +67,16 @@ enum class PTraceName {
...
         PTraceDescriptor() = default;
-        PTraceDescriptor(std::string pTraceTitle, std::string pTraceDescription, PTraceName pTraceName,
-                         PTraceName parentPTraceName = PTraceName::None) :
+        PTraceDescriptor(std::string pTraceTitle, std::string pTraceDescription, const PTraceName pTraceName,
+                         const PTraceName parentPTraceName = PTraceName::None, const double _customSampleRate = 1.0) :
             _pTraceName{pTraceName}, _parentPTraceName{parentPTraceName}, _pTraceTitle{std::move(pTraceTitle)},
-            _pTraceDescription{std::move(pTraceDescription)} {}
+            _pTraceDescription{std::move(pTraceDescription)}, _customSampleRate{_customSampleRate} {}
 
         const PTraceName _pTraceName = PTraceName::None;
```
luc-guyot-infomaniak, line 75: Why using the underscore prefix for public data members?

src/libcommon/log/sentry/abstractptrace.h
```diff
@@ -56,13 +62,22 @@ class AbstractPTrace {
...
 
     private:
         AbstractPTrace &operator=(AbstractPTrace &&) = delete;
         pTraceId _pTraceId{0};
         PTraceDescriptor _pTraceInfo;
         int _syncDbId = -1;
+
+        bool _rateLimited = false;
+        bool checkCustomSampleRate() const {
+            if (_pTraceInfo._customSampleRate >= 1.0) return true;
```
luc-guyot-infomaniak, line 76: Use a public data member or a public accessor.

src/libcommon/log/sentry/abstractptrace.h
```diff
@@ -29,22 +30,27 @@ class AbstractPTrace {
...
+            }
+            _rateLimited = false;
             _pTraceId = sentry::Handler::instance()->startPTrace(_pTraceInfo, _syncDbId);
             return *this;
         }
 
         // Stop the performance trace.
-        void _stop(PTraceStatus status = PTraceStatus::Ok) noexcept {
+        void _stop(const PTraceStatus status = PTraceStatus::Ok) noexcept {
+            if (_rateLimited) return;
```
luc-guyot-infomaniak, line 53: So, we call `_stop`, but it won't stop?

**Suggestion.** Either check the value of `_rateLimited` before calling `_stop` or rename `_stop` with `_stopIfRequired`.

src/libcommon/log/sentry/abstractptrace.h
```diff
@@ -29,22 +30,27 @@ class AbstractPTrace {
...
         virtual void restart() { _restart(); }
-
     protected:
         explicit AbstractPTrace(const PTraceDescriptor &info) : _pTraceInfo(info) {};
-        explicit AbstractPTrace(const PTraceDescriptor &info, int dbId) : _pTraceInfo(info), _syncDbId(dbId) {};
+        explicit AbstractPTrace(const PTraceDescriptor &info, const int dbId) : _pTraceInfo(info), _syncDbId(dbId) {};
 
         // Start a new performance trace.
         inline AbstractPTrace &_start() {
+            if (!checkCustomSampleRate()) {
```
luc-guyot-infomaniak, line 42: Wouldn't it be clearer to call a method named something like `isRateLimited`?

Actually, I don't understand the logic very well. Could we discuss it?

PR #507, opened 2025-01-30
KDESKTOP-1530-Refactor-VfsStatus

ClementKunz:
Create a `VfsStatus` structure to avoid passing too many arguments in functions

### DIFF
--- a/src/libcommonserver/vfs/mac/litesyncextconnector.h
+++ b/src/libcommonserver/vfs/mac/litesyncextconnector.h
@@ -20,13 +20,13 @@
 
 #include "libcommon/utility/types.h"
 #include "libcommonserver/log/log.h"
+#include "vfs/vfs.h"
 
 #include <sys/stat.h>
 
 #include <QMap>
 #include <QHash>
 #include <QSet>
-#include <QPixmap>
 #include <QString>
 
 namespace KDC {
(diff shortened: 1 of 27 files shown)


src/libcommonserver/io/iohelper.h
```diff
@@ -377,7 +377,7 @@ struct IoHelper {
          \param isHydrated true if the file is hydrated.
          \return the 'status' extended attribute value.
          */
-        inline static std::string statusXAttr(bool isSyncing, int progress, bool isHydrated) {
+        static std::string statusXAttr(const bool isSyncing, const int64_t progress, const bool isHydrated) {
```
ChristopheLarchier, line 380: Why int64_t instead of int? (progress is between 0 and 100)

ClementKunz: I change that because of the new SonarCloud rules. However, int16_t is indeed enough since the value is between 0 and 100. I will make the change.

test/libsyncengine/jobs/network/testnetworkjobs.cpp
```diff
@@ -512,18 +512,15 @@ void TestNetworkJobs::testDownloadAborted() {
...
-        isSyncingRes = isSyncing;
-        progressRes = progress;
-        isHydratedRes = isHydrated;
-        return ExitCode::Ok;
-    });
+    VfsStatus vfsStatusRes;
+    vfs->setForceStatusMock(
+            [&forceStatusCalled, &vfsStatusRes]([[maybe_unused]] const SyncPath &path, const VfsStatus &vfsStatus) -> ExitInfo {
+                forceStatusCalled = true;
+                vfsStatusRes.isSyncing = vfsStatus.isSyncing;
```
luc-guyot-infomaniak, line 519: Wouldn't copy assignment be shorter to write? (the `placeholder` data member is however ignored in the current implementation).

ClementKunz: Done in 4741ac114cfd3573953325c69e20e5b779a1e9f4

luc-guyot-infomaniak (approved):

PR #508, opened 2025-01-30
Do not retry to upload a file if it exceeds the drive quota

luc-guyot-infomaniak:


luc-guyot-infomaniak:
Tests indicate that an upload request that fails because the drive quota is exceeded is not re-sent by means of the automated `re-try` mechanism. 

Thus the proposed changes are not relevent.

PR #522, opened 2025-02-06
KDESKTOP-1552 - Fix download job

ClementKunz:
Hydrated placeholder that was edited from remote replica were renamed with "conflict" suffix locally

### DIFF
--- a/src/libsyncengine/jobs/network/API_v2/downloadjob.cpp
+++ b/src/libsyncengine/jobs/network/API_v2/downloadjob.cpp
@@ -161,11 +161,13 @@ void DownloadJob::runJob() noexcept {
         bool dummyIsPlaceHolder = false;
         bool dummyIsSyncing = false;
         int dummyProgress = 0;
-        if (_vfs) {
-            _vfs->status(_localpath, dummyIsPlaceHolder, _isHydrated, dummyIsSyncing, dummyProgress);
-        } else {
-            _isHydrated = true;
-        }
+         if (const ExitInfo exitInfo = _vfs->status(_localpath, dummyIsPlaceHolder, _isHydrated, dummyIsSyncing, dummyProgress);
+            !exitInfo) {
+            LOGW_WARN(_logger, L"Error in vfsStatus: " << Utility::formatSyncPath(_localpath) << L": " << exitInfo);
+            _exitCode = exitInfo.code();
+            _exitCause = exitInfo.cause();
+            return;
+        }  
 
         if (const ExitInfo exitInfo = _vfs->forceStatus(_localpath, true, 0, _isHydrated); !exitInfo) {
             LOGW_WARN(_logger, L"Error in vfsForceStatus: " << Utility::formatSyncPath(_localpath) << L": " << exitInfo);
--- a/src/libsyncengine/jobs/network/API_v2/downloadjob.cpp
+++ b/src/libsyncengine/jobs/network/API_v2/downloadjob.cpp
@@ -158,7 +158,16 @@ void DownloadJob::runJob() noexcept {
             return;
         }
 
-        if (const ExitInfo exitInfo = _vfs->forceStatus(_localpath, true, 0, false); !exitInfo) {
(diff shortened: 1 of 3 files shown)


src/libsyncengine/jobs/network/API_v2/downloadjob.cpp
```diff
@@ -158,7 +158,16 @@ void DownloadJob::runJob() noexcept {
...
 
-        if (const ExitInfo exitInfo = _vfs->forceStatus(_localpath, true, 0, false); !exitInfo) {
+        bool dummyIsPlaceHolder = false;
+        bool dummyIsSyncing = false;
+        int dummyProgress = 0;
+        if (_vfs) {
+            _vfs->status(_localpath, dummyIsPlaceHolder, _isHydrated, dummyIsSyncing, dummyProgress);
+        } else {
+            _isHydrated = true;
+        }
```
herve-er, line 168:
```suggestion
         if (const ExitInfo exitInfo = _vfs->status(_localpath, dummyIsPlaceHolder, _isHydrated, dummyIsSyncing, dummyProgress);
            !exitInfo) {
            LOGW_WARN(_logger, L"Error in vfsStatus: " << Utility::formatSyncPath(_localpath) << L": " << exitInfo);
            _exitCode = exitInfo.code();
            _exitCause = exitInfo.cause();
            return;
        }  
```

ChristopheLarchier (approved):

src/libsyncengine/jobs/network/API_v2/downloadjob.cpp
herve-er: And the default value of `_isHydrated` should be changed to `true` (`downloadjob.h`, line 85), as this code is only executed when VFS is set and it is not a creation (`downloadjob.cpp`, line 130).

herve-er (changes requested):

src/libsyncengine/jobs/network/API_v2/downloadjob.cpp
herve-er: (It will fix the CI, by the way.)

src/libsyncengine/jobs/network/API_v2/downloadjob.cpp
```diff
@@ -158,7 +158,18 @@ void DownloadJob::runJob() noexcept {
             return;
         }
 
-        if (const ExitInfo exitInfo = _vfs->forceStatus(_localpath, true, 0, false); !exitInfo) {
+        bool dummyIsPlaceHolder = false;
+        bool dummyIsSyncing = false;
+        int dummyProgress = 0;
+         if (const ExitInfo exitInfo = _vfs->status(_localpath, dummyIsPlaceHolder, _isHydrated, dummyIsSyncing, dummyProgress);
```
luc-guyot-infomaniak, line 164: **For my undestanding.** How does it relate to the PR description? I understand that the proposed changes handle the errors encountered when calling `_vfs->status(...)` in  such a way that the `_isHydrated` flag is left unchanged. What kind of errors were raised in the problem you have reproduced?

ClementKunz: The problem was that the status was set to "Syncing" at the beginning of the download. Therefor, when the status was read later, it was always "dehydrated".
So, with this fix, the variable `_isHydrated ` is set before changing the file's status and the behavior during download is based on this value

herve-er (dismissed):

src/libsyncengine/jobs/network/API_v2/downloadjob.cpp
luc-guyot-infomaniak, line 164: Is it manageable to write a unit test that would prevent a regression?

ClementKunz: Done in f5c93fd02650888aa3cf44c1d47d14ade272c329

luc-guyot-infomaniak (approved):

PR #525, opened 2025-02-07
Kdesktop 1555 improve handling of client timeout leading to zombie server

herve-er:


src/server/appserver.cpp
```diff
@@ -1876,8 +1878,18 @@ void AppServer::onRequestReceived(int id, RequestNum num, const QByteArray &para
             break;
         }
         case RequestNum::UTILITY_DISPLAY_CLIENT_REPORT: {
-            if (static bool appStartPTraceStopped = false; !appStartPTraceStopped) {
-                appStartPTraceStopped = true;
+            using namespace std::chrono;
+            if (duration_cast<seconds>(system_clock::now() - _lastClientRestartByUserTimeStamp).count() < 10 /*seconds*/) {
+                if (!_appStartPTraceStopped) { // If the client never started, we abort the AppStart pTrace
```
ClementKunz, line 1883: How the fact `_appStartPTraceStopped` is `false` means that the client never start?
This whole portion of code is quite difficult to understand IMO, perhaps a small comment explaining what is done here would be nice

herve-er:
```suggestion
                // If the client initially started by the server never sends the UTILITY_DISPLAY_CLIENT_REPORT,  
                // we consider the client's startup aborted, and the user was forced to manually start the client again.
                if (!_appStartPTraceStopped) { 
```

src/server/appserver.cpp
```diff
@@ -2201,6 +2213,13 @@ void AppServer::onMessageReceivedFromAnotherProcess(const QString &message, QObj
         showSynthesis();
     } else if (message == showSettingsMsg) {
         showSettings();
+    } else if (message == restartClientMsg) {
+        _lastClientRestartByUserTimeStamp = std::chrono::system_clock::now();
+        if (!startClient()) {
```
ChristopheLarchier, line 2218: You don't check if kdrive_client is already running before trying to start it again?

src/server/appserver.cpp
```diff
@@ -1876,8 +1878,18 @@ void AppServer::onRequestReceived(int id, RequestNum num, const QByteArray &para
             break;
         }
         case RequestNum::UTILITY_DISPLAY_CLIENT_REPORT: {
-            if (static bool appStartPTraceStopped = false; !appStartPTraceStopped) {
-                appStartPTraceStopped = true;
+            using namespace std::chrono;
+            if (duration_cast<seconds>(system_clock::now() - _lastClientRestartByUserTimeStamp).count() < 10 /*seconds*/) {
```
ChristopheLarchier, line 1882: Why 10 seconds?

src/server/appserver.cpp
herve-er: I don't. If the client isn't running, it will start again, and if it was already running, it will display the message: 'kDrive client is already running.'

We have to show a message in response to the user's action. As the client already implements the right behavior to display a clear message to the user, this seems to be the most straightforward solution.

src/server/appserver.cpp
herve-er: It is in fact not needed anymore. 
Here is a simplified version: [0821936](https://github.com/Infomaniak/desktop-kDrive/pull/525/commits/082193671ac68a4ba88fa5f4428bb2f8f2cfa2ee)

ChristopheLarchier (approved):

PR #534, opened 2025-02-11
Kdesktop 1562 use precompiled headers

ChristopheLarchier:
macOS: 2'41'' => 2'19''
Windows: 13'01'' => 6'10''
Linux: 4'47'' => 2'36''

### DIFF
--- a/src/libcommon/log/sentry/handler.cpp
+++ b/src/libcommon/log/sentry/handler.cpp
@@ -21,6 +21,7 @@
 #include "version.h"
 #include "utility/utility.h"
 
+#include <algorithm>
 #include <iostream>
 #include <fstream>
 #include <sstream>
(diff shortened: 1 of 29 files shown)


src/libcommon/log/sentry/handler.cpp
```diff
@@ -352,7 +353,7 @@ void Handler::handleEventsRateLimit(SentryEvent &event, bool &toUpload) {
     }
 
     auto &storedEvent = it->second;
-    storedEvent.captureCount = std::min(storedEvent.captureCount + 1, UINT_MAX - 1);
+    storedEvent.captureCount = (std::min)(storedEvent.captureCount + 1, UINT_MAX - 1);
```
luc-guyot-infomaniak, line 356: What is the effect of this change? Why is it necessary?

ChristopheLarchier: It's needed to force the use of std function instead of Windows' macro defined in windef.h

luc-guyot-infomaniak (approved):

PR #535, opened 2025-02-11
Kdesktop 1170 send all logs with matching session name when the last session option is selected

herve-er:


src/libsyncengine/jobs/network/API_v2/loguploadjob.cpp
```diff
@@ -0,0 +1,624 @@
...
+}
+
+ExitInfo LogUploadJob::generateUserDescriptionFile(const SyncPath &outputPath) const {
+    const std::string osName = CommonUtility::platformName().toStdString();
+    const std::string osArch = CommonUtility::platformArch().toStdString();
+    const std::string appVersion = CommonUtility::userAgentString();
+
+    std::ofstream file((outputPath / "user_description.txt").string());
+    if (!file.is_open()) {
+        LOGW_WARN(Log::instance()->getLogger(), L"Unable to create : " << Utility::formatSyncPath(outputPath).c_str());
```
luc-guyot-infomaniak, line 402:
```suggestion
        LOGW_WARN(Log::instance()->getLogger(), L"Unable to create : " << Utility::formatSyncPath(outputPath));
```

src/libsyncengine/jobs/network/API_v2/loguploadjob.cpp
```diff
@@ -0,0 +1,624 @@
...
+    if (!_progressCallback) {
+        _progressCallback = [](LogUploadState, int) { return true; };
+    }
+}
+
+void LogUploadJob::abort() {
+    AbstractJob::abort();
+    AppStateValue appStateValue = LogUploadState::None;
+    if (bool found = false; !ParmsDb::instance()->selectAppState(AppStateKey::LogUploadState, appStateValue, found) || !found) {
+        LOG_WARN(Log::instance()->getLogger(), "Error in ParmsDb::getAppState");
```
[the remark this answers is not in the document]
herve-er: As the job can't insert the key (since the logic of key creation is handled during the creation of `ParmsDb`), I don't see a reason to handle these differently.  

A missing key is technically a development logic error, as adding a key to the `AppStateKey` enum without adding it to the `appstate` table should not happen.(there is by the way a unit test to ensure all the  `AppStateKey` are mapped). 

For debugging purposes, `ParmsDb::selectAppState` already logs different messages depending on the cause of the failure:  

```cpp
bool ParmsDb::selectAppState(AppStateKey key, AppStateValue &value, bool &found) {
    found = false;
    std::string valueStr;

    ASSERT(queryResetAndClearBindings(SELECT_APP_STATE_REQUEST_ID))
    ASSERT(queryBindValue(SELECT_APP_STATE_REQUEST_ID, 1, static_cast<int>(key)))
    if (!queryNext(SELECT_APP_STATE_REQUEST_ID, found)) {
        LOG_WARN(_logger, "Error getting query result: " << SELECT_APP_STATE_REQUEST_ID);
        return false;
    }

    if (!found) {
        LOG_WARN(_logger, "AppStateKey not found: " << static_cast<int>(key));
        return true;
    }
    ASSERT(queryStringValue(SELECT_APP_STATE_REQUEST_ID, 0, valueStr))
    ASSERT(queryResetAndClearBindings(SELECT_APP_STATE_REQUEST_ID))

    if (!CommonUtility::stringToAppStateValue(valueStr, value)) {
        LOG_WARN(_logger, "Unable to convert value from string in selectAppState");
        return false;
    }

    return true;
}
```

Let me know if you have any thoughts on a reason to handle both cases differently.

src/libsyncengine/jobs/network/API_v2/loguploadjob.cpp
```diff
@@ -0,0 +1,624 @@
...
+    }
+
+    if (logUploadState != LogUploadState::Uploading && logUploadState != LogUploadState::Archiving) {
+        LOG_INFO(Log::instance()->getLogger(), "The operation is not in progress");
+        return;
+    }
+
+    if (bool found = false;
+        !ParmsDb::instance()->updateAppState(AppStateKey::LogUploadState, LogUploadState::CancelRequested, found) || !found) {
+        LOG_WARN(Log::instance()->getLogger(), "Error in ParmsDb::updateAppState");
```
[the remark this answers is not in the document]
herve-er: https://github.com/Infomaniak/desktop-kDrive/pull/535/files#r1954063807

src/libsyncengine/jobs/network/API_v2/loguploadjob.cpp
```diff
@@ -0,0 +1,624 @@
...
+        _runningJob->abort();
+        return;
+    }
+    LOG_WARN(Log::instance()->getLogger(), "No log upload in progress, unable to cancel the job.");
+}
+
+bool LogUploadJob::getLogDirEstimatedSize(uint64_t &size, IoError &ioError) {
+    const SyncPath logPath = Log::instance()->getLogFilePath().parent_path();
+    bool result = false;
+    for (int i = 0; i < 2; i++) { // Retry once in case a log file is archived/created during the first iteration
```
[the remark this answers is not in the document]
herve-er: The error will happen only if a file is renamed or deleted between the moment when `IoHelper::getDirectorySize` creates the iterator on the directory content and the moment it processes the renamed or deleted entry. In this case, we can retry immediately to obtain a new valid iterator.  

As the user is waiting on a loading screen while this method is running, I try to avoid any non-essential delay.

src/libsyncengine/jobs/network/API_v2/loguploadjob.cpp
```diff
@@ -0,0 +1,624 @@
...
+    std::scoped_lock lock(_runningJobMutex);
+    if (_runningJob) {
+        LOG_INFO(Log::instance()->getLogger(), "Cancelling log uplaod job.");
+        _runningJob->abort();
+        return;
+    }
+    LOG_WARN(Log::instance()->getLogger(), "No log upload in progress, unable to cancel the job.");
+}
+
+bool LogUploadJob::getLogDirEstimatedSize(uint64_t &size, IoError &ioError) {
```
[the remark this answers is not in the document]
herve-er: Done here: https://github.com/Infomaniak/desktop-kDrive/pull/535/commits/8e8b982db5fdabab95bb9aa098895cdb1afcf5f2

It does not seem to be a common practice in the codebase (input arguments are usually initialized by the caller), but it might be a good practice for robustness.

Anyway, I think we should assume that output arguments should not be used by the caller if the function fails.

src/libsyncengine/jobs/network/API_v2/loguploadjob.cpp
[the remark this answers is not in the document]
luc-guyot-infomaniak: Thanks, it makes sense.

**Sugestion.** Define a new function `bool ParmsDb::selectAppState(AppStateKey key, AppStateValue &value)` that calls `bool ParmsDb::selectAppState(AppStateKey key, AppStateValue &value, bool &found)` an returns `false` if `found` is false.

src/libsyncengine/jobs/network/API_v2/loguploadjob.cpp
```diff
@@ -0,0 +1,624 @@
...
+        return exitInfo;
+    }
+
+    if (bool found = false;
+        !ParmsDb::instance()->updateAppState(AppStateKey::LastLogUploadArchivePath, std::string(""), found) || !found) {
+        LOG_WARN(Log::instance()->getLogger(), "Error in ParmsDb::updateAppState");
+        // Do not return here because it is not a critical error, especially in this context where we are trying to send logs
+    }
+
+    if (const ExitInfo exitInfo = getTmpJobWorkingDir(_tmpJobWorkingDir); !exitInfo) {
```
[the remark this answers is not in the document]
herve-er: As we are in `LogUploadJob::init()`, this method sets `_tmpJobWorkingDir` to ensure it is available and correctly set for the next step of the job (see: `LogUploadJob::runJob()`).

I might have missed the point of your comment?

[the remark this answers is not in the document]
luc-guyot-infomaniak: This is just that we check the validity of some data at a time it is not required yet (it might become valid later). 
I presume it is simpler to handle the problem of an invalid directory at the time of initializing the object than later.
Looks fine.

src/libsyncengine/jobs/network/API_v2/loguploadjob.cpp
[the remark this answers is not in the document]
herve-er: Done here: https://github.com/Infomaniak/desktop-kDrive/pull/535/commits/567045ec2247c270e3d51543f98bfffd5edb1bb1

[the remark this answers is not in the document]
herve-er: and here: https://github.com/Infomaniak/desktop-kDrive/pull/535/commits/bb46cfb861e514cf3c90607a664cec1c23ab9a27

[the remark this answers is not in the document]
ChristopheLarchier: Disagree, a DB function can fail if the DB is corrupted, has been deleted, is not accessible etc.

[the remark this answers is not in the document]
luc-guyot-infomaniak: @ChristopheLarchier The new function could be renamed so that it does not look like a pure DB function anymore? The goal is reduce code duplication to the minimum.

src/libsyncengine/jobs/network/API_v2/loguploadjob.cpp
```diff
@@ -0,0 +1,624 @@
...
+        if (ioError == IoError::Success) {
+            size = static_cast<uint64_t>(static_cast<double>(size) *
+                                         0.8); // The compressed logs will be smaller than the original ones. We estimate at worst
+                                               // 80% of the original size.
+            return true;
+        }
+        Utility::msleep(100);
+    }
+    LOGW_WARN(Log::instance()->getLogger(),
+              L"Error in LogArchiver::getLogDirEstimatedSize: " << Utility::formatIoError(logPath, ioError).c_str());
```
ChristopheLarchier, line 105: "c_str" can be removed

src/libsyncengine/jobs/network/API_v2/loguploadjob.cpp
```diff
@@ -0,0 +1,624 @@
...
+    }
+
+    return ExitCode::Ok;
+}
+
+ExitInfo LogUploadJob::getArchiveName(SyncName &archiveName) const {
+    // Generate archive name: <drive id 1>-<drive id 2>...-<drive id N>-yyyyMMdd-HHmmss.zip
+    std::string archiveNameStr;
+    std::vector<Drive> driveList;
+    try {
```
ChristopheLarchier, line 295: What is the use of this try/catch?

src/libsyncengine/jobs/network/API_v2/loguploadjob.cpp
```diff
@@ -0,0 +1,624 @@
...
+        LOGW_WARN(Log::instance()->getLogger(),
+                  L"Error in DirectoryIterator: " << Utility::formatIoError(logDirPath, ioError).c_str());
+        return ExitCode::SystemError;
+    }
+
+    DirectoryEntry entry;
+    bool endOfDirectory = false;
+    while (dir.next(entry, endOfDirectory, ioError) && !endOfDirectory) {
+        if (entry.is_directory()) {
+            LOG_INFO(Log::instance()->getLogger(), "Ignoring temp directory " << entry.path().filename().string().c_str());
```
ChristopheLarchier, line 340: "c_str" to remove

src/libsyncengine/jobs/network/API_v2/loguploadjob.cpp
```diff
@@ -0,0 +1,624 @@
...
+    return ExitCode::Ok;
+}
+
+ExitInfo LogUploadJob::copyParmsDbTo(const SyncPath &outputPath) const {
+    const SyncPath parmsDbName = ".parms.db";
+    const SyncPath parmsDbPath = CommonUtility::getAppSupportDir() / parmsDbName;
+    DirectoryEntry entryParmsDb;
+    IoError ioError = IoError::Unknown;
+    if (!IoHelper::getDirectoryEntry(parmsDbPath, ioError, entryParmsDb)) {
+        LOGW_DEBUG(KDC::Log::instance()->getLogger(), L"Creating ");
```
ChristopheLarchier, line 375: Is it useful?

src/libsyncengine/jobs/network/API_v2/loguploadjob.cpp
```diff
@@ -0,0 +1,624 @@
...
+    std::vector<User> userList;
+    try {
+        if (ParmsDb::instance()->selectAllUsers(userList)) {
+            for (const User &user: userList) {
+                file << user.userId() << " | ";
+            }
+            file << std::endl;
+        } else {
+            file << "Unable to retrieve user ID(s)" << std::endl;
+            LOG_WARN(Log::instance()->getLogger(), "Error in ParmsDb::selectAllUsers");
```
ChristopheLarchier, line 422: false returns from ParmsDb functions cannot be ignored.

src/libsyncengine/jobs/network/API_v2/loguploadjob.cpp
```diff
@@ -0,0 +1,624 @@
...
+
+        file << "Drive ID(s): ";
+        std::vector<Drive> driveList;
+        if (ParmsDb::instance()->selectAllDrives(driveList)) {
+            for (const Drive &drive: driveList) {
+                file << drive.driveId() << " | ";
+            }
+            file << std::endl;
+        } else {
+            file << "Unable to retrieve drive ID(s)" << std::endl;
```
ChristopheLarchier, line 433: Same as last comment.

src/libsyncengine/jobs/network/API_v2/loguploadjob.cpp
```diff
@@ -0,0 +1,624 @@
...
+    return ExitCode::Ok;
+}
+
+ExitInfo LogUploadJob::generateArchive(const SyncPath &directoryToCompress, const SyncPath &destPath,
+                                       const SyncName &archiveNameWithoutExtension, SyncPath &finalPath) {
+    // Generate the archive
+    const SyncPath archivePath = destPath / (archiveNameWithoutExtension + Str2SyncName(".zip"));
+    int err = 0;
+    zip_t *archive = zip_open(archivePath.string().c_str(), ZIP_CREATE | ZIP_EXCL, &err);
+    if (err != ZIP_ER_OK) {
```
ChristopheLarchier, line 457: I would prefer using a pattern like this one (see https://libzip.org/documentation/zip_open.html).
If err != ZIP_ER_OK, archive will be NULL, so what is the behaviour of "zip_strerror(archive)"?

   if ((za = zip_open(name, 0, &err)) == NULL) {
        zip_error_t error;
        zip_error_init_with_code(&error, err);
        fprintf(stderr, "%s: cannot open zip archive '%s': %s\n",
	        progname, name, zip_error_strerror(&error));
        zip_error_fini(&error);
        return -1;
    }

src/server/appserver.cpp
```diff
@@ -2061,86 +2060,23 @@ void AppServer::sendLogUploadStatusUpdated(LogUploadState status, int percent) {
...
-        LOG_WARN(_logger, "Error in LogArchiverHelper::sendLogToSupport: code=" << exitCode << " cause=" << exitCause);
-        addError(Error(errId(), ExitCode::LogUploadFailed, exitCause));
-    }
-    sendLogUploadStatusUpdated(exitCode == ExitCode::Ok ? LogUploadState::Success : LogUploadState::Failed, 0);
-    return {exitCode, exitCause};
+    const std::function<void(UniqueId)> jobResultCallback = [this, logUploadJob](const UniqueId id) {
+        if (const ExitInfo exitInfo = logUploadJob->exitInfo(); !exitInfo && exitInfo.code() != ExitCode::OperationCanceled) {
+            LOG_WARN(_logger, "Error in LogArchiverHelper::sendLogToSupport: " << exitInfo);
+            addError(Error(errId(), ExitCode::LogUploadFailed, ExitCause::Unknown));
+            addError(Error(errId(), exitInfo.code(), exitInfo.cause()));
```
ChristopheLarchier, line 2076: Why this double addError call?

src/server/appserver.cpp
```diff
@@ -2916,7 +2852,7 @@ void AppServer::handleCrashRecovery(bool &shouldQuit) {
     } else {
         timestampStr = std::to_string(selfRestarterNoCrashDetected);
     }
-
+    bool found = false;
     KDC::ParmsDb::instance()->updateAppState(AppStateKey::LastServerSelfRestartDate, timestampStr, found);
```
ChristopheLarchier, line 2856: ParmsDb methods false return code must be managed separately from !found.

src/libsyncengine/jobs/network/API_v2/loguploadjob.cpp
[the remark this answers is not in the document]
herve-er: In the case of the `appState` method, a "not found" error is equivalent to a corrupted database row and cols are static and can/should only change with a parmsDb migration.  

> A missing key is technically a development logic error, as adding a key to the `AppStateKey` enum without adding it to the `appState` table should not happen. (There is, by the way, a unit test to ensure that all `AppStateKey` values are mapped.)  

For the caller, there is no significant difference between a missing/corrupted `parmsDb` and a missing `appState` key (which is supposed to be "hard-coded").  

Currently, both cases are handled as follows:

- In the case of log upload, if this happens, we will attempt to continue the upload even if we cannot access the database, as the user might need to send their logs due to database corruption.  
- In `logUsefulInformation`, we log the error, but no error is raised since it is not critical.  
- In crash recovery, we will assume that we are not recovering from a crash.  
- For all other cases, if the `appStateKey` is not found in the database, we will return a `dbError` and display it to the user.

src/libsyncengine/jobs/network/API_v2/loguploadjob.cpp
herve-er: Done here: https://github.com/Infomaniak/desktop-kDrive/pull/535/commits/78b0ec776722859b5799e307b9190712f91f3ec3
(This isn't really new code, I moved it from logarchiver to loguploadjob).

src/libsyncengine/jobs/network/API_v2/loguploadjob.cpp
herve-er: Done here: https://github.com/Infomaniak/desktop-kDrive/pull/535/commits/78b0ec776722859b5799e307b9190712f91f3ec3
(This isn't really new code, I moved it from logarchiver to loguploadjob).

src/libsyncengine/jobs/network/API_v2/loguploadjob.cpp
```diff
@@ -0,0 +1,624 @@
...
+        LOGW_WARN(Log::instance()->getLogger(), L"Unable to create : " << Utility::formatSyncPath(outputPath));
+        return ExitCode::Ok; // We don't want to stop the process if we can't create the file
+    }
+    file << "OS Name: " << osName << std::endl;
+    file << "OS Architecture: " << osArch << std::endl;
+    file << "App Version: " << appVersion << std::endl;
+    file << "User ID(s): ";
+
+    std::vector<User> userList;
+    try {
```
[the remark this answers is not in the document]
herve-er: Same -> https://github.com/Infomaniak/desktop-kDrive/pull/535#discussion_r1959340539

src/libsyncengine/jobs/network/API_v2/loguploadjob.cpp
herve-er: As we are in the context of log upload, I've tried to make it as resilient as possible. `ParmsDb::instance()->selectAllDrives` may fail, but we should still be able to send the log.  

If `ParmsDb` is completely missing or too corrupted, we will fail anyway later when starting the upload job and will exit with a proper `dbError`.

src/libsyncengine/jobs/network/API_v2/loguploadjob.cpp
herve-er: https://github.com/Infomaniak/desktop-kDrive/pull/535#discussion_r1959363586

src/libsyncengine/jobs/network/API_v2/loguploadjob.cpp
herve-er: Done here: https://github.com/Infomaniak/desktop-kDrive/pull/535/commits/a030cc8043eca15449009a203d5c06841ef9fe3d

src/server/appserver.cpp
herve-er: This was initially intended to provide more information to the user, but it might not be desirable. Moreover, the error description for server-level errors will not really help the user resolve the issue, so I removed it here: https://github.com/Infomaniak/desktop-kDrive/pull/535/commits/1da82fb51284e68e72a5a4e1fa33f866fa94528d

src/server/appserver.cpp
herve-er: [For the `AppState` related method, how should they be handled? ](https://github.com/Infomaniak/desktop-kDrive/pull/535#discussion_r1959298280)

src/libsyncengine/jobs/network/API_v2/loguploadjob.cpp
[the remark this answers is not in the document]
ChristopheLarchier: You could wrap ParmsDb::instance()->selectAppState in a LogUploadJob function that throws an addError if it returns false.
I would prefer not to have a version of ParmsDb::selectAppState without the "found" parameter because all the DB functions are on the same model.
Idem for updateAppState.

src/server/appserver.cpp
```diff
@@ -2843,7 +2778,7 @@ void AppServer::logUsefulInformation() const {
 
     // Log app ID
     AppStateValue appStateValue = "";
-    if (bool found = false; !ParmsDb::instance()->selectAppState(AppStateKey::AppUid, appStateValue, found) || !found) {
+    if (!ParmsDb::instance()->selectAppState(AppStateKey::AppUid, appStateValue)) {
```
ChristopheLarchier, line 2781: if !found, this is not a DbError.
(the problem existed before)
You can wrap ParmsDb::instance()->selectAppState in an AppServer function to avoid code duplication.

src/libsyncengine/jobs/network/API_v2/loguploadjob.cpp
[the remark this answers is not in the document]
herve-er: I reverted the commit that implemented the `parmsDb` wrapper.  
I handle "not found" in this commit: [[8b17c60](https://github.com/Infomaniak/desktop-kDrive/pull/535/commits/8b17c6028022c2c14e9ce226a32acda60bd83eb0)](https://github.com/Infomaniak/desktop-kDrive/pull/535/commits/8b17c6028022c2c14e9ce226a32acda60bd83eb0).

src/libsyncengine/jobs/network/API_v2/loguploadjob.cpp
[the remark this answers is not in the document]
herve-er: https://github.com/Infomaniak/desktop-kDrive/pull/535#discussion_r1961478471

src/libsyncengine/jobs/network/API_v2/loguploadjob.cpp
herve-er: https://github.com/Infomaniak/desktop-kDrive/pull/535#discussion_r1961478471

src/server/appserver.cpp
herve-er: https://github.com/Infomaniak/desktop-kDrive/pull/535/commits/8b17c6028022c2c14e9ce226a32acda60bd83eb0

src/server/appserver.cpp
herve-er: https://github.com/Infomaniak/desktop-kDrive/pull/535/commits/8b17c6028022c2c14e9ce226a32acda60bd83eb0

src/server/updater/updatechecker.cpp
```diff
@@ -114,7 +114,7 @@ void UpdateChecker::versionInfoReceived(UniqueId jobId) {
 
 ExitCode UpdateChecker::generateGetAppVersionJob(std::shared_ptr<AbstractNetworkJob> &job) {
     AppStateValue appStateValue = "";
-    if (bool found = false; !ParmsDb::instance()->selectAppState(AppStateKey::AppUid, appStateValue, found) || !found) {
+    if (!ParmsDb::instance()->selectAppState(AppStateKey::AppUid, appStateValue)) {
```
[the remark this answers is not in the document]
herve-er: I would said this is a `Exitcode::DbError` & `ExitCause::DbEntryNotFound` but `generateGetAppVersionJob` only return an ExitCode.

src/libsyncengine/jobs/network/API_v2/loguploadjob.cpp
herve-er: Done here: https://github.com/Infomaniak/desktop-kDrive/pull/535/commits/aeaf4ef0f06e64376e58c8065e91930c848ff839

src/libsyncengine/jobs/network/API_v2/loguploadjob.cpp
herve-er: https://github.com/Infomaniak/desktop-kDrive/pull/535/commits/aeaf4ef0f06e64376e58c8065e91930c848ff839

src/server/updater/updatechecker.cpp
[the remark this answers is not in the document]
ChristopheLarchier: ExitCause::DbEntryNotFound is an ExitCode::DataError (not DbError).
I found 4 "addError(Error(errId(), ExitCode::DbError, ExitCause::DbEntryNotFound));" in appserver.cpp, this should be fixed.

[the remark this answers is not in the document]
herve-er: It seems that we are going out of the scope of this PR. I have created a dedicated card: https://infomaniak.atlassian.net/browse/KDESKTOP-1577

ClementKunz (approved):

PR #538, opened 2025-02-13
KDESKTOP-1564 Add distribution channel tag in sentry.

herve-er:


ChristopheLarchier (dismissed):

src/libcommon/utility/types.h
```diff
@@ -504,7 +504,7 @@ enum class UpdateState {
 };
 std::string toString(UpdateState e);
 
-enum class DistributionChannel { Prod, Next, Beta, Internal, Unknown };
+enum class DistributionChannel { Prod, Next, Beta, Internal, Legacy, Unknown };
```
ClementKunz, line 507: I am not sure to understand why we need this `Legacy` channel. This does not refer to a real channel.

src/libcommon/log/sentry/handler.cpp
```diff
@@ -445,6 +450,10 @@ void Handler::writeEvent(const std::string &eventStr, bool crash) noexcept {
     }
 }
 
+void Handler::setDistributionChannel(const DistributionChannel channel) {
+    setTag("distribution_channel", toString(channel));
```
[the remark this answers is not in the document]
herve-er: Done here: https://github.com/Infomaniak/desktop-kDrive/pull/538/commits/1ab0bdc0805e35d2624cc93e976fab7454ca6283

src/server/updater/abstractupdater.cpp
```diff
@@ -93,6 +99,38 @@ bool AbstractUpdater::isVersionSkipped(const std::string& version) {
...
+    }
+
+    if (allVersions.contains(DistributionChannel::Internal) &&
+        allVersions.at(DistributionChannel::Internal).fullVersion() == currentVersion) {
+        return DistributionChannel::Internal;
+    }
+
+    if (allVersions.contains(DistributionChannel::Prod) &&
+        CommonUtility::isVersionLower(currentVersion, allVersions.at(DistributionChannel::Prod).fullVersion())) {
+        return DistributionChannel::Legacy;
```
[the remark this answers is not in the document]
herve-er: > Could be fine for now, but I strongly believe that we should include it in the build somehow.  

It implies doing a new build each time a version reaches a new channel. If a ✨version✨ is bug-free from the internal stage to production, we will have to build it four times.  

This is possible, but it should be discussed with the team.  
Additionally, it might be better to wait for a smoother build/release process if we do so(full CI?).

src/libcommon/utility/types.h
herve-er: Since `Legacy` is a read-only channel, meaning it's not an actual distribution option but rather a status for outdated versions, it could makes sense to separate concerns into two distinct enums?

1. **`DistributionChannel `** – Defines the channels an app can actively subscribe to.  
2. **`VersionChannel `** – Defines the status of a given app version, including whether it is actively deployed or considered legacy.  

Here’s the improved version:  

```cpp
// Represents the distribution channel the app is subscribed to or intends to subscribe to.
enum class DistributionChannel { Production, Next, Beta, Internal, Unknown };

// Represents the status of the currently running version of the app.
enum class VersionChannel { Production, Next, Beta, Internal, Legacy, Unknown };
```

src/libcommon/log/sentry/handler.cpp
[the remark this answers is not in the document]
ClementKunz: Fine for me like that but it does not have to be an attribute of the class, just a static const variable defined in the cpp file.
Maybe this is something to be discussed in team meeting to make sure we all do the same

src/libcommon/utility/types.h
ClementKunz: In that case, I would keep only `VersionChannel`. The name is more generic and can apply for both the distribution channel we are monitoring and the current version.
Perhaps even something more generic like `VersionType`. What do you think?

src/server/updater/abstractupdater.cpp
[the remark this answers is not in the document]
ClementKunz: Yeah that would be definitively not ideal either.
Let's keep that for later

src/libcommon/log/sentry/handler.cpp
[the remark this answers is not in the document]
herve-er: As discussed: https://github.com/Infomaniak/desktop-kDrive/pull/538/commits/f48a1ba0145eb688b9e539ea15afdcb94441be34  

I also got rid of the `toString(const DistributionChannel&)` method defined in `types.h` and implemented a custom `switch` to avoid any issues if the `toString` method is later modified, as this could break some dashboards/filters in Sentry.

src/libcommon/utility/types.h
herve-er: `VersionType` is maybe a bit too abstract. For example:  
`RequestNum::UPDATER_CHANGE_CHANNEL` would become `RequestNum::UPDATER_CHANGE_TYPE` or `RequestNum::UPDATER_CHANGE_VERSION_TYPE`, which seems way less intuitive. I would like to keep the word "channel" even though I understand it also brings some confusion because "Legacy" is not a subscribable channel.  

`VersionChannel` sounds good to me.

ClementKunz: Yeah, fine for me too. Let's go with `VersionChannel `

herve-er: Done here: https://github.com/Infomaniak/desktop-kDrive/pull/538/commits/7fae3ba3bd7b7211d6e1f44764809a27c925913d

ClementKunz (approved):

PR #545, opened 2025-02-19
[HOTFIX] Initialize SyncPal even if Vfs has not started yet

luc-guyot-infomaniak:
After installation, and depending on the time at which `LiteSync` authorizations are given by a MacOS user, the _Vfs_ start failure errors 
Capture d’écran 2025-02-19 à 14 43 40

Capture d’écran 2025-02-19 à 14 45 09

remains persistently in _desktop-kDrive GUI_ even if the _Vfs_ is subsequently and successfully initialized.

The issue has been frequently reproduced.

### DIFF
--- a/src/server/appserver.cpp
+++ b/src/server/appserver.cpp
@@ -1132,7 +1132,6 @@ void AppServer::onRequestReceived(int id, RequestNum num, const QByteArray &para
 
                 if (ExitInfo exitInfo = tryCreateAndStartVfs(sync); !exitInfo) {
                     LOG_WARN(_logger, "Error in tryCreateAndStartVfs for syncDbId=" << sync.dbId() << " " << exitInfo);
-                    return;
                 }
 
                 // Create and start SyncPal


ChristopheLarchier (approved):

herve-er (changes requested):
Not sure I fully understand the issue, but in `AppServer::startSyncs`, we also do not initialize the syncpal if the VFS fails to start. Shouldn't we have the same behavior as the one implemented in this PR?

herve-er (approved):

luc-guyot-infomaniak:
> Not sure I fully understand the issue, but in `AppServer::startSyncs`, we also do not initialize the syncpal if the VFS fails to start. Shouldn't we have the same behavior as the one implemented in this PR?

There is mechanism to suppress _Vfs_ startup errors (which are expected on MacOS) once the *Vfs* is properly set up. It consists in a time calling the `Qt` slot `onRestartSyncs` at regular intervals. The issue is that this slot expects the `_syncPalMap` to be filled whereas the `return` statement that is removed in this PR prevents the map to be filled.

Regarding the `continue` statement in `startSyncs`, we have checked that it is harmless, because on the long run, _syncPalMap_ gets eventually filled.

PR #547, opened 2025-02-20
Kdesktop 1516 refactoring move server code into server module

ChristopheLarchier:


src/libcommon/utility/types.h
```diff
@@ -17,17 +17,22 @@
  */
 #pragma once
 
-#include <string>
+#include <QApplication> // /!\ Moc bug on Linux: filesystem must be included after QApplication
```
ClementKunz, line 20: `types.h` is used everywhere, but we should not include Qt in lib like `libsyncengine`. Is ot ok anyway to include Qt's file here?

ChristopheLarchier: QDebug is already included...

If QApplication is not included here, it must be included in dozens of other places.
I added a #define and a comment to remove it when migrating to a newer version of Qt.

https://github.com/Infomaniak/desktop-kDrive/pull/547/commits/e9fa62c81a92806e51de810379968fa708af6bf1

PR #557, opened 2025-02-26
KDESKTOP-1125-Windows-Removes-redundant-casts-in-SqliteQuery-syncNameValue-and-SqliteQuery-syncNameValue

ClementKunz:
Quick refactoring of `SqliteQuery`


ClementKunz:
Addressing Sonar issue require deeper refactoring. Let's keep it for a dedicated task

src/libcommonserver/db/sqlitequery.cpp
```diff
@@ -198,20 +196,20 @@ bool SqliteQuery::nullValue(int index) const {
     return sqlite3_column_type(_stmt.get(), index) == SQLITE_NULL;
 }
 
-const std::string SqliteQuery::stringValue(int index) const {
+std::string SqliteQuery::stringValue(int index) const {
```
luc-guyot-infomaniak, line 199:
```suggestion
std::string SqliteQuery::stringValue(const int index) const {
```
To please SonarCloud.

PR #562, opened 2025-02-27
KDESKTOP-1382-BUG-Migrated-parameters-database-cause-encoding-corruption-breaking-file-status-displayed

ClementKunz:
No kDrive menu available if the local path to the sync folder contains NFD encoded characters. This came from the fact that `QFileDialog::getExistingDirectory`return a NFC encoded path.


herve-er (commented):
It seems like the issue will still occur if a user has a mixed path with both NFC and NFD characters, but I don't see any reasonable solution to this...

However, does this mean that up until now, it has not been possible to have a correctly working sync on macOS if there was an accent in the sync path?

src/server/requests/serverrequests.cpp
```diff
@@ -1490,14 +1490,18 @@ ExitCode ServerRequests::addSync(int driveDbId, const QString &localFolderPath,
     Sync sync;
     sync.setDbId(syncDbId);
     sync.setDriveDbId(driveDbId);
-    sync.setLocalPath(QStr2Path(localFolderPath));
+    SyncName localPath = Str2SyncName(localFolderPath.toStdString());
+#ifdef __APPLE__
+    (void) Utility::normalizedSyncName(localFolderPath.toStdString(), localPath, Utility::UnicodeNormalization::NFD);
```
luc-guyot-infomaniak, line 1495: We should use `normalizedSyncPath` but this function does support the third argument (yet).

src/server/requests/serverrequests.cpp
```diff
@@ -1490,14 +1490,18 @@ ExitCode ServerRequests::addSync(int driveDbId, const QString &localFolderPath,
     Sync sync;
     sync.setDbId(syncDbId);
     sync.setDriveDbId(driveDbId);
-    sync.setLocalPath(QStr2Path(localFolderPath));
+    SyncName localPath = Str2SyncName(localFolderPath.toStdString());
```
luc-guyot-infomaniak, line 1493:
```suggestion
    SyncName localPath = QStr2Path(localFolderPath);
```
To keep a path type for a variable that stores a path.

src/server/requests/serverrequests.cpp
```diff
@@ -1490,14 +1490,18 @@ ExitCode ServerRequests::addSync(int driveDbId, const QString &localFolderPath,
...
+#endif
+    sync.setLocalPath(localPath);
     sync.setTargetPath(QStr2Path(serverFolderPath));
     sync.setTargetNodeId(serverFolderNodeId.toStdString());
     sync.setPaused(false);
 
     // Check vfs support
-    QString fsName(KDC::CommonUtility::fileSystemName(SyncName2QStr(sync.localPath().native())));
-    bool supportVfs = (fsName == "NTFS" || fsName == "apfs");
+    const QString fsName(CommonUtility::fileSystemName(SyncName2QStr(sync.localPath().native())));
```
luc-guyot-infomaniak, line 1503: It's surprising that that we call two functions that operate on names and not on paths.

src/server/requests/serverrequests.cpp
luc-guyot-infomaniak: By the way, why is special handling required for MacOSX? A comment about it?

ClementKunz: Done in 74f95fc7778e687ed0be1733f5c8ae3bf5b60b73

src/server/requests/serverrequests.cpp
ClementKunz: Done in 74f95fc7778e687ed0be1733f5c8ae3bf5b60b73

src/server/requests/serverrequests.cpp
ClementKunz: Done in 74f95fc7778e687ed0be1733f5c8ae3bf5b60b73

ChristopheLarchier (approved):

luc-guyot-infomaniak (approved):

PR #573, opened 2025-03-03
Change ASSERT macro

ClementKunz:
As discussed in team meeting, `ASSERT` is not a real `assert` and should be renamed. The new proposed name is `LOG_IF_FAIL`


src/libcommonserver/CMakeLists.txt
```diff
@@ -28,7 +28,7 @@ add_subdirectory(vfs)
 set(libcommonserver_SRCS
     # Utility
     utility/utility.h utility/utility.cpp
-    utility/asserts.h
+        utility/log_if_fail.h
```
herve-er, line 31:
```suggestion
    utility/log_if_fail.h
```

herve-er, line 31: We generally name files all in lowercase and all attached, i.e., `logiffail.h`.  

We seem to have made some exceptions when we have multiple instances of the "same" file for different OS (`utility_win`, `utility_mac`, ...), but that is not the case here.  

Even if I agree that this isn't the most readable format.

herve-er (approved):

PR #574, opened 2025-03-03
KDESKTOP-1521-Linux-QA-feedback-Events-inside-a-moved-folder-are-not-detected

ClementKunz:
In Linux, if a parent folder is renamed, all the changes done inside it are omitted until app restart.


ClementKunz:
Will be fixed in https://github.com/Infomaniak/desktop-kDrive/pull/575

PR #571, opened 2025-03-03
KDesktop 1586: 0-byte files to develop

herve-er:
Due to recent changes in download job beahviour, a 0-byte file download now leads to a `BackError InvalidSize`, which  leads in turn to the sync failure.

Added a new test case in `TestNetworkJobs::testDownload` to verify that downloading an empty file is handled correctly.

PR #577, opened 2025-03-03
Kdesktop 1586 0 bytes

herve-er:


test/libsyncengine/jobs/network/testnetworkjobs.cpp
```diff
@@ -466,6 +466,31 @@ void TestNetworkJobs::testDownload() {
                                          ExitInfo(ExitCode::SystemError, ExitCause::NotEnoughDiskSpace), exitInfo);
         }
     }
+
+    // Empty file
+    {
+        const LocalTemporaryDirectory temporaryDirectory("tmp");
+        const SyncPath local0bytesFilePath = temporaryDirectory.path() / "0bytes.txt";
+        const RemoteTemporaryDirectory remoteTmpDir(_driveDbId, _remoteDirId, "testDownload0bytesFile");
+        std::ofstream(local0bytesFilePath).close();
```
herve-er, line 475: > @luc-guyot-infomaniak https://github.com/Infomaniak/desktop-kDrive/pull/571#discussion_r1977499900   


 The suggested change does not seem to work as the file isn't created. I have made the above change instead, which is working and is closer to your suggestion.

luc-guyot-infomaniak: The pattern I had in mind was `{ std::ofstream ofs(path); }` (it is frequently used in the unit tests). My suggestion was indeed flawed because of the extra inner braces; all apologies!

luc-guyot-infomaniak: You are right, the test name the stream variables. I thought this would work anyway (tested in a sandbox on my Mac with `g++` and `CFLAGS=-std=c++20`). Compilers might optimize the code by removing the statement depending on the compilation flags.

luc-guyot-infomaniak (approved):

test/libsyncengine/jobs/network/testnetworkjobs.cpp
herve-er: I tested this more thoroughly, and it turns out that if we pass a `char*` as an argument, the file is always created:  

```cpp
std::ofstream file("C:\\Users\\Herve\\Documents\\test1.txt");
file.close();

{ std::ofstream file("C:\\Users\\Herve\\Documents\\test2.txt"); }
{ std::ofstream("C:\\Users\\Herve\\Documents\\test3.txt"); }
{ std::ofstream{"C:\\Users\\Herve\\Documents\\test4.txt"}; }
std::ofstream("C:\\Users\\Herve\\Documents\\test5.txt").close();
```

However, when using a `std::filesystem::path`, there is one case where the file is not created (see below):  

```cpp
const std::filesystem::path local0bytesFilePath1 = "C:\\Users\\Herve\\Documents\\0bytes1.txt";
const std::filesystem::path local0bytesFilePath2 = "C:\\Users\\Herve\\Documents\\0bytes2.txt";
const std::filesystem::path local0bytesFilePath3 = "C:\\Users\\Herve\\Documents\\0bytes3.txt";
const std::filesystem::path local0bytesFilePath4 = "C:\\Users\\Herve\\Documents\\0bytes4.txt";
const std::filesystem::path local0bytesFilePath5 = "C:\\Users\\Herve\\Documents\\0bytes5.txt";

std::ofstream file(local0bytesFilePath1);
file.close();

{ std::ofstream file(local0bytesFilePath2); }
{ std::ofstream(local0bytesFilePath3); }  // <- This is the case that matches what we had in the PR
{ std::ofstream{local0bytesFilePath4}; }
std::ofstream(local0bytesFilePath5).close();
```
So it seems that `{ std::ofstream(path); }` does not always work when `path` is a `std::filesystem::path`, which could explain why the file wasn’t created in the initial implementation. This might indeed depend on compiler optimizations.

luc-guyot-infomaniak: The issue with `{ std::ofstream(local0bytesFilePath3); }` is actually not related to compiler options:
```
 warning: parentheses were disambiguated as redundant parentheses around declaration of variable named 'path' [-Wvexing-parse]
```

PR #581, opened 2025-03-04
[KDESKTOP-1129] Designs updaters as singletons

luc-guyot-infomaniak:
The initial impetus of these changes comes from a `cppcheck` warning about the `static` method `AbstractUpdater::unskipVersion` that was redefined as another `static` method of the derived class `SparkelUpdater`.

The `static` qualifier of the method `unskipVersion` is now removed so that polymorphism can be used appropriately. Each derived class of `AbstractUpdater` is additionally turned into a singleton.

### DIFF
--- a/src/server/updater/updatemanager.h
+++ b/src/server/updater/updatemanager.h
@@ -48,7 +48,7 @@ class UpdateManager final : public QObject {
         void startInstaller() const;
         void setQuitCallback(const std::function<void()> &quitCallback) const { _updater->setQuitCallback(quitCallback); }
 
-        std::shared_ptr<AbstractUpdater> updater() { return _updater; };
+        std::shared_ptr<AbstractUpdater> updater() const { return _updater; };
 
     signals:
         void updateAnnouncement(const QString &title, const QString &msg);
--- a/src/server/updater/updatemanager.h
+++ b/src/server/updater/updatemanager.h
@@ -48,6 +48,8 @@ class UpdateManager final : public QObject {
         void startInstaller() const;
         void setQuitCallback(const std::function<void()> &quitCallback) const { _updater->setQuitCallback(quitCallback); }
 
+        const std::shared_ptr<AbstractUpdater> updater() { return _updater; };
+
     signals:
         void updateAnnouncement(const QString &title, const QString &msg);
         void requestRestart();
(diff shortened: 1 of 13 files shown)


ClementKunz:
If the failed quality gate raised by Sonar is something that you want to be fixed in this PR? Or should we start reviewing now?

luc-guyot-infomaniak:
> Is the failed quality gate raised by Sonar is something that you want to be fixed in this PR? Or should we start reviewing now?

SonarCloud complains about Singletons! (The recommendation is to make the static members named `_instance` `constexpr` variables). So I would prefer to ignore these two complains. There is another comment about `catch(...)` that is legitimate (the same pattern is already used in several places of the code). It requires some thoughts: I am preparing new changes.

src/server/updater/updatemanager.h
```diff
@@ -66,7 +68,7 @@ class UpdateManager final : public QObject {
 
         void onUpdateStateChanged(UpdateState newState);
 
-        std::unique_ptr<AbstractUpdater> _updater;
+        static std::shared_ptr<AbstractUpdater> _updater;
```
ClementKunz, line 71: Why changing it to `shared_ptr` instead of `unique_ptr`? There should be only 1 updater and its instance should be owned by the `UpdateManager`

test/server/updater/testabstractupdater.cpp
```diff
@@ -28,6 +37,18 @@
 
 namespace KDC {
 
+namespace {
+void skipVersion() {
```
ClementKunz, line 41: Incoherence between `skip` and `unskip`

src/server/updater/updatemanager.h
luc-guyot-infomaniak: Because the derived classes of `AbstractUpdater` are singletons, the `UpdateManager` does not own an updater object per se, but only a reference to it.

test/server/updater/testabstractupdater.cpp
luc-guyot-infomaniak: Ooops ! Thanks. Fxied with https://github.com/Infomaniak/desktop-kDrive/pull/581/commits/9a2453975f1223025ee5c4bcd20dcf7b4d7f573b.

ClementKunz (approved):

PR #584, opened 2025-03-05
Redirection URL should be read from the response header

ChristopheLarchier:


src/libsyncengine/jobs/network/abstractnetworkjob.cpp
```diff
@@ -498,26 +498,13 @@ bool AbstractNetworkJob::receiveResponse(const Poco::URI &uri) {
...
-        return false;
-    }
-
+bool AbstractNetworkJob::followRedirect() {
+    // Get redirection URL
     std::string redirectUrl;
-    Poco::XML::Node *pNode = pDoc->getNodeByPath(redirectUrlPathKey);
-    if (pNode != nullptr) {
-        redirectUrl = pNode->innerText();
+    try {
```
luc-guyot-infomaniak, line 504: **Suggestion.** Use `find` instead of `get` to make clear that the only possible *issue* is that the key `Location` is not found, which is not exceptional. Using exceptions in non-exceptional cases can degrade performances (https://stackoverflow.com/questions/16784601/does-try-catch-block-decrease-performance).

ChristopheLarchier: I finally kept the **get** but with the default value as the 2nd parameter.
8f4e69b4de0510fe4226305808693ed1a275b9f3

luc-guyot-infomaniak (commented):
If feasible (mocking all relevant methods, checking only that `sendRequest` is called with the expected uri), we may unit test the `followRedirect()` method.

ChristopheLarchier:
> If feasible (mocking all relevant methods, checking only that `sendRequest` is called with the expected uri), we may unit test the `followRedirect()` method.

No idea how to do this, the expected URI does not have a fixed value.
For instance, for 2 successive downloads of the same file:
https://1-next-v3-9.download.kdrive.infomaniak.com/2/download/01957f76-d341-7b05-b47b-152a6b907759/file
https://1-next-v3-9.download.kdrive.infomaniak.com/2/download/01957f79-74f1-75d1-9c56-d1688c63b91d/file

Note that there are already tests of the download process (TestNetworkJobs::testDownload).

ChristopheLarchier:
Remaining Sonar issues are not something that needs to be fixed, I think.

luc-guyot-infomaniak (approved):

PR #585, opened 2025-03-05
KDESKTOP-1572-Duplication-in-case-of-file-access-error

ClementKunz:
This PR aims at fixing the app behavior if a change cannot be propagated due to temporary access denied error (e.g. : a file is opened in Excel).
The fix is to keep items that have been temporarily blacklisted in the DB, so that the change can be propagated as soon as the app can access the item again.


src/libsyncengine/syncpal/tmpblacklistmanager.cpp
```diff
@@ -22,19 +22,19 @@
 
 #include <unordered_set>
 
-static const int oneHour = 3600;
+static constexpr int oneHour = 3600;
 
 namespace KDC {
 
-TmpBlacklistManager::TmpBlacklistManager(std::shared_ptr<SyncPal> syncPal) : _syncPal(syncPal) {
+TmpBlacklistManager::TmpBlacklistManager(const std::shared_ptr<SyncPal> &syncPal) : _syncPal(syncPal) {
```
ChristopheLarchier, line 29: Why pass syncPal shared ptr by reference?

ClementKunz: That is highlighted by my IDE (CLion). Which does make sens since `std::shared_ptr` is not directly a pointer but an object managing the smart pointer.
But that should not change anything in terms of optimization, I agree.

ChristopheLarchier: IMO TmpBlacklistManager doesn't need a shared_ptr of SyncPal because the 2 objects live in the same thread.
Moreover, pass a "const std::shared_ptr<SyncPal>&" means :
- the shared_ptr counter is not incremented 
- the object (SyncPal) can be modified (you don't pass "std::shared_ptr<const SyncPal>&")
In this case, why not pass a "SyncPal*"?
See https://medium.com/@vgasparyan1995/pass-by-value-vs-pass-by-reference-to-const-c-f8944171e3ce

ClementKunz: I do agree 100% with your argument. I have rolled back this change in b31c77695a979cec9a1df32a6edba22fd833027c.

ChristopheLarchier (approved):

src/libsyncengine/propagation/executor/executorworker.cpp
```diff
@@ -1017,19 +1017,20 @@ ExitInfo ExecutorWorker::handleMoveOp(SyncOpPtr syncOp, bool &ignored, bool &byp
             }
         }
 
-        if (ExitInfo exitInfo = propagateMoveToDbAndTree(syncOp); !exitInfo) {
```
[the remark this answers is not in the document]
ClementKunz: I can't remember why nor I find a good reason... So I reverted this change in 407e5f6edec3b1c22c77bd731e08c785c3919b41

ChristopheLarchier (approved):

PR #589, opened 2025-03-07
[KDESKTOP-1233] Reactivate CI analysis on the Windows runner

luc-guyot-infomaniak:


.github/workflows/windows.yml
```diff
@@ -57,49 +58,61 @@ jobs:
...
+          & "cov01.exe" "-s" # show status
+
           & "covxml.exe" "-f" "$env:COVFILE" "-o" "coverage.xml"
 
-      - name: Install sonar-scanner
-        if: false
-        uses: SonarSource/sonarcloud-github-c-cpp@v3
-        env:
-          SONAR_HOST_URL: https://sonarcloud.io
+          & "cov01.exe" "-1" # coverage on
```
herve-er, line 89: Do we really need to turn on the coverage here?

luc-guyot-infomaniak: Good question! This block was added after several trials and errors. 
As I read it now, it might be useful for the next step, but this is rather speculative by now. 

Let me try to remove it and check if the build is still successful.

luc-guyot-infomaniak: So, it's actually required! (It helps SonarCloud detecte the compiler used during the build).

ChristopheLarchier (approved):

PR #594, opened 2025-03-11
A drive without any synchronisation appears with an error status in the user drive list

ChristopheLarchier:


src/gui/synthesispopover.cpp
```diff
@@ -678,32 +678,39 @@ void SynthesisPopover::refreshStatusBar(const DriveInfoClient &driveInfo) {
...
                 syncsInPropagationStep++;
             }
             if (syncInfo.virtualFileMode() != VirtualFileMode::Off) {
                 statusInfo._liteSyncActivated = true;
             }
+
+            if (syncsProcessed++ == 0) {
+                statusInfo._estimatedRemainingTime = syncInfo.estimatedRemainingTime();
+                statusInfo._status = syncInfo.status();
+                statusInfo._syncStep = syncInfo.step();
```
ClementKunz, line 702: This part is a bit hard to understand on first reading. Wouldn't the initial values applied when creating `statusInfo` be sufficient?

ChristopheLarchier: `statusInfo` variable is the synthesis of the statusInfo of all the syncs.
This code is in a "for" loop on the syncs and "syncsProcessed++ == 0" for the first iteration.

ClementKunz (approved):

PR #591, opened 2025-03-07
Log source file names instead of source file paths

luc-guyot-infomaniak:

PR #595, opened 2025-03-11
Address several complex move scenario issues.

herve-er:
# Main Changes in this PR

### Issue  
If a node is moved, but a temporary equivalent exists (due to another operation further down the tree), when merging the real and temporary nodes, the new name is lost.  

#### Example  
**Initial situation:**
```text
|_ A
|_ A1
|_ B
```
**Final affected side situation:**
```text
|_ A
|   |_ A1b
|       |_ B
```
In the update tree generation:
1. We first treat the move of `B` to `A1`, which creates temporary nodes for `A` and `A1`.
2. We then treat the move (rename) of `A1` to `A1b`, creating a real node.
3. In the last step, we convert the temporary node to a real node by transferring the real node's information (change events, move origins if applied, child nodes, and parent relationships).

The issue arose because we did not copy the real node's name to the tmp one . If it was changed, it was not accounted for, leading to the following incorrect result:  

**Bugged corresponding side situation:**
```text
|_ A
|   |_ A1
|       |_ B
```
Here, `A1` is moved correctly, but its name remains unchanged. Fortunately, when the sync ends, a new `computeFSop` is triggered, detecting and propagating a new move operation to correct the missing change.  

### Fix  
I added the necessary logic to also fetch the real node name if it has moved. A unit test has also been added:  
**`TestUpdateTree::step8b`**.  

---

### Issue  
In `ConflictResolverWorker`, inside `undoMove`, we were using `UpdateTree::getNodeByPath` with `moveNode::originPath.parent_path()` to get the origin parent node. However, `getNodeByPath` searches within the update tree **after** all move operations. If one of the parents has also moved, it will not be found with `originPath` wich reflect the sync before any move operations.

#### Example  
**Initial situation:**
```text
|_ A
|    |_ A1
|_ 1
|    |_ 2
```
**Intermediate situation (local):**  
(`/1/2` moved to `/A/A1/2`, `/1` renamed to `/1b`)  
```text
|_ A
|    |_ A1
|        |_ 2
|_ 1b
```
**Intermediate situation (remote):**  
(`/A/A1` deleted)  
```text
|_ A
|_ 1
|    |_ 2
```
At this point, we have a `moveParentDelete` case. We try to undo the move of `2` and restore it to its original directory. However, the original path `/1/2` no longer exists in the update tree since `/1` was renamed to `/1b`. This leads to an endless sync restart and failure at the same point.  

### Fix  
I made several changes to the way we store origin location data:  

1. **New `MoveOriginInfos` class:**  
   - The `Node` class now has a `MoveOriginInfos` class that stores all necessary origin location data.
   - This ensures that all required information is always set, preventing missing data issues.
   
2. **Storing parent node ID instead of `parentNodeDbId`:**  
   - Previously, we stored `originPath` and `parentNodeDbId`.  
   - Now, we store `originPath` and `parentNodeId`, since `parentNodeDbId` was used only once and could be replaced easily:  
     ```cpp
     _syncDb->node(tmpNode->originParentNodeDbId(), dbNode, found);
     ```
     is now  
     ```cpp
     _syncDb->node(tmpNode->side(), tmpNode->moveOriginInfos()->parentNodeId(), dbNode, found);
     ```
   
3. **Added security checks:**  
   - A log, assertion, and Sentry event are triggered if a move event is inserted on a node without setting its `MoveOriginInfos`.

4. **Fix in `ConflictResolverWorker`:**  
   - Instead of retrieving the origin parent using its path:
     ```cpp
     const auto originParentNode = updateTree->getNodeByPath(moveNode->moveOriginInfos()->path().parent_path());
     ```
   - We now retrieve it using its ID:
     ```cpp
     const auto originParentNode = updateTree->getNodeById(moveNode->moveOriginInfos()->parentNodeId());
     ```
   - This ensures that we can always undo the move, even if the original parent has moved.

#### Example Fix  
**Final correct situation:**
```text
|_ A
|_ 1b
|    |_ 2
```
A unit test has been added for this scenario:  
**`TestConflictResolverWorker::MoveParentDeleteb`**.  

---  

## Bug 3  

By increasing the code coverage, new tests revealed the following bug:  

If a user locally creates a directory `dir1` and moves a file from `A.txt` to `dir1/A.txt`, while remotely a directory `dir2` is created and `A.txt` is moved to `dir2/A.txt`, the `MoveMove (source)` conflict will be considered a pseudo-conflict.  

As a result, the conflict will not be handled, and the sync will complete with the file being in `dir1` locally and `dir2` remotely. However, in the database, both files will be registered in `dir1`.  

At the next `computeFSop`, a remote move from `dir1/A.txt` to `dir2/A.txt` will be detected and propagated.  

This is not a major issue since everything eventually falls into place, but it creates a discrepancy between reality and the database. Additionally, it may lead to strange, unidentified behaviors that are difficult to debug.  

### Fix  
To resolve this issue, I added a condition in `isPseudoConflict`. If both destination parent nodes are new, we check if they have the same path. If they do, it is effectively a pseudo-conflict. Otherwise, since they belong to two different folders, the `MoveMoveSource` conflict is **not** a pseudo-conflict.


src/libsyncengine/reconciliation/conflict_resolver/conflictresolverworker.cpp
```diff
@@ -410,14 +410,12 @@ ExitCode ConflictResolverWorker::findAllChildNodeIdsFromDb(const std::shared_ptr
 }
 
 ExitCode ConflictResolverWorker::undoMove(const std::shared_ptr<Node> moveNode, SyncOpPtr moveOp) {
-    if (!moveNode->moveOrigin().has_value()) {
-        LOG_SYNCPAL_WARN(_logger, "Failed to retrieve origin parent path");
-        return ExitCode::DataError;
-    }
+    LOG_IF_FAIL(moveNode)
```
ClementKunz, line 413: So you've changed the strategy here and accept a crash in case the pointer is null?

src/libcommonserver/utility/logiffail.h
```diff
@@ -30,13 +30,21 @@
     (__VA_ARGS__)
 
 #define LOG_IF_FAIL(...) LOG_IF_FAIL_OVERLOAD(LOG_IF_FAIL, __VA_ARGS__)
+#define LOG_IF_FAIL_LOGGER(logger, ...) LOG_IF_FAIL_LOG(logger, __VA_ARGS__)
```
ClementKunz, line 33: Instead of creating a new macro, could we use `Log::instance()->getLogger()` everywhere in this file instead of `_logger `

src/libsyncengine/update_detection/update_detector/node.h
```diff
@@ -29,18 +30,39 @@
...
+                MoveOriginInfos(const SyncPath &path, const NodeId &parentNodeId) :
+                    _path(path), _parentNodeId(parentNodeId), _isValid(true) {}
+
+                bool isValid() const;
+                const SyncPath &path() const;
+                const NodeId &parentNodeId() const;
+
+            private:
+                bool _isValid = false;
+                SyncPath _path = ":\0/:\0"; // Invalid path for increased safety
```
ClementKunz, line 47: Is there a reason for this specific pattern?

test/libcommon/utility/testtypes.cpp
```diff
@@ -105,4 +105,62 @@ void TestTypes::testExitInfo() {
     CPPUNIT_ASSERT_EQUAL(ExitCause::Unknown, ei.cause());
     CPPUNIT_ASSERT_EQUAL(600, static_cast<int>(ei));
 }
+
+/* Test that all values of an enum class can be converted to a string
+ * Unfortunatly, this test cannot detect missing conversion for the last value of the enum class.
```
ClementKunz, line 110: Maybe we can start adding a last element in each enum to be sure to count properly the number of item, as suggested here: https://stackoverflow.com/questions/14989274/is-it-possible-to-determine-the-number-of-elements-of-a-c-enum-class.
ex: `enum class Example { A, B, C, D, E, Count };` where `Count` would have the exact same name in every enum

src/libcommonserver/utility/logiffail.h
herve-er: `Log::instance()` is not suitable everywhere, any call `Log::instance()->xxw` from a `dll` will lead to a crash.

test/libsyncengine/syncpal/testoperationprocessor.cpp
```diff
@@ -0,0 +1,219 @@
...
+}
+
+void TestOperationProcessor::testIsPseudoConflict() {
+    class MockOperationProcessor : public OperationProcessor {
+        public:
+            using OperationProcessor::OperationProcessor;
+            void execute() override {}
+            bool isPseudoConflictTest(std::shared_ptr<Node> node, std::shared_ptr<Node> correspondingNode) {
+                return OperationProcessor::isPseudoConflict(node, correspondingNode);
+            }
```
ClementKunz, line 98: What's the point of redefining this method without change? Just to expose it publicly?
Why not making `TestOperationPRocessor` a friend class of `OperationProcessor` as we usually do? I am not saying that friend class is the only way, but we should keep some consistence between our test classes.

test/libsyncengine/syncpal/testoperationprocessor.cpp
```diff
@@ -0,0 +1,219 @@
...
+    CPPUNIT_ASSERT(localNodeFile->setParentNode(newLocalNodeNotSyncedDir)); // Moved to a not synced directory
+    CPPUNIT_ASSERT(remoteNodeFile->setParentNode(newRemoteNodeNotSyncedDir)); // Moved to a not synced directory
+    CPPUNIT_ASSERT(!opProcessor.isPseudoConflictTest(localNodeFile, remoteNodeFile));
+
+
+    // Ensure that all the other combinations of operations do not generate pseudo conflicts
+    const std::array<OperationType, 4> allOp = {
+            OperationType::Edit, OperationType::Delete, OperationType::Move /*, OperationType::Create CreateCreate is handled and
+                                                                               other CreateXXXXX combinations are not possible*/
+    };
```
ClementKunz, line 200: Can we change the names?:
- `op` -> `localOp`
- `op2` -> `remoteOp`

test/libsyncengine/syncpal/testoperationprocessor.cpp
```diff
@@ -0,0 +1,219 @@
...
+    _driveDbId = 1;
+    int driveId = atoi(testVariables.driveId.c_str());
+    Drive drive(_driveDbId, driveId, account.dbId(), std::string(), 0, std::string());
+    ParmsDb::instance()->insertDrive(drive);
+    Sync sync(1, drive.dbId(), localPathStr, testVariables.remotePath);
+    ParmsDb::instance()->insertSync(sync);
+
+    // Setup proxy
+    Parameters parameters;
+    bool found;
```
ClementKunz, line 70: initialization

test/libsyncengine/syncpal/testoperationprocessor.h
```diff
@@ -0,0 +1,46 @@
...
+
+    public:
+        TestOperationProcessor() {}
+        void setUp() override;
+        void tearDown() override;
+
+    private:
+        std::shared_ptr<SyncPal> _syncPal;
+        std::shared_ptr<ParmsDb> _parmsDb;
+        int _driveDbId;
```
ClementKunz, line 40: initialization

test/libsyncengine/syncpal/testsyncpal.cpp
```diff
@@ -306,78 +306,78 @@ void TestSyncPal::testAll() {
...
 
 void TestSyncPal::testConflictQueue() {
-    std::shared_ptr<Node> localrootNode = std::shared_ptr<Node>(
-            new Node(std::nullopt, _syncPal->updateTree(ReplicaSide::Local)->side(), Str(""), NodeType::Directory,
-                     OperationType::None, std::nullopt, 1234567890, 1234567890, 12345, nullptr));
-    std::shared_ptr<Node> remoterootNode = std::shared_ptr<Node>(
-            new Node(std::nullopt, _syncPal->updateTree(ReplicaSide::Remote)->side(), Str(""), NodeType::Directory,
-                     OperationType::None, std::nullopt, 1234567890, 1234567890, 12345, nullptr));
+    std::shared_ptr<Node> localrootNode =
+            std::shared_ptr<Node>(new Node(std::nullopt, ReplicaSide::Local, Str(""), NodeType::Directory, OperationType::None,
```
ClementKunz, line 310: Since you touch that part, can we use `std::make_shared` everywhere instead?

src/libsyncengine/reconciliation/conflict_resolver/conflictresolverworker.cpp
herve-er: Well, until now, if `moveNode` was empty, we were not doing any checks and therefore just crashing.

For `moveOriginInfos`, it is not an optional. I have added a lot of safety checks to ensure it is safe to use on any node with a move operation. If, however, `moveOriginInfos` is not set on `moveNode`, the call to `moveNode->moveOriginInfos().parentNodeId()` will log, assert, and send to Sentry, then return an empty `nodeId`, which is caught on line 421. (See `node.h` for the implementation of the safety checks mentioned above.)

src/libsyncengine/update_detection/update_detector/node.h
herve-er: This is an invalid path on Windows, macOS, and Linux (based on `platformInconsistencyCheckerWorker` -> `forbiddenFilenameChars`).

This should work as well:
```suggestion
                SyncPath _path = ":\0///"; // Invalid path for increased safety
```

It is quite the last safety net. This path should never be getted (if that happens, we will have a sentry, along with assertions and logs).

test/libsyncengine/syncpal/testoperationprocessor.cpp
herve-er: Friends are not inherited, which unfortunately means that making `TestOperationProcessor` a friend of `OperationProcessor` will have no effect on `MockOperationProcessor`.  
One possible workaround is to create a `MockOperationProcessor` pointer, cast it to an `OperationProcessor` pointer, and then I can access the private members of `OperationProcessor`. However, this approach seems overly complicated for little benefit.  
 
(Additionally, making `MockOperationProcessor` a friend of `TestOperationProcessor` would only grant access to `MockOperationProcessor`'s private members, not those of its parent.)

test/libsyncengine/syncpal/testsyncpal.cpp
herve-er: Done here: https://github.com/Infomaniak/desktop-kDrive/pull/595/commits/f613e87d6371028cf814191946d09346081d6835

test/libsyncengine/syncpal/testoperationprocessor.cpp
herve-er: Done here: https://github.com/Infomaniak/desktop-kDrive/pull/595/commits/f613e87d6371028cf814191946d09346081d6835

test/libsyncengine/syncpal/testoperationprocessor.cpp
herve-er: Done here: https://github.com/Infomaniak/desktop-kDrive/pull/595/commits/08efce914944a1cafaad7c701c199226a827c3b4

test/libsyncengine/syncpal/testoperationprocessor.cpp
ClementKunz: Ok, it was not clear to me why you needed to mock `OperationProcessor` here since you only call `isPseudoConflictTest` but `OperationProcessor` is virtual pure.

test/libcommon/utility/testtypes.cpp
ClementKunz: Yes but on CI the solution is build in release mode so your test will still not be working for last values

ClementKunz:
Will testing another issue on your branch, I noticed some errors in log while everything seems to work fine:
```
2025-03-18 14:42:28:841 [F] (0x16d84f000) node.cpp:222 - ENFORCE: "_isValid" in file /Users/clement/Projects/desktop-kDrive/src/libsyncengine/update_detection/update_detector/node.cpp, line 222
2025-03-18 14:42:28:841 [F] (0x16d84f000) node.h:111 - ENFORCE: "_moveOriginInfos.isValid()" in file /Users/clement/Projects/desktop-kDrive/src/libsyncengine/update_detection/update_detector/node.h, line 111
2025-03-18 14:42:28:841 [D] (0x16d84f000) updatetreeworker.cpp:980 - *1* Local(1) update tree: Node 'A' (node ID: '62509881', DB ID: '16', parent ID: 'tmp_6KPnbqGY8j') inserted. Operation Move(2) inserted in change events.
2025-03-18 14:42:28:841 [D] (0x16d967000) updatetreeworker.cpp:832 - *1* Remote(2) update tree: Node 'Common documents' (node ID: '3', DB ID: '2', parent ID: '1') inserted. Without change events.
2025-03-18 14:42:28:841 [F] (0x16d84f000) node.cpp:222 - ENFORCE: "_isValid" in file /Users/clement/Projects/desktop-kDrive/src/libsyncengine/update_detection/update_detector/node.cpp, line 222
2025-03-18 14:42:28:842 [F] (0x16d84f000) node.h:111 - ENFORCE: "_moveOriginInfos.isValid()" in file /Users/clement/Projects/desktop-kDrive/src/libsyncengine/update_detection/update_detector/node.h, line 111
2025-03-18 14:42:28:842 [D] (0x16d967000) updatetreeworker.cpp:832 - *1* Remote(2) update tree: Node 'testC' (node ID: '2724678', DB ID: '5', parent ID: '1') inserted. Without change events.
2025-03-18 14:42:28:842 [D] (0x16d84f000) updatetreeworker.cpp:980 - *1* Local(1) update tree: Node 'fileA.txt' (node ID: '62509901', DB ID: '18', parent ID: 'tmp_6KPnbqGY8j') inserted. Operation Move(2) inserted in change events.
```

src/libcommonserver/utility/logiffail.h
luc-guyot-infomaniak: I have made proposal here https://github.com/Infomaniak/desktop-kDrive/pull/591 to handle this situation.

herve-er: Thank, I will wait for your PR to be merged  and revert my change to LOG_IF_FAIL

test/libcommon/utility/testtypes.cpp
herve-er: Done here: https://github.com/Infomaniak/desktop-kDrive/pull/595/commits/92fde55b3865d73007df64c91f6c2345517aa377

herve-er:
@ClementKunz 
https://github.com/Infomaniak/desktop-kDrive/pull/595#issuecomment-2733320065
Fixed here: https://github.com/Infomaniak/desktop-kDrive/pull/595/commits/bb73131768b6e18d0ed456437e35e1b1b0678a0d

src/libsyncengine/reconciliation/operation_generator/operationgeneratorworker.cpp
```diff
@@ -251,14 +251,11 @@ void OperationGeneratorWorker::generateMoveOperation(std::shared_ptr<Node> curre
...
-                                       << op->id() << L" to be propagated on " << op->targetSide() << L" replica from \""
-                                       << (currentNode->moveOrigin()
-                                                   ? Utility::formatSyncPath(currentNode->moveOrigin().value()).c_str()
-                                                   : L"")
-                                       << L"\" to \"" << Utility::formatSyncPath(currentNode->getPath()).c_str() << L"\" (ID: "
-                                       << Utility::s2ws(currentNode->id() ? currentNode->id().value() : "-1").c_str() << L")");
+            LOGW_SYNCPAL_DEBUG(_logger, L"Move operation "
+                                                << op->id() << L" to be propagated on " << op->targetSide() << L" replica from \""
+                                                << Utility::formatSyncPath(currentNode->moveOriginInfos().path())
+                                                << L"\" to \"" << Utility::formatSyncPath(currentNode->getPath()) << L"\" (ID: "
```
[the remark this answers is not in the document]
herve-er: Done here: https://github.com/Infomaniak/desktop-kDrive/pull/595/commits/0203d0b4a5b622238cded337ae18a2bd6b3cd331

src/libsyncengine/reconciliation/operation_generator/operationgeneratorworker.cpp
```diff
@@ -251,14 +251,11 @@ void OperationGeneratorWorker::generateMoveOperation(std::shared_ptr<Node> curre
...
-            LOGW_SYNCPAL_DEBUG(_logger,
-                               L"Move operation "
-                                       << op->id() << L" to be propagated on " << op->targetSide() << L" replica from \""
-                                       << (currentNode->moveOrigin()
-                                                   ? Utility::formatSyncPath(currentNode->moveOrigin().value()).c_str()
-                                                   : L"")
-                                       << L"\" to \"" << Utility::formatSyncPath(currentNode->getPath()).c_str() << L"\" (ID: "
-                                       << Utility::s2ws(currentNode->id() ? currentNode->id().value() : "-1").c_str() << L")");
+            LOGW_SYNCPAL_DEBUG(_logger, L"Move operation "
+                                                << op->id() << L" to be propagated on " << op->targetSide() << L" replica from \""
```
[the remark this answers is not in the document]
herve-er: Done here: https://github.com/Infomaniak/desktop-kDrive/pull/595/commits/0203d0b4a5b622238cded337ae18a2bd6b3cd331

src/libsyncengine/update_detection/update_detector/node.h
```diff
@@ -86,9 +107,9 @@ class Node {
         inline void setSize(int64_t size) { _size = size; }
         inline void setPreviousId(const std::optional<NodeId> &previousNodeId) { _previousId = previousNodeId; }
         bool setParentNode(const std::shared_ptr<Node> &parentNode);
-        inline void setMoveOrigin(const std::optional<SyncPath> &moveOrigin) { _moveOrigin = moveOrigin; }
-        inline void setMoveOriginParentDbId(const std::optional<DbNodeId> &moveOriginParentDbId) {
-            _moveOriginParentDbId = moveOriginParentDbId;
+        inline void setMoveOriginInfos(const MoveOriginInfos &moveOriginInfos) {
+            LOG_IF_FAIL_LOGGER(Log::instance()->getLogger(), _moveOriginInfos.isValid());
```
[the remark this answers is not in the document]
herve-er: Up until now, we have always put a semicolon at the end of the `LOG` macro. Is there a specific reason to change this now?  
 If yes, we should update all `LOG_` macro calls to maintain consistency. Therefore, it would be best to create a new PR and wait until the PR list is nearly empty, as this change will introduce many modifications and potential conflicts.

test/libsyncengine/reconciliation/conflict_finder/testconflictfinderworker.cpp
```diff
@@ -973,30 +955,28 @@ void TestConflictFinderWorker::testCase516() {
...
-    std::shared_ptr<Node> rNodeM = std::shared_ptr<Node>(new Node(dbNodeIdDirM, _syncPal->updateTree(ReplicaSide::Remote)->side(),
-                                                                  Str("m"), NodeType::Directory, OperationType::None, "rm",
-                                                                  createdAt, lastmodified, size, rNodeQ));
-    std::shared_ptr<Node> rNodeR = std::shared_ptr<Node>(new Node(
-            dbNodeIdDirR, _syncPal->updateTree(ReplicaSide::Remote)->side(), Str("r"), NodeType::Directory, OperationType::None,
-            "rr", createdAt, lastmodified, size, _syncPal->updateTree(ReplicaSide::Remote)->rootNode()));
-    std::shared_ptr<Node> rNodeN = std::shared_ptr<Node>(new Node(dbNodeIdDirN, _syncPal->updateTree(ReplicaSide::Remote)->side(),
-                                                                  Str("n"), NodeType::Directory, OperationType::None, "rn",
-                                                                  createdAt, lastmodified, size, rNodeR));
+    std::shared_ptr<Node> nodeQ = std::shared_ptr<Node>(new Node(dbNodeIdDirQ, ReplicaSide::Local, Str("q"), NodeType::Directory,
```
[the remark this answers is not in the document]
herve-er: Done here: https://github.com/Infomaniak/desktop-kDrive/pull/595/commits/5d7746f6d66c87f6efb13ab8e53ed7eb72c218d6

test/libsyncengine/reconciliation/conflict_finder/testconflictfinderworker.cpp
```diff
@@ -973,30 +955,28 @@ void TestConflictFinderWorker::testCase516() {
...
+                                                                 OperationType::None, "q", createdAt, lastmodified, size,
+                                                                 _syncPal->updateTree(ReplicaSide::Local)->rootNode()));
```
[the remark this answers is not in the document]
herve-er: Done here: https://github.com/Infomaniak/desktop-kDrive/pull/595/commits/5d7746f6d66c87f6efb13ab8e53ed7eb72c218d6

src/libsyncengine/propagation/executor/executorworker.cpp
```diff
@@ -996,19 +994,20 @@ ExitInfo ExecutorWorker::handleMoveOp(SyncOpPtr syncOp, bool &ignored, bool &byp
...
-                              : "",
-                      syncOp->conflict().remoteNode() != nullptr
-                              ? (syncOp->conflict().remoteNode()->id().has_value() ? syncOp->conflict().remoteNode()->id().value()
-                                                                                   : "")
-                              : "",
-                      syncOp->conflict().localNode() != nullptr ? syncOp->conflict().localNode()->type() : NodeType::Unknown,
-                      syncOp->affectedNode()->moveOrigin().has_value() ? syncOp->affectedNode()->moveOrigin().value()
-                                                                       : syncOp->affectedNode()->getPath(),
-                      syncOp->conflict().type());
+            const Error err(
```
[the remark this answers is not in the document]
herve-er: I agree, but this PR is becoming quite heavy. I'll leave that for a future one if you don't mind.

ClementKunz:
> @ClementKunz [#595 (comment)](https://github.com/Infomaniak/desktop-kDrive/pull/595#issuecomment-2733320065) Fixed here: [bb73131](https://github.com/Infomaniak/desktop-kDrive/commit/bb73131768b6e18d0ed456437e35e1b1b0678a0d)

Still something that bug me:
```
bool Node::MoveOriginInfos::isValid() const {
    LOG_IF_FAIL_LOGGER(Log::instance()->getLogger(), _isValid);
    return _isValid;
}
```

Here we will log an error while we were just checking the validity of the `MoveOriginInfos`. In some cases, it might be completely ok to have invalid info and that's why we are checking

src/libcommonserver/utility/logiffail.h
ClementKunz: The PR https://github.com/Infomaniak/desktop-kDrive/pull/591 has been merged

herve-er:
> > @ClementKunz [#595 (comment)](https://github.com/Infomaniak/desktop-kDrive/pull/595#issuecomment-2733320065) Fixed here: [bb73131](https://github.com/Infomaniak/desktop-kDrive/commit/bb73131768b6e18d0ed456437e35e1b1b0678a0d)
> 
> Still something that bug me:
> 
> ```
> bool Node::MoveOriginInfos::isValid() const {
>     LOG_IF_FAIL_LOGGER(Log::instance()->getLogger(), _isValid);
>     return _isValid;
> }
> ```
> 
> Here we will log an error while we were just checking the validity of the `MoveOriginInfos`, not using them. In some cases, it might be completely ok to have invalid info and that's why we are checking

I designed the logic as follows:  

- `Node::moveOrigins()` should only be called on a node that has a move event. Therefore, invoking any method on an invalid `MoveOriginInfos` (including `isValid()`) should never happen.  
- To check if a node has a move event, we should not rely on `Node::moveOrigins()->isValid()`, but rather on `Node::hasChangeEvent(OperationType::Move)`.  
- Then, we can safely use `Node::moveOrigins()`, as the absence of move origins (when OperationType::Move is set) is already handled with a sentry check and an assertion. The `isValid()` check then serves as an additional safeguard.  

From your comment, it seems that exposing `isValid()` publicly might be confusing, could lead to misuse, and doesn’t add significant value (just duplicating an already existing safety net). Here’s a proposed change https://github.com/Infomaniak/desktop-kDrive/pull/595/commits/fdc74f82101e2caa7b8f8f069e12ee787deadb4e, let me know what you think.

ClementKunz:
About your changes in https://github.com/Infomaniak/desktop-kDrive/commit/fdc74f82101e2caa7b8f8f069e12ee787deadb4e: yes that looks way better to me since it will avoid logging unnecessary warning and hide the internal logic to the caller.
Thx 👍

src/libsyncengine/update_detection/update_detector/node.cpp
```diff
@@ -74,6 +83,16 @@ std::shared_ptr<Node> Node::getChildExcept(SyncName name, OperationType except)
...
 
+void Node::setChangeEvents(const OperationType ops) {
+    _changeEvents = ops;
+    LOG_IF_FAIL(Log::instance()->getLogger(), (!hasChangeEvent(OperationType::Move) || _moveOriginInfos.isValid()));
+}
+
+void Node::insertChangeEvent(const OperationType &op) {
+    _changeEvents |= op;
+    LOG_IF_FAIL(Log::instance()->getLogger(), (!hasChangeEvent(OperationType::Move) || _moveOriginInfos.isValid()));
+}
```
ClementKunz, line 94: I'm being nitpicky but this will show a fatal error in log if the caller setup the move origin info after setting the operation type, which is not forbidden. Not a big deal but that might cause log reading more difficult in the future.
Don't you think that the checks in `Node::MoveOriginInfos::path()` and `Node::MoveOriginInfos::parentNodeId()` are enough?

herve-er: It is now forbidden to set `MoveOriginInfos` after inserting the move event 😺  . I checked and updated all the code, so this never happens anywhere.  

The inconvenience of relying only on `Node::MoveOriginInfos::path()` and `Node::MoveOriginInfos::parentNodeId()` is that the error log can be quite far from the actual origin of the issue (where we forgot to set `MoveOriginInfos`) and make the debugging harder.

Currently, it generates no spurious warnings. If, in a future development, we ever forget to set `MoveOriginInfos` or set it after `insertChange`, an assertion will be triggered, making it highly likely that the developer will notice it during testing.  

I haven't found any case so far where we would need to set `MoveOriginInfos` after inserting the move change. But if one day such a case arises, it wouldn't be a big deal to remove this assertion.

src/libsyncengine/update_detection/update_detector/node.h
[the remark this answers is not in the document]
luc-guyot-infomaniak: Second thought: keeping the semi column at the end of the line seems more reasonable to me.

PR #598, opened 2025-03-12
Fix crash in update manager destructor

luc-guyot-infomaniak:


src/server/updater/updatemanager.cpp
```diff
@@ -34,6 +34,8 @@ namespace KDC {
 
 std::shared_ptr<AbstractUpdater> UpdateManager::_updater;
 
+std::shared_ptr<UpdateManager> UpdateManager::_instance;
```
ClementKunz, line 37: As discussed together, I am not a huge fan of having a shared pointer pointing to the instance of a singleton, only the singleton itself should have access to its instance.
The solution that we discussed is:
- Removing the shared pointer on an `AbstractUpdater` from `UpdateManager`
- Make `AbstractUpdater` not abstract anymore (so renaming it `Updater` or something like that)
- The `instance()` method of `Updater` should create and return either a pointer to a `SparkleUpdater`, `WindowsUpdater` or `LinuxUpdater` depending on the OS

luc-guyot-infomaniak: I'll definitely do this in a more ambitious PR (Jira issue created). Can we agree that it is sufficient in the context of the forthcoming release?

ClementKunz (approved):

PR #544, opened 2025-02-19
Kdesktop 1553 Some worker may be stuck in pause

herve-er:

PR #604, opened 2025-03-13
Regression - Step 1 slow down

ClementKunz:
This is a regression introduced in https://github.com/Infomaniak/desktop-kDrive/pull/599 and  https://github.com/Infomaniak/desktop-kDrive/pull/544.

The `_initializing` must be set to `false` as soon as the remote snapshot is created and before sending the first LongPoll request

### DIFF
--- a/src/libsyncengine/update_detection/file_system_observer/remotefilesystemobserverworker.cpp
+++ b/src/libsyncengine/update_detection/file_system_observer/remotefilesystemobserverworker.cpp
@@ -73,12 +73,13 @@ void RemoteFileSystemObserverWorker::execute() {
             }
         }
 
+        if (_initializing) _initializing = false;
+
         exitCode = processEvents();
         if (exitCode != ExitCode::Ok) {
             LOG_SYNCPAL_DEBUG(_logger, "Error in processEvents: code=" << exitCode);
             break;
         }
-        if (_initializing) _initializing = false;
         Utility::msleep(LOOP_EXEC_SLEEP_PERIOD);
     }
     LOG_SYNCPAL_DEBUG(_logger, "Worker stopped: name=" << name().c_str());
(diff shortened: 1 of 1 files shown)


src/libsyncengine/update_detection/file_system_observer/remotefilesystemobserverworker.cpp
```diff
@@ -73,12 +73,13 @@ void RemoteFileSystemObserverWorker::execute() {
             }
         }
 
+        if (_initializing) _initializing = false;
```
luc-guyot-infomaniak, line 76:
```suggestion
        _initializing = false;
```

herve-er (approved):

herve-er (changes requested):
When we lose the network, the RFSO is stopped, but the snapshot is still valid. Therefore, at the next `execute` we will not call `generateInitialSnapshot`. In the current state of the PR, there is no guarantee that the RFSO will be able to initialize (processEvents will fail if the network is still unavailable), but `initialize  `is set.

herve-er:
```cpp
if (stopAsked()) {
            exitCode = ExitCode::Ok;
            tryToInvalidateSnapshot();
            break;
        }
```
should be replace by 
```cpp
if (stopAsked()) {
            exitCode = ExitCode::Ok;
            invalidateSnapshot();
            break;
        }
```

Both in RFSO and LFSO, this change will resolve the aforementioned issue as well as a larger one. Since the introduction of `tryToInvalidateSnapshot`, any event occurring on the local filesystem while the sync is in internal pause is missed and only detected when the user manually pauses and restarts the sync or if a data error triggers a full snapshot regeneration.

This explains a Redmine issue I encountered last week.

herve-er (approved):

src/libsyncengine/update_detection/file_system_observer/remotefilesystemobserverworker.cpp
herve-er: Done here: https://github.com/Infomaniak/desktop-kDrive/pull/604/commits/80969d17fb284109b785527495e0d6a1bb5a6022

ChristopheLarchier (approved):

PR #606, opened 2025-03-14
Change git checkout ref to default in the checkout Github action of the Windows runner

luc-guyot-infomaniak:
This PR aims at fixing the inaccuracies observed in SonarCloud reports. The configuration of the `git checkout` action in our Github workflow seems to be the culprit. 

The Sonar's scanner parameters are documented [here](https://docs.sonarsource.com/sonarqube-cloud/advanced-setup/analysis-parameters/).
The corresponding github workflow variables are documented [here](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/store-information-in-variables).

### DIFF
--- a/.github/workflows/build-and-run-extended-tests.yml
+++ b/.github/workflows/build-and-run-extended-tests.yml
@@ -96,7 +96,7 @@ jobs:
 
       - name: Code coverage preview
         id: html_preview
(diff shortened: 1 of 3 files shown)


luc-guyot-infomaniak:
Inconclusive. New tests will be done on a new PR.

PR #609, opened 2025-03-14
Do not run CI for changes in RN or translations

ClementKunz:


ClementKunz:
not conclusive for now

PR #614, opened 2025-03-17
[KDESKTOP-1605] Turns the AbstractUpdater into a concrete singleton

luc-guyot-infomaniak:


luc-guyot-infomaniak:
Abandoned because of initialization problems with interdependent singletons. The design proposed by this PR is not sound.

PR #615, opened 2025-03-18
Fix rn

ClementKunz:
Linux specific entry was visible in all RNs


infomaniak-build-tools/kDrive-template.html
```diff
@@ -29,7 +29,7 @@ <h2><b>Bug fixes</b></h2>
     <li>Improved the management of access denied errors.</li>
     <li>Improved user experience for sending logs.</li>
     <li>macOS - Improved German translation in installer.</li>
-    <li>linux - Fixed a bug where new changes were not synchronized if a parent folder had been moved.</li>
+    <li>Linux - Fixed a bug where new changes were not synchronized if a parent folder had been moved.</li>
```
luc-guyot-infomaniak, line 32: I updated the script `translate-release-notes.py` so that it handles lower case and upper case in the same way.
So this change is not required anymore.
 
The release notes have already been re-generated.

PR #641, opened 2025-03-26
[KDESKTOP-1577] Always associates `ExitCause::DbEntryNotFound` to `ExitCode::DataError`

luc-guyot-infomaniak:


src/server/appserver.cpp
```diff
@@ -509,10 +509,12 @@ void AppServer::handleClientCrash(bool &quit) {
         LOG_FATAL(_logger, "Client has crashed twice in a short time, exiting");
 
         // Reset client restart date in DB
-        bool found = false;
-        if (!KDC::ParmsDb::instance()->updateAppState(AppStateKey::LastClientSelfRestartDate, 0, found) || !found) {
+        if (bool found = false; !KDC::ParmsDb::instance()->updateAppState(AppStateKey::LastClientSelfRestartDate, 0, found)) {
             addError(Error(errId(), ExitCode::DbError, ExitCause::DbEntryNotFound));
```
ChristopheLarchier, line 513: Must be a ExitCause::DbAccessError

ClementKunz:
Why is this change needed?
Originally, `ExitCode::DataError` was triggered when inconsistency was detected between the snapshots or updates trees and the file system, meaning that those data structures needs to be rebuilt from scratch and the sync restarted.
In the context of `AppState`, the behavior is different. Should we really keep the same `ExitCode`?

src/server/appserver.cpp
```diff
@@ -521,10 +523,14 @@ void AppServer::handleClientCrash(bool &quit) {
         const long timestamp =
                 std::chrono::time_point_cast<std::chrono::seconds>(std::chrono::system_clock::now()).time_since_epoch().count();
         const std::string timestampStr = std::to_string(timestamp);
-        bool found = false;
-        if (!KDC::ParmsDb::instance()->updateAppState(AppStateKey::LastClientSelfRestartDate, timestampStr, found) || !found) {
+
+        if (bool found = false;
+            !KDC::ParmsDb::instance()->updateAppState(AppStateKey::LastClientSelfRestartDate, timestampStr, found)) {
             addError(Error(errId(), ExitCode::DbError, ExitCause::DbEntryNotFound));
```
ChristopheLarchier, line 529: Must be a ExitCause::DbAccessError

src/server/appserver.cpp
```diff
@@ -2881,11 +2887,16 @@ bool AppServer::serverCrashedRecently(int seconds) {
 
     AppStateValue appStateValue = int64_t(0);
     if (bool found = false;
-        !KDC::ParmsDb::instance()->selectAppState(AppStateKey::LastServerSelfRestartDate, appStateValue, found) || !found) {
+        !KDC::ParmsDb::instance()->selectAppState(AppStateKey::LastServerSelfRestartDate, appStateValue, found)) {
         addError(Error(errId(), ExitCode::DbError, ExitCause::DbEntryNotFound));
```
ChristopheLarchier, line 2891: Must be a ExitCause::DbAccessError

src/server/appserver.cpp
```diff
@@ -2904,10 +2915,14 @@ bool AppServer::clientCrashedRecently(int seconds) {
     AppStateValue appStateValue = int64_t(0);
 
     if (bool found = false;
-        !KDC::ParmsDb::instance()->selectAppState(AppStateKey ::LastClientSelfRestartDate, appStateValue, found) || !found) {
+        !KDC::ParmsDb::instance()->selectAppState(AppStateKey ::LastClientSelfRestartDate, appStateValue, found)) {
         addError(Error(errId(), ExitCode::DbError, ExitCause::DbEntryNotFound));
```
ChristopheLarchier, line 2919: Must be a ExitCause::DbAccessError

src/server/appserver.cpp
luc-guyot-infomaniak: Fixed with 7f3d2a46016bd1f9b1200f52e966b2122dfbe5d0.

src/server/appserver.cpp
luc-guyot-infomaniak: Fixed with 7f3d2a46016bd1f9b1200f52e966b2122dfbe5d0.

luc-guyot-infomaniak:
> Why is this change needed? Originally, `ExitCode::DataError` was triggered when inconsistency was detected between the snapshots or updates trees and the file system, meaning that those data structures needs to be rebuilt from scratch and the sync restarted. In the context of `AppState`, the behavior is different. Should we really keep the same `ExitCode`?

Let's discuss offline which error code should be returned when some DB item is not found in the context of `appserver.cpp`. Then I'll document the decision in this PR.

luc-guyot-infomaniak:
> Why is this change needed? Originally, `ExitCode::DataError` was triggered when inconsistency was detected between the snapshots or updates trees and the file system, meaning that those data structures needs to be rebuilt from scratch and the sync restarted. In the context of `AppState`, the behavior is different. Should we really keep the same `ExitCode`?

As discussed offline:
- we keep the _status quo_ for the error code associated to `ExitCause::DbEntryNotFound`, that is, we keep `ErrorCode::DataError` (although `ErrorCode::LogicError` is also a valid alternative; to discuss further in a larger context).
- we remove the useless `addError` calls happening while the error will be removed during the re-start of the application.

ChristopheLarchier (approved):

PR #642, opened 2025-03-26
[KDESKTOP-1538] Normalizes item names in UpdateTree::getNodeByPath only if necessary

luc-guyot-infomaniak:
The proposed changes ensure that `getNodeByPath` 
- do not perform NFC-normalizations of file names if they are not required,
- do not normalize the same file name twice in a given call.

The assumption is that within a given `UpdateTree`, the normalization type of a file name is the same for every node and reflects the normalization of file names on the corresponding file systems.

After analysis, it turns out that NFC-normalization is only required when looking for a node by its path within the method `OperationProcessor::findCorrespondingNodeFromPath`.


src/libcommonserver/utility/utility.cpp
```diff
@@ -480,6 +480,34 @@ bool Utility::checkIfSameNormalization(const SyncPath &a, const SyncPath &b, boo
...
 }
 
+bool Utility::areEqual(const SyncName &lhs, const SyncName &rhs, bool normalizeNames) {
+    bool areEqual = false;
+
+    if (normalizeNames) {
+        if (!Utility::checkIfSameNormalization(lhs, rhs, areEqual)) {
+            LOGW_WARN(logger(), L"Error in Utility::checkIfSameNormalization: " << Utility::formatSyncName(lhs) << L" / "
+                                                                                << Utility::formatSyncName(rhs));
+            return false;
```
ChristopheLarchier, line 490: A normalization fail doesn't mean that the names are not equal.

luc-guyot-infomaniak: Thanks, you are right of course! Addressed with 8e954e678c7e416b842d21f2c4fd3e2632b0ae67.

src/libsyncengine/update_detection/update_detector/updatetree.cpp
```diff
@@ -78,42 +78,31 @@ bool UpdateTree::deleteNode(const NodeId &id) {
...
-                continue;
-            }
 
-            if (isEqual) {
-                tmpChildNode = childNode.second;
+    for (auto nameIt = itemNames.rbegin(); nameIt != itemNames.rend(); ++nameIt) {
+        std::shared_ptr<Node> tmpChildNode = nullptr;
+        for (const auto &[_, childNode]: tmpNode->children()) {
+            bool areEquivalent = false;
+            if (Utility::checkIfEquivalent(*nameIt, childNode->name(), areEquivalent, normalizePaths) && areEquivalent) {
```
herve-er, line 94: My main thought when creating this variable was to store the normalized version of nameIt to avoid renormalizing it in each iteration over its children.

It could be something like this:

```cpp
   for (auto nameIt = itemNames.rbegin(); nameIt != itemNames.rend(); ++nameIt) {
        std::shared_ptr<Node> tmpChildNode = nullptr;
        const auto normalizedName = normalizePaths ? normalize(nameIt) : nameIt;
        for (const auto &[_, childNode]: tmpNode->children()) {
            bool areEquivalent = false;
            const auto normalizedChildName = normalizePaths ? normalize(childNode->name()) : nameIt;
            if (Utility::checkIfEquivalent(normalizedName , normalizedChildName , areEquivalent, false) && areEquivalent) {
                tmpChildNode = childNode;
                break;
            }
        }
    }
```

src/CMakeLists.txt
```diff
@@ -9,8 +9,8 @@ if(NOT MSVC)
       set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fstack-protector --param=ssp-buffer-size=4")
       set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fstack-protector --param=ssp-buffer-size=4")
     else()
-      set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -fstack-protector-strong")
-      set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fstack-protector-strong")
+      set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -O0 -fno-omit-frame-pointer -mno-omit-leaf-frame-pointer -fstack-protector-strong")
```
luc-guyot-infomaniak, line 12: Ooops! To be removed.

src/libsyncengine/update_detection/update_detector/updatetree.cpp
luc-guyot-infomaniak: Excellent point! I cannot really tell why I didn't address this from the start. New commit ongoing.

src/libsyncengine/update_detection/update_detector/updatetree.cpp
```diff
@@ -89,9 +89,41 @@ std::shared_ptr<Node> UpdateTree::getNodeByPath(const SyncPath &path, const bool
 
     for (auto nameIt = itemNames.rbegin(); nameIt != itemNames.rend(); ++nameIt) {
         std::shared_ptr<Node> tmpChildNode = nullptr;
+        SyncName normalizedSyncName;
+        if (!Utility::normalizedSyncName(*nameIt, normalizedSyncName)) continue;
```
herve-er: If we fail to normalize here, we should probably return a `nullptr`.  
By using `continue`, we might skip a step in the path.  

```
|__dir1 (id:1)
|       |__ dir2 (id:2)
|       |         |___ dir3 (id:3)
|       |___dir3 (id:4)
```

If we call `getNodeByNormalizedPath("dir1/dir2/dir3")` and the method fails to normalize `dir2`, it will return `dir3` (id:4) instead of `dir3` (id:3).

ChristopheLarchier (approved):

src/libsyncengine/update_detection/update_detector/updatetree.cpp
```diff
@@ -78,45 +78,75 @@ bool UpdateTree::deleteNode(const NodeId &id) {
...
     }
 
-    // Split path
-    std::vector<SyncName> names;
-    SyncPath pathTmp(path);
-    while (pathTmp != pathTmp.root_path()) {
-        names.push_back(pathTmp.filename().native());
-        pathTmp = pathTmp.parent_path();
+    const std::vector<SyncName> itemNames = Utility::splitPath(path);
+    std::shared_ptr<Node> tmpNode = _rootNode;
```
ClementKunz, line 88: Not really an issue with the algorithm but more on the readability.
At this point, `tmpNode` should never be equal to root node. It is a bit unsettling to initialize it to root node. Since it is a shared pointer, why not just initialize it with the default constructor (so no explicit initialization)?

ClementKunz (approved):

PR #648, opened 2025-03-28
Kdesktop 1493 adapt the application responses to the new asleep and waking up status and related backend errors

ClementKunz:
After a long period of inactivity, the drive will be put into a "sleeping" mode in order to save resources.
This PR aims to support this new maintenance status.


src/gui/adddriveserverfolderswidget.cpp
```diff
@@ -190,22 +192,21 @@ void AddDriveServerFoldersWidget::setLogoColor(const QColor &color) {
...
+void AddDriveServerFoldersWidget::onSubfoldersLoaded(const bool error, const ExitCause exitCause) {
     if (error) {
-        CustomMessageBox msgBox(QMessageBox::Warning, tr("An error occured while loading the list of subfolders."),
-                                QMessageBox::Ok, this);
+        QString errorMsg = tr("An error occurred while loading the list of subfolders.");
+        if (exitCause == ExitCause::ServiceUnavailable) {
+            QString driveLink = QString(APPLICATION_PREVIEW_URL).arg(_driveInfo.driveId()).arg("");
+            errorMsg =
+                    tr(R"(Impossible to load the list of subfolders. Your kDrive might be in maintenance.<br>)"
+                       R"(Please, login to the <a style="%1" href="%2">web version</a> to check your kDrive's status, or contact your administrator.)")
```
luc-guyot-infomaniak, line 202: Just wondering whether the link and the suggestion to log in to the web application should appear in every maintenance-related message.

src/libcommon/info/driveinfo.cpp
```diff
@@ -20,19 +20,16 @@
...
 
-DriveInfo::DriveInfo(int dbId, int accountDbId, const QString &name, const QColor &color) :
-    _dbId(dbId), _accountDbId(accountDbId), _name(name), _color(color) {}
-
-DriveInfo::DriveInfo() : _dbId(0), _accountDbId(0) {}
+DriveInfo::DriveInfo() {}
 
 QDataStream &operator>>(QDataStream &in, DriveInfo &info) {
-    in >> info._dbId >> info._accountDbId >> info._name >> info._color >> info._notifications >> info._maintenance >>
+    in >> info._dbId >> info._id >> info._accountDbId >> info._name >> info._color >> info._notifications >> info._maintenance >>
```
luc-guyot-infomaniak, line 26: Why referring to private data members of `info` and not public accessors? (Should the private data members be made public?)

src/gui/drivepreferenceswidget.cpp
```diff
@@ -1122,7 +1122,7 @@ void DrivePreferencesWidget::onOpenFolder(const QString &filePath) {
     emit openFolder(filePath);
 }
 
-void DrivePreferencesWidget::onSubfoldersLoaded(bool error, bool empty) {
+void DrivePreferencesWidget::onSubfoldersLoaded(bool error, ExitCause exitCause, bool empty) {
```
ChristopheLarchier, line 1125: exitCause seems not used

src/gui/adddriveserverfolderswidget.cpp
ClementKunz: Ideally not but unfortunately if the user try to create a new sync on a sleeping drive, the only error that we get is a 503 without more detailed information.
The `asleep` status is only retrieved from request `https://api.kdrive.infomaniak.com/2/drive/<driveID>` that is called when sync is already set up

src/libcommon/info/driveinfo.cpp
ClementKunz: Changed in 4ba66fad5344a97b4eda607263ced2c321d0f352 and 7b0d5f08f195202855b6f951a293d3c3bde6fffa

src/gui/drivepreferenceswidget.cpp
ClementKunz: Done in e2847a5c8624da06f32ca7281e0a02cd4de5dcb9
Unfortunately, the same signal is connected to several slots that do not use the same parameters. Since we plan to re-write the UI, I kept it like that and simply not use the undesired parameters

luc-guyot-infomaniak (approved):

PR #654, opened 2025-04-01
Update Release notes.

herve-er:


release_notes/kDrive-3.6.10.20250401/kDrive-3.6.10.20250401-macos-en.html
```diff
@@ -28,6 +28,7 @@ <h2><b>Bug fixes</b></h2>
     <li>Improved the management of access denied errors.</li>
     <li>Improved user experience for sending logs.</li>
     <li>Improved German translation in installer.</li>
+    <li>Fixed a bug that could block synchronization when renaming a file with only a case change.</li>
```
ChristopheLarchier, line 31: Fixed a bug that could block synchronization when renaming an item with only a change in the case of its name.

infomaniak-build-tools/kDrive-template.html
```diff
@@ -29,6 +29,7 @@ <h2><b>Bug fixes</b></h2>
     <li>Improved the management of access denied errors.</li>
     <li>Improved user experience for sending logs.</li>
     <li>macOS - Improved German translation in installer.</li>
+    <li>macOS - Fixed a bug that could block synchronization when renaming a file with only a case change.</li>
```
luc-guyot-infomaniak, line 32:
```suggestion
    <li>macOS - Fixed a bug that could block synchronization when renaming a file or a directory while changing only the type case.</li>
```
Hopefully "type case" (https://en.wikipedia.org/wiki/Type_case) will correctly be translated in  the french word "casse".

release_notes/kDrive-3.6.10.20250401/kDrive-3.6.10.20250401-macos-en.html
herve-er: Done here: https://github.com/Infomaniak/desktop-kDrive/pull/654/files/9b10420796eb9cec2189f143c5e5f14b79d06e02..4e63a4898af40548f5ec0684385a488d13ca6ed4

ChristopheLarchier (approved):

PR #664, opened 2025-04-07
Poc/poc do not use macdeployqt

ClementKunz:
Running `macdeployqt` on every build takes a lot of time (~50sec) while it should be useful only when creating the binary folder for the very first time.
Therefore, `macdeployqt` is now run only in release mode by default and must be run manually when generating the binary folder.
The suggested changes allow starting the application in debug mode almost instantly.

CMake install + macdeployqt : ~50sec
CMake install only : ~15sec
None : instantly

So, in conclusion, when rebuilding the binary folder in debug or if a library has changed, we need to : 
- Reload CMake project
- Run `macdeployqt` manually once with the following command : `/Users/clement/Projects/CLion-build-debug/install/kDrive.app -libpath=/usr/local/lib -no-strip -executable=/Users/clement/Projects/CLion-build-debug/install/kDrive.app/Contents/MacOS/kDrive`. This will install all necessary dependencies inside the kDrive.app bundle.
- Run `CMake Install`once to copy mandatory files such as `sync-exclude-osx.lst` inside the app bundle and to update the rpaths


src/gui/CMakeLists.txt
```diff
@@ -405,9 +405,11 @@ if(BUILD_KDRIVE_OSX_BUNDLE AND NOT BUILD_LIBRARIES_ONLY)
     set(ENV_QTDIR $ENV{QTDIR})
     set(ENV_DESTDIR $ENV{DESTDIR})
 
-    install(CODE "
-        message(STATUS \"Deploying (Qt) dependencies for kdrive_client...\")
-        execute_process(COMMAND ${ENV_QTDIR}/bin/macdeployqt ${ENV_DESTDIR}${CMAKE_INSTALL_PREFIX}/${KDRIVE_OSX_BUNDLE} -libpath=/usr/local/lib -no-strip -executable=${ENV_DESTDIR}${CMAKE_INSTALL_PREFIX}/${KDRIVE_OSX_BUNDLE}/Contents/MacOS/${APPLICATION_CLIENT_EXECUTABLE})
-        " COMPONENT RUNTIME)
+    if (NOT CMAKE_BUILD_TYPE STREQUAL "Debug")
```
luc-guyot-infomaniak, line 408: It would be nice to display the command in a message for the`DEBUG` mode, saying something like "this command is required if you did not generate the bin directory yet".
Might be useful to document this change in the `README.md` file and/or in the KB.

ClementKunz: Done in 4eba2c56fe282540a108eb6994ece5b9468cf450

ChristopheLarchier (approved):

luc-guyot-infomaniak (approved):

PR #667, opened 2025-04-07
Optimize `IoHelper::getItemType` and `IoHelper::fileStat` calls in LFSO::explorDir

herve-er:
Currently, IoHelper::getItemType was call twice per item. Idem for IoHelper::fileStat. 
This is now reduce to once each per item (excepted if the parent isn't in the snapshot yet in this case we still have two call to IoHelper::fileStat.

Benchmark:
Before:
Local snapshot generated in: **in progress**s for 147335 items | **in progress**s for 147335 items

After:
Local snapshot generated in: 364.965s for 147335 items | 386.056s for 147335 items

~~Not the biggest improvement of the epic but it is always good to take. ~~

### DIFF
--- a/src/libcommonserver/utility/utility.cpp
+++ b/src/libcommonserver/utility/utility.cpp
@@ -767,27 +767,11 @@ bool Utility::checkIfDirEntryIsManaged(const std::filesystem::recursive_director
                                        IoError &ioError) {
     return checkIfDirEntryIsManaged(*dirIt, isManaged, isLink, ioError);
 }
-
-bool Utility::checkIfDirEntryIsManaged(const DirectoryEntry &dirEntry, bool &isManaged, bool &isLink, IoError &ioError) {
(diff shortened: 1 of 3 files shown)


ClementKunz (approved):

src/libcommonserver/utility/utility.cpp
```diff
@@ -767,12 +767,28 @@ bool Utility::checkIfDirEntryIsManaged(const std::filesystem::recursive_director
...
+        LOGW_WARN(logger(),
+                  L"Ignore " << formatSyncPath(dirEntry.path()) << L" because size > " << CommonUtility::maxPathLength());
+        isManaged = false;
+        return true;
+    }
+
+    return true;
+}
+
+bool Utility::checkIfDirEntryIsManaged(const DirectoryEntry &dirEntry, bool &isManaged, bool &isLink, IoError &ioError) {
```
luc-guyot-infomaniak, line 791: `isLink` should be set to `false` at the beginning of the call.

PR #683, opened 2025-04-10
[KDESKTOP-1129] Removes most recent compiler warnings

luc-guyot-infomaniak:
Addressing the warning about unused function in `src/common/utility_linux.cpp` led me to refactor several os-dependent utility files, removing there some dead code.


src/3rdparty/QProgressIndicator/QProgressIndicator.cpp
```diff
@@ -106,12 +106,12 @@ void QProgressIndicator::paintEvent(QPaintEvent* /*event*/) {
 
     for (int i = 0; i < 12; i++) {
         QColor color = m_color;
-        color.setAlphaF(1.0f - (i / 12.0f));
+        color.setAlphaF(1.0f - (static_cast<float>(i) / 12.0f));
         p.setPen(Qt::NoPen);
         p.setBrush(color);
         p.save();
         p.translate(rect().center());
-        p.rotate(m_angle - i * 30.0f);
```
ClementKunz, line 114: Should we really change the 3rd party libs? Can't we remove them from the Sonar analysis?

luc-guyot-infomaniak: No, we should not change them, thank you! Instead, we should ignore/silence the warnings coming from the 3rd party build.

src/gui/clientgui.cpp
```diff
@@ -638,6 +638,7 @@ void ClientGui::computeTrayOverallStatus(SyncStatus &status, bool &unresolvedCon
                         abortOrPausedSeen++;
                         break;
                     case SyncStatus::Undefined:
+                        break;
```
ClementKunz, line 641: Why? It is not allowed to have thew same logic for several enum items?

src/libsyncengine/update_detection/file_system_observer/snapshot/snapshot.cpp
```diff
@@ -530,7 +530,7 @@ bool Snapshot::checkIntegrityRecursively() const {
 bool Snapshot::checkIntegrityRecursively(const std::shared_ptr<SnapshotItem> &parentItem) const {
     // Check that we do not have the same file twice in the same folder
     std::set<SyncName> names;
-    for (const auto child: parentItem->children()) {
+    for (const auto &child: parentItem->children()) {
```
ClementKunz, line 533: This topic has already been discussed in other PR but it is still not very clear. Should we retrieve reference on a `shared_ptr` or should we always copy the pointer? Using references on `shared_ptr` bypass its purpose since we will not increase the internal reference counter.
Perhaps this should be the topic of a dedicated PR to uniformize the usage of `shared_ptr` everywhere

src/gui/clientgui.cpp
luc-guyot-infomaniak: This suprrises me as well! Reverted with 0d8adc24cbcb1168eae4de9efe2d10a75ad0265a. I'll check now if a warnig shows up on some OS (no warning on MacOSX after revert).

src/libsyncengine/update_detection/file_system_observer/snapshot/snapshot.cpp
luc-guyot-infomaniak: Noted. We need to take a decision and address the problem in uniform way.

ClementKunz (approved):

PR #685, opened 2025-04-11
[KDESKTOP-1588] Removes the obsolete _isDehydrated attribute from the DownloadJob class

luc-guyot-infomaniak:


luc-guyot-infomaniak:
Abandoned: the assumption that `_isHydrated` is used by a single method is false.

PR #689, opened 2025-04-15
Kdesktop 1029 display a dedicated error message for file names finishing with a space

ClementKunz:
Create a dedicated error message for the cases where the app cannot sync a file because its name ends with a space (Windows only).


src/libsyncengine/reconciliation/platform_inconsistency_checker/platforminconsistencycheckerutility.cpp
```diff
@@ -124,6 +119,16 @@ bool PlatformInconsistencyCheckerUtility::isNameOnlySpaces(const SyncName &name)
     return Utility::ltrim(name).empty();
 }
 
+bool PlatformInconsistencyCheckerUtility::nameEndsWithSpace([[maybe_unused]] const SyncName &name) {
```
luc-guyot-infomaniak, line 122: The name of the method does not seem to reflect what it returns.
What about `nameEndWithForbiddenSpace`?

ClementKunz: Much better indeed :) Done in 833866575169f463d62710520b0f3e354f334129

luc-guyot-infomaniak (approved):

PR #722, opened 2025-04-30
Fix permission handling in LFSO::explorDir

herve-er:


src/libsyncengine/update_detection/file_system_observer/localfilesystemobserverworker.cpp
```diff
@@ -607,137 +607,145 @@ ExitInfo LocalFileSystemObserverWorker::exploreDir(const SyncPath &absoluteParen
...
-                dirIt.disableRecursionPending();
-                sendAccessDeniedError(absolutePath);
-                continue;
-            }
-
-            if (!isManaged) {
-                LOGW_SYNCPAL_DEBUG(_logger, L"Directory entry is not managed: " << Utility::formatSyncPath(absolutePath));
-                toExclude = true;
-            }
+            try {
```
ChristopheLarchier, line 611: Would it be possible to put the try/catch only around the call that can throw an exception?

herve-er: Done here: https://github.com/Infomaniak/desktop-kDrive/pull/722/commits/66fd46c18459d6e6df7dff497a3d267cbd3793a9

src/libsyncengine/update_detection/file_system_observer/localfilesystemobserverworker.cpp
```diff
@@ -682,7 +683,8 @@ ExitInfo LocalFileSystemObserverWorker::exploreDir(const SyncPath &absoluteParen
                     LOGW_SYNCPAL_INFO(
                             _logger, L"Item: " << Utility::formatSyncPath(absolutePath) << L" rejected because access is denied");
                     sendAccessDeniedError(absolutePath);
-                    toExclude = true;
+                    dirIt.disableRecursionPending();
+                    continue;
                 }
```
ClementKunz, line 688: Why changing this? This is exactly what is done on line 692 when `toExclude` is `true`.
Or perhaps we should remove the `toExclude` and immediately call `disableRecursionPending()`.

src/libcommonserver/io/iohelper.cpp
```diff
@@ -706,7 +706,15 @@ bool IoHelper::checkIfPathExistsWithSameNodeId(const SyncPath &path, const NodeI
...
 
 bool IoHelper::getFileStat(const SyncPath &path, FileStat *filestat, IoError &ioError) noexcept {
-    return _getFileStat(path, filestat, ioError);
+    try {
+        return _getFileStat(path, filestat, ioError);
+    } catch (std::filesystem::filesystem_error &e) {
+        LOGW_WARN(Log::instance()->getLogger(), L"IoHelper::getFileStat error: " << e.code().value() << L" error="
+                                                                                 << Utility::s2ws(e.what()) << L" "
+                                                                                 << Utility::formatSyncPath(path));
+        ioError = IoError::AccessDenied;
```
ClementKunz, line 715: How can we be sure that this exception has been thrown because of an access denied error?

ClementKunz:
It is not very clear for me how this PR fix the permission handling? Is it because there were some uncatched exceptions? So the app was crashing in case of access denied?

herve-er:
> It is not very clear for me how this PR fix the permission handling? Is it because there were some uncatched exceptions? So the app was crashing in case of access denied?

Exactly -> 
image

src/libsyncengine/update_detection/file_system_observer/localfilesystemobserverworker.cpp
herve-er: I remove the toExclude and immediately call disableRecursionPending(): https://github.com/Infomaniak/desktop-kDrive/pull/722/commits/71bd1ea149300e4d1ee73312f25e38e400c87540

PR #727, opened 2025-05-02
Windows - Disable indexation of eml files

ChristopheLarchier:


src/libsyncengine/jobs/network/API_v2/downloadjob.cpp
```diff
@@ -231,6 +231,14 @@ bool DownloadJob::handleResponse(std::istream &is) {
 
                 _responseHandlingCanceled = fetchCanceled || fetchError || (!fetchFinished);
             } else if (_isHydrated) {
+#ifdef _WIN32
+                if (indexingIsProblematic(_localpath)) {
+                    if (!SetFileAttributesA(_localpath.string().c_str(), FILE_ATTRIBUTE_NOT_CONTENT_INDEXED)) {
```
ClementKunz, line 236: Is it really the application's role to change those kinds of attributes. This might lead to undesired behavior in other apps that are using these files.
Could we, instead, simply ignore those changes if the size of the file did not change (or use content checksum if available)??

ChristopheLarchier: The file date changes so we must at least propagate the date change.
We could implement content checksum verification but it's much more complicated.

ChristopheLarchier: Parameter added to let the user accept this behavior or not.

PR #732, opened 2025-05-05
[KDESKTOP-1199] Consider both the NFC and the NFD normalization of each exclusion pattern

luc-guyot-infomaniak:


luc-guyot-infomaniak:
The proposed changes won't work : the extra normalization operations should be done in `gui/fileexclusiondialog.cpp`. Superseded by https://github.com/Infomaniak/desktop-kDrive/pull/751 and a forthcoming PR.

PR #734, opened 2025-05-06
Crash in OperationSorterWorker

ClementKunz:
There is a crash in `OperationSorterWorker` due to missing condition in the newly added "filter" class.

### DIFF
--- a/src/libsyncengine/propagation/operation_sorter/operationsorterfilter.cpp
+++ b/src/libsyncengine/propagation/operation_sorter/operationsorterfilter.cpp
@@ -61,7 +61,7 @@ void OperationSorterFilter::filterMoveBeforeCreateCandidates(const SyncOpPtr &op
     if (op->affectedNode()->hasChangeEvent(OperationType::Create)) {
         if (const auto &[_, ok] = moveBeforeCreateCandidates.try_emplace(op->affectedNode()->name(), op); !ok) {
             const auto &otherOp = moveBeforeCreateCandidates.at(op->affectedNode()->name());
-            if (op->targetSide() != otherOp->targetSide()) {
+            if (op->targetSide() != otherOp->targetSide() || !otherOp->affectedNode()->hasChangeEvent(OperationType::Move)) {
                 return;
             }
             (void) _fixMoveBeforeCreateCandidates.emplace_back(op, otherOp);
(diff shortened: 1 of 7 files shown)


test/libsyncengine/propagation/operation_sorter/testoperationsorterworker.cpp
```diff
@@ -503,6 +503,39 @@ void TestOperationSorterWorker::testFixMoveBeforeMoveParentChildFlip3() {
...
+
+    // Create A
+    const auto nodeA2 = _testSituationGenerator.createNode(ReplicaSide::Local, NodeType::File, "a2", "");
+    nodeA2->setName(Str("A"));
+    const auto createOp = generateSyncOperation(OperationType::Create, nodeA2);
+
+    (void) _syncPal->syncOps()->pushOp(createOp);
+    (void) _syncPal->syncOps()->pushOp(moveOp);
+
+    // ... but apply all sorter method
```
luc-guyot-infomaniak, line 520:
```suggestion
    // ... but apply all sorter methods
```

src/libsyncengine/propagation/operation_sorter/operationsorterfilter.cpp
```diff
@@ -61,7 +61,7 @@ void OperationSorterFilter::filterMoveBeforeCreateCandidates(const SyncOpPtr &op
     if (op->affectedNode()->hasChangeEvent(OperationType::Create)) {
         if (const auto &[_, ok] = moveBeforeCreateCandidates.try_emplace(op->affectedNode()->name(), op); !ok) {
```
herve-er, line 62: Shouldn't we compare normalized names?

Consider the following structure:

```
|
|____dir
|        |_  fileA_nfc
```

Now suppose we move `dir/fileA_nfc` to `/fileA_nfc`, and then create `dir/fileA_nfd`.

With the current comparison logic, there's no guarantee that the move will be processed before the creation. This can lead to issues when propagating changes to the remote, as the two files may be treated as distinct even though they are visually identical.

This behavior seems relatively easy to reproduce when using an office suite on macOS.

ClementKunz: You are right. Done in bc6df69aeaaf751c23c4f855c42d9e8d0b2e9ad9.
Nodes now keeps their normalized name in cache.

herve-er (approved):

PR #737, opened 2025-05-07
[GUI] Make the button to add a drive much more salient

ChristopheLarchier:


herve-er (commented):
It looks like the button background colors have been inverted:

Before:
drive selection GUI review  before

Now:
drive selection GUI review

herve-er (approved):

PR #754, opened 2025-05-12
[KDESKTOP-1199] Apply exclusion patterns for both NFC and NFD normalization of the pattern string

luc-guyot-infomaniak:


src/gui/fileexclusiondialog.cpp
```diff
@@ -311,123 +315,151 @@ void FileExclusionDialog::onExit() {
...
-                done = true;
-                break;
-            }
-            templateIt++;
+    if (!warningCheckBox) {
+        qCDebug(lcFileExclusionDialog()) << "Null pointer!";
+        return;
+    }
+
+    const QString &template_ = warningCheckBox->property(patternProperty).toString();
```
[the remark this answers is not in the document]
ClementKunz: Ok I have tested, it works, but I do not understand why... the `QVariant::toString()` return a copy of the string. How is it possible to still have a valid reference to this string after this line?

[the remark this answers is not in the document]
luc-guyot-infomaniak: It is safe indeed, as `const ref` extends the lifetime of temporary objects. However, it is likely that  taking a reference`&` is superfluous wrt optimization (most compilers are likely to avoid superfluous copy).

src/gui/fileexclusiondialog.cpp
```diff
@@ -311,123 +315,151 @@ void FileExclusionDialog::onExit() {
...
+    const auto &syncNameTemplate = QStr2SyncName(exclusionTemplate);
+    std::set<QString> result;
+    result.emplace(exclusionTemplate);
+
+    SyncName nfcNormalizedTemplate;
+
+    if (const bool nfcSuccess =
+                CommonUtility::normalizedSyncName(syncNameTemplate, nfcNormalizedTemplate, UnicodeNormalization::NFC);
+        !nfcSuccess) {
+        qCWarning(lcFileExclusionDialog()) << "Failed to NFC-normalize the template " << exclusionTemplate;
```
luc-guyot-infomaniak, line 344:
```suggestion
        qCWarning(lcFileExclusionDialog()) << "Error in normalizedSyncName: Failed to NFC-normalize the template " << exclusionTemplate;
```

src/gui/fileexclusiondialog.cpp
```diff
@@ -311,123 +315,151 @@ void FileExclusionDialog::onExit() {
...
     }
 }
 
+namespace {
+//
+//! Computes and returns the NFC and NFD normalizations of `exclusionTemplate`.
+/*!
+  \param exclusionTemplate is the pattern string the normalizations of which are queried.
+  \return a set of QString containing the NFC and NFD normalizations of exclusionTemplate, if those have been successful.
+  The returned set contains additionally the string exclusionTemplate in any case.
```
ChristopheLarchier, line 332: For this solution to work, it would be necessary to take into account all possible NFC/NFD combinations when there are several accented characters in the template, this seems unrealistic to me.

luc-guyot-infomaniak: In the context of the bug that is described at the top of this PR, one path is the initial path whose file name constituents can contain a mixture of NFC and NFD encoded characters, the other one is the remote path whose constituents are all NFD-normalized. So this solution is just enough to fix the bug because the exclusion patterns will contain the two paths of interest in this case.

ChristopheLarchier: I think the remote path is NFC-normalized.
A local file name contains a mixture of N NFC and NFD encoded characters. You will need to generate 2^N combinations of templates...
For N=1, your solution actually works.

ClementKunz: Since we do not want to retrieve the item from the back in that case, we only need to NFC normalization to be added to the exclusion template list (in addition to the one added by the user). Therefor, we do not care if there is a mixture of NFC and NFD encoded characters, they all be normalized to NFC.

luc-guyot-infomaniak: @ChristopheLarchier Indeed, the remote path names are NFC-encoded, my mistake. But the argument I provided above is still valid.

@ClementKunz It is true that adding the NFC normalization of the user template is sufficient to solve the bug. I thought that adding also the NFD normalization would prevent us from causing other bugs (but not all bugs, as observed by Christophe) et gets us closer to the simple principle "one inserted template rules out all its normalizations".

src/gui/fileexclusiondialog.cpp
```diff
@@ -166,6 +164,61 @@ void FileExclusionDialog::initUI() {
...
+                                                                      UnicodeNormalization::NFC);
+            !nfcSuccess) {
+            qCWarning(lcFileExclusionDialog()) << "Failed to NFC-normalize the template " << templateInfo.templ();
+        } else {
+            templateInfo.setTempl(SyncName2QStr(normalizedName));
+        }
+    }
+}
+
+void filterOutTemplatesWrtNfcNormalization(QList<ExclusionTemplateInfo> &templateList) {
```
ClementKunz, line 181: I am not sure to understand why you need those 3 loops, seems overly complicated.
Isn't it possible to simple check the return value of `emplace` to know if the normalized name has been inserted?
If it has been inserted, update the template with its normalized name
Otherwise, remove the item from the list and get the new iterator

luc-guyot-infomaniak: Indeed, I see clearly now to make it simpler!

luc-guyot-infomaniak: Simplified with https://github.com/Infomaniak/desktop-kDrive/pull/754/commits/e5946023d2702882f96f84a79ad37bda3e6731ce

ClementKunz (approved):

src/gui/fileexclusiondialog.cpp
```diff
@@ -323,117 +350,164 @@ void FileExclusionDialog::onExit() {
...
+    if (!nfdSuccess) {
+        qCDebug(lcFileExclusionDialog()) << "Error in CommonUtility::normalizedSyncName. Failed to NFD-normalize the template '"
+                                         << SyncName2QStr(template_) << "'.";
+    }
+
+    if (!nfcSuccess || !nfdSuccess) {
+        qCWarning(lcFileExclusionDialog())
+                << "File exclusion based on user templates may fail to exclude file names depending on their normalizations.";
+    }
+}
```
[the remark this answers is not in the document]
luc-guyot-infomaniak: The difficulty here is that `CommonUtility::computePathNormalizations` is going to be called from `src/libparms/parmsdb.cpp` is a subsequent PR.
If we had a uniform way to log messages across the application components that consume the `CommonUtility` API, the logging would indeed be better placed inside `CommonUtility`. But it is not the case, so I made the following choice : write log-agnositic logic in `CommonUtility` and implement the logging logic in the caller component.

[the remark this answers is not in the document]
ClementKunz: Sorry to re-open this thread.
What is bogging me here is that we have this portion of code normalizing something that will be normalized anyway again afterward. And if failing, the result is not used to warn the user at all, it is only written in log and will probably never come to our attention.
Since this is call only punctually, we can keep it for now if you think it is necessary. But it could be a good idea to improve it in the future.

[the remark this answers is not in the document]
luc-guyot-infomaniak: As we discussed the matter offline, I clearly understood that detecting the normalization error, and logging, without interrupting the flow of the function `std::unordered_set<QString> computeNormalizations(const QString &templateString)` is surprising. 

The situation is improved I believe with, where the function reacts to the error, that is logs and skips useless computations.

ClementKunz (approved):

PR #760, opened 2025-05-13
Kdesktop 1494 issue when trying to propagate an edit on a file opened in only offfice

herve-er:
depends on https://github.com/Infomaniak/desktop-kDrive/pull/761


src/gui/parametersdialog.cpp
```diff
@@ -756,6 +758,11 @@ QString ParametersDialog::getBackErrorText(const ErrorInfo &errorInfo) const {
         case ExitCause::NotFound: {
             return tr("Impossible to download the file.");
         }
+        case ExitCause::FileLocked: {
+            return tr(
+                    "This item is currently locked by another user online.<br>"
+                    "We will retry uploading your changes later.");
+        }
```
ClementKunz, line 765: Should we tell the user that it might generate a conflict if the other user has made some changes?

herve-er: I tried to keep the message short and straightforward.
If there is a conflict, it will be shown once the remote file is unlocked.
There’s also a chance that there is no conflict (i.e., the Online user is just reading the document or the Excel file).

ClementKunz (approved):

PR #766, opened 2025-05-15
Kdesktop 1603 adapt the number of upload and download jobs

ClementKunz:
The goal of this PR is to reduce the number of parallel upload and download jobs in order to reduce the chances to have connection lost.

New rules are:
- By default, the maximum number of threads in the pool is 10 (so max 10 jobs running in parallel).
- We always keep 1 thread available for high priority jobs (hydration) to be started immediately. So, the number of parallel jobs in a normal sync is actually 9.
- We run max 1 upload session at a time, with max 3 chunk jobs in parallel.
- Download job for big files (> 100Mo) can be started only if more than 50% of the thread pool capacity is available (so max 5 big file download jobs in parallel)

All those numbers are to be discussed and can be adjusted easily.


src/libsyncengine/jobs/jobmanager.cpp
```diff
@@ -166,166 +154,136 @@ void JobManager::run() noexcept {
...
-                }
-                auto jobItem = _queuedJobs.top();
-                const auto &job = jobItem.first;
+        auto availableThreads = availableThreadsInPool();
+        // Always keep 1 thread available for jobs with highest priority
+        while (availableThreads > 1 && !_stop && !_queuedJobs.empty()) {
+            std::shared_ptr<AbstractJob> job;
+            auto priority = Poco::Thread::PRIO_NORMAL;
+            {
+                const std::scoped_lock lock(_mutex);
```
luc-guyot-infomaniak, line 163: It would be good to explain which other functions can modify `_queuedJobs` in a thread that is different from `JobManage::run`'s thread.

If this can happen, every other access to `_queuedJobs` should be protected by a  mutex.

**Suggestion.** List all the methods that can access `_queuedJobs`, `_managedJobs` and `_pendingJobs` concrurrently.

luc-guyot-infomaniak (commented):
A first pass.

src/libsyncengine/jobs/jobmanager.cpp
ClementKunz: I have created a thread safe class to access JobManager's data instead.
What do you think about it?

luc-guyot-infomaniak: It looks good!

src/libsyncengine/jobs/jobmanager.cpp
```diff
@@ -66,97 +49,69 @@ std::shared_ptr<JobManager> JobManager::instance() noexcept {
     return _instance;
 }
 
+void JobManager::startThreadIfNeeded() {
```
ChristopheLarchier, line 52: Rename startMainThreadIfNeeded ?

src/libsyncengine/jobs/jobmanager.cpp
```diff
@@ -66,97 +49,69 @@ std::shared_ptr<JobManager> JobManager::instance() noexcept {
...
 
 void JobManager::decreasePoolCapacity() {
-    if (JobManager::instance()->maxNbThreads() > threadPoolMinCapacity) {
+    if (_maxNbThread > threadPoolMinCapacity) {
         // Decrease pool capacity
-        // TODO: Store the pool capacity in DB?
-        _maxNbThread -= static_cast<int>(std::ceil((JobManager::instance()->maxNbThreads() - threadPoolMinCapacity) / 2.0));
-        Poco::ThreadPool::defaultPool().addCapacity(_maxNbThread - Poco::ThreadPool::defaultPool().capacity());
-        LOG_DEBUG(Log::instance()->getLogger(), "Job Manager capacity set to " << _maxNbThread);
+        setPoolCapacity(std::max(static_cast<int>(std::ceil(_maxNbThread / 2)), threadPoolMinCapacity));
```
ChristopheLarchier, line 104: By doing it like this, the function only does something on the first call it seems to me.

src/libsyncengine/jobs/jobmanager.cpp
ClementKunz: Done in 6d1b0c8042398c5392c2d4bc0ce7a8d82ca46423

src/libsyncengine/jobs/jobmanager.cpp
ClementKunz: Indeed, thanks for pointing this out. Fix in 79a68d259aa9bf8d6750f91c70cfac9d06e3c211

src/libsyncengine/jobs/jobmanagerdata.cpp
```diff
@@ -0,0 +1,106 @@
...
+    const std::scoped_lock lock(_mutex);
+    const auto [_, inserted] = _pendingJobs.try_emplace(jobId, job, priority);
+    return inserted;
+}
+
+void JobManagerData::removeFromPendingJobs(const UniqueId jobId) {
+    (void) _pendingJobs.erase(jobId);
+}
+
+void JobManagerData::erase(const UniqueId jobId) {
```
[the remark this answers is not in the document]
ClementKunz: All public methods in this class might be accessed from a different thread. Should we put a general comment at the beginning of the header file?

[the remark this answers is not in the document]
luc-guyot-infomaniak: Discussed and addressed.

ChristopheLarchier (approved):

ClementKunz:
Quick benchmark to compare the old implementation and the new. The test involves downloading 50 files of 10MB each and uploading 50 files of 10MB each:

| Version | # threads | Average time |
| :------ |:----------- |:----- |
| Old | 10 | 10.456323 |
| Old | 20 | 9.0718235 |
| Old | 50 | 8.9774517 |
| New | 10 | 10.9898273 |
| New | 20 | 9.6082931 |
| New | 50 | 9.407708 |

PR #710, opened 2025-04-22
Kdesktop 1645 add xx hash to conan

R-Gld:
# Conan
We introduced Conan, a package manager, to handle the installation of project dependencies.

- The first dependency managed is xxHash version 0.8.2 ([GitHub repo](https://github.com/Cyan4973/xxHash/tree/v0.8.2)).

PR #768, opened 2025-05-15
Manage log4cplus with conan

R-Gld:
## Add Log4Cplus to Conan

log4cplus provides a CMake option, enabled by default: `LOG4CPLUS_ENABLE_DECORATED_LIBRARY_NAME` that modify the library’s target name (see [relevant lines in CMakeLists.txt](https://github.com/log4cplus/log4cplus/blob/7b90d06f2061f9a50607ad1ee1f989ceaeb2a58c/CMakeLists.txt#L137-L150)):

When building a static library, appends S to the target name

When the `UNICODE` option is enabled, appends U to the target name, and we used `log4cplus::log4cplus`

The Conan recipe does not use these decorated names and always exports a single CMake target:
```cmake
log4cplus::log4cplus
```

In /src/libcommon/CMakeLists.txt we were previously linking against log4cplus::log4cplusU; this change updates it to use the standard target name instea

Depends on #710


infomaniak-build-tools/windows/build-drive.ps1
```diff
@@ -380,6 +379,13 @@ function Prepare-Archive {
             Copy-Item -Path $file".dll" -Destination "$archivePath"
         }
     }
+    $find_dep_script = "$path/infomaniak-build-tools/conan/find_conan_dep.ps1"
+
+    $xxhash_folder = & $find_dep_script -Package "xxhash" -Version "0.8.2" -Ci
+    $log4cplus_folder = & $find_dep_script -Package "log4cplus" -Version "2.1.2" -Ci
```
ClementKunz, line 385: The `-Ci` parameter is always set, even for release build?

R-Gld: You are right, this is an oversight. Here is the commit: https://github.com/Infomaniak/desktop-kDrive/pull/768/commits/2ea8d823203c29bb2dd6f609ec67718268dc0f3f

PR #769, opened 2025-05-15
Kdesktop 1461 implement snapshot copy class

herve-er:
### **Background & Motivation**

Previously, the `SyncPal` class managed four instances of the `Snapshot` class:
- **Two "live" snapshots:** Representing the current, ever-changing state of the local and remote filesystems.
- **Two "copied" snapshots:** Static copies of the live snapshots, taken at the start of each sync operation.

This design introduced ambiguity:
- The term "snapshot" typically implies a static view of data at a specific point in time. Our "live" snapshots, which are constantly modified during sync operations, did not fit this definition.
- Additionally, there was no clear separation of responsibility: `SyncPal` managed both static and live snapshots, even though live snapshots are only relevant to the file system observer workers.

### **What This PR Changes**

- **New Classes for Snapshot Semantics:**
  - Introduced the `LiveSnapshot` class, which explicitly represents a mutable, always-updating snapshot of the filesystem. This class is now used exclusively by the file system observer workers.
  - Introduced the `ConstSnapshot` class, which represents an immutable, static snapshot. This is the version taken at the beginning of each sync, and it remains unchanged until the next sync starts.

- **Responsibility Separation:**
  - Moved live snapshot management out of `SyncPal` and into the relevant file system observer worker classes (`LocalFileSystemObserverWorker`, `RemoteFileSystemObserverWorker`). Now, only these workers hold and update the `LiveSnapshot` instances.
  - `SyncPal` now only holds references to the immutable `ConstSnapshot` copies, which aligns with the true definition of a snapshot.

### **Future Plans**

In a future PR (#776), I plan to remove the snapshot copy objects from `SyncPal` entirely. Instead, `ComputeFSOperationWorker` will request a fresh snapshot copy from `SyncPal` when needed. This will further reduce memory usage, as snapshot copies will only be kept in memory during computation and released immediately after.

---

## **Summary**

- Clarifies the distinction between mutable and immutable snapshots.
- Moves responsibility for live snapshots to the file system observer workers.
- Improves code readability and maintainability by enforcing clearer class responsibilities.
- Provides a foundation for future RAM optimizations and cleaner snapshot management. (Kdesktop 1461 part 2)


src/libsyncengine/syncpal/syncpal.cpp
```diff
@@ -417,9 +411,7 @@ void SyncPal::freeSharedObjects() {
 
     // Check that there is no memory leak
     LOG_IF_FAIL(_localSnapshot.use_count() == 0);
-    LOG_IF_FAIL(_remoteSnapshot.use_count() == 0);
-    LOG_IF_FAIL(_localSnapshotCopy.use_count() == 0);
-    LOG_IF_FAIL(_remoteSnapshotCopy.use_count() == 0);
+    LOG_IF_FAIL(_localSnapshot.use_count() == 0);
```
ClementKunz, line 414: Same line twice?

src/libsyncengine/syncpal/syncpal.cpp
```diff
@@ -1373,23 +1320,23 @@ ExitInfo SyncPal::handleAccessDeniedItem(const SyncPath &relativeLocalPath, std:
...
+    _remoteSnapshot = std::make_shared<ConstSnapshot>(liveSnapshot(ReplicaSide::Remote));
+    liveSnapshot(ReplicaSide::Local).startRead();
+    liveSnapshot(ReplicaSide::Remote).startRead();
 }
 
 void SyncPal::freeSnapshotsCopies() {
-    if (_localSnapshotCopy) {
-        _localSnapshotCopy.reset();
+    if (_localSnapshot) {
+        assert(_localSnapshot.use_count() == 1);
```
ClementKunz, line 1334: Is this check really necessary? We are simply resetting the pointer. Is there really a chance to crash here?

luc-guyot-infomaniak:
>     are no longer needed.

Should this half-sentence be removed or completed?

src/libsyncengine/syncpal/sharedobject.h
```diff
@@ -24,10 +24,10 @@ class SharedObject {
     public:
         [[nodiscard]] bool updated() const { return _updated; }
         void startUpdate() { _updated = true; }
-        void startRead() { _updated = false; }
+        void startRead() const { _updated = false; }
 
     private:
-        bool _updated = false;
+        mutable bool _updated = false;
```
ClementKunz, line 30: Why?

From [](https://en.cppreference.com/w/cpp/language/cv) :
> mutable is used to specify that the member does not affect the externally visible state of the class

Here `_updated` is used to change the external visible state.

src/libsyncengine/syncpal/syncpal.h
```diff
@@ -261,7 +261,11 @@ class SYNCENGINE_EXPORT SyncPal : public std::enable_shared_from_this<SyncPal> {
         void resetSharedObjects();
 
         std::shared_ptr<UpdateTree> updateTree(ReplicaSide side) const;
-        std::shared_ptr<Snapshot> snapshot(ReplicaSide side, bool copy = false) const;
+
+        // Returns a snapshot of the filesystem state at the start of the ongoing sync.
+        // Returns nullptr if no sync is currently in progress.
+        std::shared_ptr<ConstSnapshot> snapshot(ReplicaSide side) const;
+        const LiveSnapshot &liveSnapshot(ReplicaSide side) const;
```
luc-guyot-infomaniak, line 268: The `liveSnaphsot` method deserves a docstring too.

src/libsyncengine/syncpal/syncpal.h
```diff
@@ -286,12 +290,10 @@ class SYNCENGINE_EXPORT SyncPal : public std::enable_shared_from_this<SyncPal> {
         std::shared_ptr<SyncDb> _syncDb{nullptr};
 
         // Shared objects
-        std::shared_ptr<Snapshot> _localSnapshot{nullptr}; // Real time local snapshot
-        std::shared_ptr<Snapshot> _remoteSnapshot{nullptr}; // Real time remote snapshot
-        std::shared_ptr<Snapshot> _localSnapshotCopy{
-                nullptr}; // Copy of the real time local snapshot that is used by synchronization workers
-        std::shared_ptr<Snapshot> _remoteSnapshotCopy{
-                nullptr}; // Copy of the real time remote snapshot that is used by synchronization workers
+        std::shared_ptr<ConstSnapshot> _localSnapshot{nullptr}; // A copy of the real-time local snapshot taken at the start
```
luc-guyot-infomaniak, line 293: Are there exceptions, i.e., workers that use the live snapshots instead?

herve-er:
> are no longer needed.
Should this half-sentence be removed or completed?

I could not find this half sentence? Do you know in wich file/line it is?

src/libsyncengine/syncpal/syncpal.cpp
herve-er: Fixed here: https://github.com/Infomaniak/desktop-kDrive/pull/769/commits/c3d297b3b07da9c0a84946762e5402f053eb5290

src/libsyncengine/syncpal/syncpal.cpp
herve-er: Removed here: https://github.com/Infomaniak/desktop-kDrive/pull/769/commits/85e619b03f4882a0edf0b8e5f8800f50058cf435

src/libsyncengine/update_detection/file_system_observer/computefsoperationworker.cpp
```diff
@@ -126,6 +125,45 @@ void ComputeFSOperationWorker::execute() {
...
+    // Remove deleted nodes from sync_node table & cache
+    NodeSet nodeIdSet;
+    ExitCode exitCode = SyncNodeCache::instance()->syncNodes(syncDbId(), syncNodeType, nodeIdSet);
+    if (exitCode != ExitCode::Ok) {
+        LOG_WARN(Log::instance()->getLogger(), "Error in SyncNodeCache::syncNodes");
+        return exitCode;
+    }
+
+    std::erase_if(nodeIdSet, [this, &syncNodeType](auto &item) {
+        const bool ok = syncNodeType == SyncNodeType::TmpLocalBlacklist ? _syncPal->snapshot(ReplicaSide::Local)->exists(item)
```
luc-guyot-infomaniak, line 138: So, we are certain that `syncNodeType` can take only two values in this method?

luc-guyot-infomaniak (commented):
A first pass

src/libsyncengine/syncpal/sharedobject.h
herve-er: Well, I was hesitant about this, but I couldn't find a better solution.
If you ever come up with one, I'll take it!

The live snapshot is now in the FSO workers. The SyncPal can only access a const reference to the live snapshot, since we shouldn't be able to modify it — only read from it.
However, by passing a const reference, we can no longer call `SharedObject::startRead` if it's not marked as `const`.

Another solution would be to add a method in the FSO workers like `FileSystemObserverWorker::startReadSnapshot()`, and on the SyncPal side: `SyncPal::startReadSnapshot(ReplicaSide side)`.

src/libsyncengine/syncpal/syncpal.cpp
ClementKunz: I was more thinking about the `assert` 😅  This is okay for me but if it is not a critical part of code that can crash, I would simply reset the shared pointer without any checks. The simpler the better

src/libsyncengine/syncpal/sharedobject.h
ClementKunz: Okay I see.
Is it feasible to change the constructor `Snapshot::Snapshot(Snapshot const &other)`to take a `LiveSnapshot` instead as argument and call the `startRead()` method on `other` from there instead? It will be therefore necessary to pass a non-const reference to the `LiveSnapshot` but maybe in this context it is acceptable.

src/libsyncengine/update_detection/file_system_observer/computefsoperationworker.cpp
herve-er: Here is the original code:
```cpp
auto nodeIdIt = nodeIdSet.begin();
while (nodeIdIt != nodeIdSet.end()) {
    const bool ok = syncNodeType == SyncNodeType::TmpLocalBlacklist ? snapshotCopy(ReplicaSide::Local)->exists(*nodeIdIt)
                                                                    : snapshotCopy(ReplicaSide::Remote)->exists(*nodeIdIt);
    if (!ok) {
        nodeIdIt = nodeIdSet.erase(nodeIdIt);
    } else {
        nodeIdIt++;
    }
}
```

It can take any of `SyncNodeType` value:
```cpp
enum class SyncNodeType {
    Undefined = 0,
    BlackList, // Nodes that are excluded from sync
    WhiteList, // Explicitly whitelisted nodes (e.g. folder size above limit but user want to sync anyway). Note: all
               // nodes in none of those lists are implicitly whitelisted
    UndecidedList, // Considered as blacklisted until user action
    TmpRemoteBlacklist, // Blacklisted temporarily
    TmpLocalBlacklist, // Blacklisted temporarily
    EnumEnd
};
```

But only `TmpLocalBlacklist` refers to local items, the others refer only to remote ones.

src/libsyncengine/syncpal/syncpal.h
herve-er: There is no worker using the contents of the `liveSnapshot`, but `SyncPalWorker` checks its state several times in `SyncPalWorker::nextStep` (`liveSnapshot().isValid()` and `liveSnapshot().updated()`).

It is also use in some methods of `SyncPal` has `SyncPal::handleAccessDeniedItem(...)`, `SyncPal::status()`, `SyncPal::checkIfExistsOnServer(...)`, `SyncPal::checkIfCanShareItem(...)` and `SyncPal::copySnapshots()`

src/libsyncengine/syncpal/sharedobject.h
ClementKunz: Forget the above comment, `Snapshot` cannot get a `LiveSnapshot` as argument since `LiveSnapshot` inherits from `Snapshot`...

I do not have a better solution to propose 👍

src/libsyncengine/syncpal/syncpal.h
herve-er: Done here: https://github.com/Infomaniak/desktop-kDrive/pull/769/commits/b63ecf28d4ec0f106ed80e1ba39e8f3bce03d43a

test/libsyncengine/reconciliation/conflict_finder/testconflictfinderworker.cpp
```diff
@@ -81,6 +81,8 @@ void TestConflictFinderWorker::testCreateCreateDifferentEncoding() {
     const auto rNodeAC = _situationGenerator.createNode(ReplicaSide::Remote, NodeType::File, "ac", "a");
     rNodeAC->setName(testhelpers::makeNfdSyncName());
 
+    _syncPal->copySnapshots();
+    _syncPal->copySnapshots();
```
ClementKunz, line 85: Only once should be enough 😄

src/libsyncengine/syncpal/syncpal.cpp
herve-er: Oh, well, it's quite an easy way to spot a "memory leak", it allows us to ensure that no worker kept a copy of the pointer, which would prevent us from releasing the memory when the sync is idle.

luc-guyot-infomaniak:
> > are no longer needed.
> > Should this half-sentence be removed or completed?
> 
> I could not find this half sentence? Do you know in wich file/line it is?

This is in the PR description, just above "Future Plans".

test/libsyncengine/reconciliation/conflict_finder/testconflictfinderworker.cpp
herve-er: Done here: https://github.com/Infomaniak/desktop-kDrive/pull/769/commits/db9dfbdd6060989840003269aa5e403f54844962

src/libsyncengine/update_detection/file_system_observer/snapshot/livesnapshot.cpp
```diff
@@ -0,0 +1,291 @@
...
+        assert(false);
+        return false;
+    }
+
+    // Check if `newItem` already exists with the same path but a different Id
+    if (const auto newParent = findItem(newItem.parentId()); newParent) {
+        for (const auto &child: newParent->children()) {
+            if (child->normalizedName() == newItem.normalizedName() && child->id() != newItem.id()) {
+                LOGW_DEBUG(Log::instance()->getLogger(),
+                           L"Item: " << SyncName2WStr(newItem.name()) << L" (" << Utility::s2ws(newItem.id())
```
luc-guyot-infomaniak, line 70:
```suggestion
                           L"Item: " << Utility::formatSyncName(newItem.name()) << L" (" << Utility::s2ws(newItem.id())
```

src/libsyncengine/update_detection/file_system_observer/snapshot/livesnapshot.cpp
```diff
@@ -0,0 +1,291 @@
...
+                                     << L") already exists in parent: " << Utility::s2ws(newItem.parentId())
+                                     << L" with a different id. Removing it and adding the new one.");
+                auto child2 = child; // removeItem cannot be called on a const ref, we need to make a copy.
+                if (!removeItem(child2)) return false;
+                break; // There should be (at most) only one item with the same name in a folder
```
luc-guyot-infomaniak, line 75:
```suggestion
                break; // There should be at most one item with the same normalized name in a folder.
```

src/libsyncengine/syncpal/syncpal.h
luc-guyot-infomaniak: It would be very nice to document this in header files.

src/libsyncengine/update_detection/file_system_observer/computefsoperationworker.cpp
```diff
@@ -126,6 +125,45 @@ void ComputeFSOperationWorker::execute() {
     setDone(exitCode);
 }
 
+ExitCode ComputeFSOperationWorker::updateSyncNode(SyncNodeType syncNodeType) {
+    // Remove deleted nodes from sync_node table & cache
```
luc-guyot-infomaniak, line 129: This comment could be moved right before the method definition.

src/libsyncengine/update_detection/file_system_observer/snapshot/livesnapshot.cpp
```diff
@@ -0,0 +1,291 @@
...
+        } else {
+            newParent->addChild(item);
+        }
+    }
+    if (parentChanged || !isOrphan(item->id())) {
+        startUpdate();
+    }
+
+    if (ParametersCache::isExtendedLogEnabled()) {
+        LOGW_DEBUG(Log::instance()->getLogger(), L"Item: " << SyncName2WStr(item->name()) << L" (" << Utility::s2ws(item->id())
```
luc-guyot-infomaniak, line 122:
```suggestion
        LOGW_DEBUG(Log::instance()->getLogger(), L"Item: " << Utility::formatSyncName(item->name()) << L" (" << Utility::s2ws(item->id())
```

src/libsyncengine/update_detection/file_system_observer/snapshot/snapshot.h
```diff
@@ -110,12 +81,30 @@ class Snapshot : public SharedObject {
...
 
+class ConstSnapshot : public Snapshot {
+    public:
+        explicit ConstSnapshot(const Snapshot &other) :
+            Snapshot(other) {}
+
+    private:
+        // Prevent any derived class to modify the snapshot content.
+        using Snapshot::_items;
+        using Snapshot::_mutex;
```
luc-guyot-infomaniak, line 107: Which concurrent accesses should we be afraid of?

src/libsyncengine/update_detection/file_system_observer/computefsoperationworker.cpp
herve-er: Done here: https://github.com/Infomaniak/desktop-kDrive/pull/769/commits/de39c5640ff4ee03cdc8bbef29bcc851bd51d661

src/libsyncengine/syncpal/syncpal.h
ChristopheLarchier: After discussion, it's ok for me.

src/libsyncengine/update_detection/file_system_observer/computefsoperationworker.cpp
herve-er: Done here: https://github.com/Infomaniak/desktop-kDrive/pull/769/commits/ed909804417f5206fa20582945ea008677ea0d60

src/libsyncengine/syncpal/syncpal.h
herve-er: @luc-guyot-infomaniak Done here: https://github.com/Infomaniak/desktop-kDrive/pull/769/commits/597e3b9541be614062a80a8cd67af2d7a4c2eca3

src/libsyncengine/update_detection/file_system_observer/snapshot/snapshot.h
herve-er: Here are my thoughts after further reflection.

* The performance trade-off of using a mutex is minimal when there is no concurrent access.
* Moving the mutex solely to the `LiveSnapshot` class would introduce more complexity when adding methods to the `Snapshot` class, as we'd need to implement them in both `Snapshot` and `LiveSnapshot`, thereby losing much of the benefit of abstraction.
* The `Snapshot::path` method also needs to be thread-safe, as it updates a cached path, so we need a mutex there regardless.
* There's no regression here, since we're already using a mutex in both `Snapshot` and its copy.

For these reasons, I plan to keep the current implementation, unless you see a major issue.

luc-guyot-infomaniak: Ok, let's keep it is as it is now.

Nevertheless, it would be interesting to document somewhere
- which threads (workers) can access concurrently the data of `LiveSnapshot` (write and read),
- which threads (workers) can access concurrently the data of `ConstSnapshot` (read only).
This would explain the usage of these two snapshots in the synchronization flow.

herve-er: Done here in response to one of your previous comments: https://github.com/Infomaniak/desktop-kDrive/commit/597e3b9541be614062a80a8cd67af2d7a4c2eca3

herve-er:
FYI: As this PR involves replacing snapshot(...) with liveSnapshot(...) in many tests, it inherits a lot of legacy code duplication, which won't be addressed in this PR as it's out of scope.

PR #783, opened 2025-05-20
Kdesktop 1368 mac os finder status icons shortcut unavailable after an external disc containing a sync folder is mounted

ClementKunz:
The sync was not correctly restarted once an external drive on which there is a synchronization is plugged in.
This PR include a simple refactoring of the methods used to start a sync in `AppServer` allowing to simplify the restart of a single sync.

### DIFF
--- a/src/gui/parametersdialog.cpp
+++ b/src/gui/parametersdialog.cpp
@@ -507,6 +507,12 @@ QString ParametersDialog::getSyncPalErrorText(const QString &fctCode, const Exit
                         "this "
                         "time.<br>"
                         "The application will use system proxy settings instead.");
+            } else if (exitCause == ExitCause::SyncDirDoesntExist) {
+                return tr(
+                        "Unfortunately your old proxy configuration could not be migrated, SOCKS5 proxies are not supported at "
+                        "this "
+                        "time.<br>"
+                        "The application will use system proxy settings instead.");
             } else {
                 return tr("A technical error has occurred (error %1).<br>"
                           "Synchronization has been restarted. Please empty the history and if the error persists, please "
--- a/src/libsyncengine/jobs/network/API_v2/downloadjob.cpp
+++ b/src/libsyncengine/jobs/network/API_v2/downloadjob.cpp
@@ -136,6 +136,20 @@ bool DownloadJob::canRun() {
         return false;
     }
 
+    // Check that parent folder exist
+    if (!IoHelper::checkIfPathExists(_localpath.parent_path(), exists, ioError)) {
+        LOGW_WARN(_logger, L"Error in IoHelper::checkIfPathExists: " << Utility::formatIoError(_localpath, ioError));
+        _exitInfo = {ExitCode::SystemError, ExitCause::FileAccessError};
+        return false;
+    }
+
+    if (!exists) {
+        LOGW_DEBUG(_logger, L"Impossible to download item " << Utility::formatSyncPath(_localpath)
+                                                            << L" because the parent does not exist.");
+        _exitInfo = {ExitCode::DataError, ExitCause::NotFound};
+        return false;
+    }
+
     return true;
 }
 
--- a/src/libsyncengine/propagation/executor/executorworker.cpp
+++ b/src/libsyncengine/propagation/executor/executorworker.cpp
@@ -1445,12 +1445,11 @@ ExitInfo ExecutorWorker::handleFinishedJob(std::shared_ptr<AbstractJob> job, Syn
                                                    << Utility::formatSyncPath(relativeLocalPath) << L" " << exitInfo);
                 return exitInfo;
             }
-        } else if (!handleExecutorError(syncOp, job->exitInfo())) {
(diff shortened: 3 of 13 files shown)


src/gui/parametersdialog.cpp
```diff
@@ -507,6 +507,12 @@ QString ParametersDialog::getSyncPalErrorText(const QString &fctCode, const Exit
                         "this "
                         "time.<br>"
                         "The application will use system proxy settings instead.");
+            } else if (exitCause == ExitCause::SyncDirDoesntExist) {
+                return tr(
```
ChristopheLarchier, line 511: Is it really the good message?
Moreover, this message was not displayed voluntarily because it is a normal case (external disk not mounted)

src/libsyncengine/jobs/network/API_v2/downloadjob.cpp
```diff
@@ -136,6 +136,20 @@ bool DownloadJob::canRun() {
         return false;
     }
 
+    // Check that parent folder exist
```
ChristopheLarchier, line 139: Is it usefull?
The parent folder can exist now and not after the file has been downloaded. In my opinion it is better to deal with the error when copying the file to its location.

src/libsyncengine/propagation/executor/executorworker.cpp
```diff
@@ -2239,6 +2238,21 @@ ExitInfo ExecutorWorker::handleExecutorError(SyncOpPtr syncOp, const ExitInfo &o
 
     LOG_WARN(_logger, "Handling " << opsExitInfo << " in ExecutorWorker::handleExecutorError");
 
+    if (opsExitInfo.cause() == ExitCause::NotFound) {
```
ChristopheLarchier, line 2241: Is it usefull?
handleOpsFileNotFound invalidates the snapshot so the inaccessibility of the sync folder will be detected.

src/libsyncengine/syncpal/syncpalworker.cpp
```diff
@@ -80,7 +82,9 @@ void SyncPalWorker::execute() {
                     stopAndWaitForExitOfWorker(fsoWorkers[index]);
                     bool shouldPause = fsoWorkers[index]->exitCode() == ExitCode::NetworkError ||
                                        (fsoWorkers[index]->exitCode() == ExitCode::BackError &&
-                                        fsoWorkers[index]->exitCause() == ExitCause::ServiceUnavailable);
+                                        fsoWorkers[index]->exitCause() == ExitCause::ServiceUnavailable) ||
+                                       (fsoWorkers[index]->exitCode() == ExitCode::SystemError &&
+                                        fsoWorkers[index]->exitCause() == ExitCause::SyncDirDoesntExist);
```
ChristopheLarchier, line 87: It seems to me that this exitCause is duplicating ExitCause::SyncDirAccesError

src/libsyncengine/syncpal/syncpalworker.cpp
```diff
@@ -267,6 +284,17 @@ void SyncPalWorker::unpause() {
     _syncPal->setRestart(true);
 }
 
+void SyncPalWorker::retryLater() {
```
ChristopheLarchier, line 287: Why is this needed?
_pauseTime is already managed by the pause fct.

src/libsyncengine/syncpal/syncpalworker.cpp
```diff
@@ -267,6 +284,17 @@ void SyncPalWorker::unpause() {
...
+    _retryLater = true;
+    _pauseTime = std::chrono::steady_clock::now();
+}
+
+bool SyncPalWorker::shouldRetry() {
```
ChristopheLarchier, line 292: appserver already systematically tries to restart paused syncs, what is the use of this new function?

src/server/appserver.cpp
```diff
@@ -4269,10 +4259,14 @@ void AppServer::onRestartSyncs() {
 
     for (const auto &[syncId, syncPtr]: _syncPalMap) {
         if (!syncPtr) continue;
-        if ((syncPtr->isPaused() || syncPtr->pauseAsked()) &&
-            syncPtr->pauseTime() + std::chrono::minutes(1) < std::chrono::steady_clock::now()) {
+        if (syncPtr->pauseTime() + std::chrono::minutes(1) > std::chrono::steady_clock::now()) return;
+        if (syncPtr->isPaused() || syncPtr->pauseAsked()) {
             syncPtr->unpause();
         }
+        if (syncPtr->shouldRetry()) {
```
ChristopheLarchier, line 4266: unpause fct doesn't already resume the sync?

src/libsyncengine/syncpal/syncpalworker.cpp
ClementKunz: When only unpausing sync, the `initSyncpal` method was not called and, therefore, all the initializations needed to show the file statuses in the Finder were not done.
So here the sync is not paused, it is stopped, and we retry every minute to start it from scratch.

ClementKunz:
implemented in https://github.com/Infomaniak/desktop-kDrive/pull/799

PR #787, opened 2025-05-21
Auto merge workflow

ClementKunz:
The goal of this PR is to create a workflow that will automatically update the branch of all pull request marked as `auto marged` and approved.
This pull request use an action from my personal repository: https://github.com/ClementKunz/auto-update-branch-action. If this solution proves satisfactory, we could fork it the Infomaniak workspace.


.github/workflows/auto-update.yml
```diff
@@ -0,0 +1,19 @@
...
+    - cron: '*/15 * * * *'
+
+concurrency:
+  group: auto-update
+  cancel-in-progress: true
+
+jobs:
+  update-branch:
+    if: ${{ (github.event.pull_request.auto_merge) && (github.event.pull_request.mergeable) }}
+    runs-on: [ self-hosted, macOS, desktop-kdrive ]
```
herve-er, line 15:
```suggestion
    runs-on: [ macOS ]
```

.github/workflows/auto-update.yml
```diff
@@ -0,0 +1,19 @@
+name: Update branch marked auto merge
+
+on:
+  # Runs every 15 minutes
+  schedule:
+    - cron: '*/15 * * * *'
```
herve-er, line 6:
```suggestion
    - cron: '*/15 7-19 * * 1-5'
```
At every 15th minute past every hour from 7 through 19 on every day-of-week from Monday through Friday. (https://crontab.guru/#*/15_7-19_*_*_1-5)

.github/workflows/auto-update.yml
```diff
@@ -0,0 +1,19 @@
...
+
+concurrency:
+  group: auto-update
+  cancel-in-progress: true
+
+jobs:
+  update-branch:
+    if: ${{ (github.event.pull_request.auto_merge) && (github.event.pull_request.mergeable) }}
```
herve-er, line 14: As it will run from a `cron`, I'm not sure `github.event.pull_request` will be set ?

ClementKunz: Yes, I think you are right.
Maybe it is not a good idea to do it like that. Perhaps it would be better to fully do it with the REST API : https://docs.github.com/en/rest/pulls?apiVersion=2022-11-28.
I put back this PR as draft. Let's see later how (if?) we do it

.github/workflows/auto-update-branch.yml
```diff
@@ -0,0 +1,20 @@
+name: Update branch marked auto merge
+
+on:
+  # Allows manual execution
+  workflow_dispatch:
+  # Runs every 15 minutes
+  schedule:
```
herve-er, line 7: It could run on every merge to develop or a release branch, and on every PR approval.

herve-er (approved):

herve-er (changes requested):
test

herve-er (approved):

.github/workflows/auto-update-branch.yml
ClementKunz: As discussed, workflow will be run every time a PR is closed: 92405e29f310d31486a9448be9686dc0752442af. I have excluded the `release` branches in order to keep full control on them.
I have also changed the action to return immediately after having successfully updated a branch.

herve-er (approved):

PR #795, opened 2025-05-23
Deactivate coverage even in case of test faillure.

herve-er:


.github/workflows/build-and-run-extended-tests.yml
```diff
@@ -80,12 +80,14 @@ jobs:
           & "covselect.exe" "--add" 'exclude folder ../'
           & "covselect.exe" "--add" 'exclude folder 3rdparty/'
           & "covselect.exe" "--add" 'exclude folder gui/'
-          
-          & "cov01.exe" "-0" # coverage off
```
luc-guyot-infomaniak, line 84: We can keep this line. Coverage should be deactivated during the execution of  `covhtml.exe`.

herve-er: Done here: https://github.com/Infomaniak/desktop-kDrive/pull/795/commits/8c8567eb62a2cb0449692381f6cda5eafcfd4119

PR #797, opened 2025-05-23
[KDESKTOP-1717] Defines a Sentry environment for each CI runner

luc-guyot-infomaniak:


test/test.cpp
```diff
@@ -28,8 +28,8 @@ int runTestSuite(const std::string &logFileName) {
     /* initialize random seed: */
     srand(static_cast<unsigned int>(time(NULL)));
 
-    // Disable sentry
-    KDC::sentry::Handler::init(KDC::AppType::None); // Disable sentry
+    // Enable sentry if the env variable KDRIVE_SENTRY_ENVIRONMENT is set
+    KDC::sentry::Handler::init(KDC::AppType::Server);
```
herve-er, line 32: Since we are building in release mode, if the `KDRIVE_SENTRY_ENVIRONMENT` environment variable is not set, events will be sent to the `production` environment by default.
The existence of this environment variable should be checked here, and if it is not set, Sentry should be explicitly deactivated by using `KDC::AppType::None`.

herve-er (approved):

PR #811, opened 2025-05-27
Clear operation sorter filter data on every new sync

ClementKunz:
Operation sorter filter data structures were not cleared on every new sync, which mean that nodes previously inserted remains in the filter even if they are not susceptible to be rearranged anymore.

### DIFF
--- a/src/libsyncengine/propagation/operation_sorter/operationsorterfilter.cpp
+++ b/src/libsyncengine/propagation/operation_sorter/operationsorterfilter.cpp
@@ -22,6 +22,8 @@ OperationSorterFilter::OperationSorterFilter(const std::unordered_map<UniqueId, 
     _ops(ops) {}
 
 void OperationSorterFilter::filterOperations() {
+    clear();
+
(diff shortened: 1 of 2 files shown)


src/libsyncengine/propagation/operation_sorter/operationsorterfilter.cpp
```diff
@@ -44,6 +46,18 @@ void OperationSorterFilter::filterOperations() {
         filterMoveBeforeMoveHierarchyFlipCandidates(op, moveBeforeMoveHierarchyFlipCandidates);
     }
 }
+
+void OperationSorterFilter::clear() {
+    _fixDeleteBeforeMoveCandidates.clear();
```
luc-guyot-infomaniak, line 51: Was it the cause of reported bugs?

ClementKunz: No, my app crashed in debug mode because of an assert. That is how I figured out the missing logic

herve-er (approved):

PR #820, opened 2025-06-02
Linux - Crash in Snapshot class

ChristopheLarchier:
https://infomaniak.kchat.infomaniak.com/infomaniak/pl/01971569-08e4-716b-b5de-2d558ec57c5c

I don't understand why the _items map is empty (which causes the crash) at this point. This PR will fix the crash but not the underlying issue.


src/libsyncengine/update_detection/file_system_observer/localfilesystemobserverworker.cpp
```diff
@@ -269,7 +269,7 @@ void LocalFileSystemObserverWorker::changesDetected(const std::list<std::pair<st
                 // afterward. Typically, editors of the MS suite (xlsx, docx) or Adobe suite (pdf) perform a
                 // Delete-followed-by-Create operation during a single edit.
                 NodeId itemId = _snapshot->itemId(relativePath);
-                if (_snapshot->removeItem(itemId)) {
+                if (!itemId.empty() && _snapshot->removeItem(itemId)) {
```
luc-guyot-infomaniak, line 272: Doesn't `removeItem` return `false` if `itemId.empty()` is `true`?

src/libsyncengine/syncpal/excludelistpropagator.cpp
```diff
@@ -87,7 +87,8 @@ ExitCode ExcludeListPropagator::checkItems() {
             if (bool isWarning = false; ExclusionTemplateCache::instance()->isExcluded(relativePath, isWarning)) {
                 if (isWarning) {
                     NodeId localNodeId = _syncPal->snapshot(ReplicaSide::Local)->itemId(relativePath);
-                    NodeType localNodeType = _syncPal->snapshot(ReplicaSide::Local)->type(localNodeId);
+                    NodeType localNodeType =
+                            localNodeId.empty() ? NodeType::Unknown : _syncPal->snapshot(ReplicaSide::Local)->type(localNodeId);
```
luc-guyot-infomaniak, line 91: Doesn't `type` return `NodeType::Unknown` if `localNodeId.empty()` is `true`?

src/libsyncengine/update_detection/file_system_observer/localfilesystemobserverworker.cpp
ChristopheLarchier: No

luc-guyot-infomaniak: Sorry, you are of course right: `removeItem` returns `true` if `itemId` cannot be found.
So there is no crash here. The idea of this change is to call `tryToInvalidateSnapshot()` when `itemId` is empty because we are in abnormal situation and snapshot invalidation could help?

ChristopheLarchier: Snapshot invalidation calls Snapshot::init() which recreates the snapshot (empties it and adds the root item).

luc-guyot-infomaniak: Thanks, but I know this. my question was: "Is the full reconstruction of the snapshot going to be useful in this case?". I presume that we don't know yet (the root problem has not been identified), but your bet is "yes".

luc-guyot-infomaniak (commented):
Happy to approve once all conflicts have been solved.

PR #822, opened 2025-06-03
Kdesktop 1695 detect migration of sync db

herve-er:


src/libparms/db/sync.cpp
```diff
@@ -24,6 +24,7 @@ Sync::Sync() :
     _dbId(0),
     _driveDbId(0),
     _localPath(std::filesystem::path()),
+    _localNodeId(NodeId()),
```
ClementKunz, line 27: Useless, it is already default constructed like this.
Note: same for `_localPath`, `_targetPath`, `_targetNodeId`and `_listingCursor` that were present before this PR

src/libsyncengine/update_detection/file_system_observer/localfilesystemobserverworker.cpp
```diff
@@ -480,6 +498,31 @@ ExitCode LocalFileSystemObserverWorker::generateInitialSnapshot() {
...
+        }
+
+        if (_syncPal->localNodeId().empty()) {
+            if (ExitInfo exitInfo = _syncPal->setLocalNodeId(rootNodeId); !exitInfo) {
+                LOGW_SYNCPAL_WARN(_logger, L"Error in SyncPal::setLocalNodeId: " << exitInfo);
+                return exitInfo;
+            }
+            isValid = true;
+            return ExitCode::Ok;
+        }
```
ClementKunz, line 516: It should not be the role of `LocalFileSystemObserver` to write the sync root's local node ID in the DB. Could we do this somewhere else? Perhaps when starting the sync (in `SyncPal`)?

PR #823, opened 2025-06-03
KDESKTOP-1751 Add auto recovery from apple assistant migration.

herve-er:
/**
         * @brief Attempts to repair local node IDs in the SyncDb after the sync directory has changed its node ID.
         *
         * This method is used when the sync directory has been moved between two filesystems or recreated (Apple migration assistant), causing its inode
         * to change.
         *
         * - It first loads the SyncDb cache if needed and retrieves all known nodes (in db).
         * - For each nodes, it verifies whether the file still exists, retrieves and save its new node ID.
         * - If all the nodes in db are present under the syncroot with the same path, the node IDs are updated in the SyncDb.
         *   Else the process is aborted and false is returned.
         * - After all updates, the new local root node ID is saved.
         *
         * @note This is a best-effort operation and will return false at the first unrecoverable error (e.g., file deleted/moved,
         *       missing entry in the SyncDb cache, unexpected ioError, etc.).
         *
         * @return true if the SyncDb was successfully updated to reflect the new node IDs,
         *         false otherwise.
         */

depends on #822


src/libsyncengine/syncpal/syncpalworker.cpp
```diff
@@ -505,6 +505,89 @@ void SyncPalWorker::stopAndWaitForExitOfAllWorkers(std::shared_ptr<ISyncWorker>
 
     LOG_SYNCPAL_INFO(_logger, "***** Stop done");
 }
+bool SyncPalWorker::tryToFixDbNodeIdsAfterSyncDirChange() {
```
luc-guyot-infomaniak, line 508: Thinking about unit tests: it might be easier to define this function in `syncdb.cpp` (with a view to call it here) so that it can been tested without any instance of `_syncpal`.

luc-guyot-infomaniak (commented):
It would be nice to have some unit tests.

src/libsyncengine/syncpal/syncpalworker.cpp
herve-er: Done here: https://github.com/Infomaniak/desktop-kDrive/pull/823/commits/cf54fcb4843c41cd0df20e97c3c2d86e67becf80

src/libsyncengine/syncpal/syncpalworker.cpp
```diff
@@ -505,6 +505,89 @@ void SyncPalWorker::stopAndWaitForExitOfAllWorkers(std::shared_ptr<ISyncWorker>
...
+    SyncDbReadOnlyCache &dbCache = _syncPal->syncDb()->cache();
+    LOG_SYNCPAL_INFO(_logger, "Trying to fix SyncDb node IDs after sync dir nodeId change...");
+    if (!dbCache.reloadIfNeeded()) {
+        LOGW_SYNCPAL_WARN(_logger, L"Unable to reload SyncDb cache.");
+        return false;
+    }
+
+    std::unordered_set<NodeIds, NodeIds::HashFunction> remainingNodeIds;
```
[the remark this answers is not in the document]
herve-er: It is the naming we already use elsewhere in the code base when we need to perform an operation on each element of a set. Those are the remaining Node Ids that have not been treated yet
See `ComputeFSOperationWorker` -> `NodeIdsSet &remainingNodesIds` or `NodeIdSet remainingDbIds` for example.

But I agree that it can be confusing, changed here: https://github.com/Infomaniak/desktop-kDrive/pull/823/commits/fdf067ee9debc21eac2bc10da91e5396c568ff64

ClementKunz (approved):

herve-er:
> It would be nice to have some unit tests.

@luc-guyot-infomaniak Done here: https://github.com/Infomaniak/desktop-kDrive/pull/823/commits/3611cd54eb8f503abc848ad07997db83e12931d1

ClementKunz (approved):

ClementKunz (approved):

PR #826, opened 2025-06-04
Set the creation date of an edited file within its upload job

ClementKunz:
The creation date was set to 0 in the upload job when editing a file. It had no real impact on the sync because the date is corrected by the backend, but some unwanted log was written in the log file.
Now, the creation date is read directly on the local file in the upload job (for file edition only, not for file creation)


ClementKunz:
Yes I figured out that there was a mismatch the call in Executor before this fix.

> (Maybe we should remove the default value for nbParallelThread, this way, the build will fail if a mismatch like this happens again.)

Yes, I like this idea

> Moreover, instead of removing the creationDate input, simply providing the parameters in the call might be sufficient?

Yes but I feel like that the constructor used only for edition should not be able to change the creation date

PR #846, opened 2025-06-11
KDESKTOP-1590-GUI-message-about-lack-of-disk-space-is-unclear-to-users

herve-er:


src/gui/parametersdialog.cpp
```diff
@@ -387,7 +387,7 @@ QString ParametersDialog::getSyncPalSystemErrorText(const QString &err, const Ex
 
         case ExitCause::NotEnoughDiskSpace:
             return tr(
-                    "There is not enough space left on your disk.<br>"
```
ClementKunz, line 390: Is it really clearer?
Moreover, it could be wrong in the case the user has several partitions and/or disk

ChristopheLarchier (approved):

luc-guyot-infomaniak (approved):

PR #852, opened 2025-06-12
[KDESKTOP-1707] Returns a meaningful ExitInfo when the Linux FolderWatcher has not enough notify watches

luc-guyot-infomaniak:
On a Linux machine, if the number of `inotify` watches is insufficient, the `FolderWatcher` will not be able to watch folders subject to synchronization. This completely blocks the synchronization. Before the changes, the synchronization loops for ever and the GUI does not display a message which is sufficiently helpful.

These changes
- remove the fallback to a synchronization without watch process when there isn't enough `inotify` watches (the application does not fully support and it is not immediate to make it work, let alone if is truly useful),
- stop the synchronization when there isn't enough `inotify` watches,
- display an error message guiding the user towards a practical resolution.


luc-guyot-infomaniak:
> We are removing the fallback to a degraded mode when the maximum number of inotify watches is reached. Just to confirm, this is intentional, correct?
> 
> If that’s the case, it sounds good to me, as it makes the application's behavior more consistent.

Yes, this is accurate: the fallback is removed (it does not work if the number of `inotify` watches is raised quickly; it seems not obvious to fix it and keep it, nor desirable). I have added a description to this PR that highlights this change.

src/libsyncengine/update_detection/file_system_observer/folderwatcher_linux.h
```diff
@@ -28,20 +28,25 @@ namespace KDC {
...
-        int _fileDescriptor = -1;
+        std::int64_t _fileDescriptor = -1;
 
         bool findSubFolders(const SyncPath &dir, std::list<SyncPath> &fullList);
-        bool inotifyRegisterPath(const SyncPath &path);
-        bool addFolderRecursive(const SyncPath &path);
+        ExitInfo inotifyRegisterPath(const SyncPath &path);
+        ExitInfo addFolderRecursive(const SyncPath &path);
         void removeFoldersBelow(const SyncPath &dirPath);
+        struct AddWatchOutCome {
```
ClementKunz, line 45: Did you mean `outcome` or `out come`? Is it the same?

src/libsyncengine/update_detection/file_system_observer/folderwatcher_linux.cpp
```diff
@@ -180,59 +183,67 @@ bool FolderWatcher_linux::findSubFolders(const SyncPath &dir, std::list<SyncPath
...
-            LOG_ERROR(_logger, "Out of memory or limit number of inotify watches reached!"); // TODO : notify the user?
-            return false;
+    const auto outcome = inotifyAddWatch(path);
+    if (outcome.returnValue == -1) {
+        if (outcome.errorNumber == ENOMEM || outcome.errorNumber == ENOSPC) {
+            // Besides running out of memory, an insufficient number of inotify watches is a common error cause.
+            // It needs to be fixed by the user, see e.g.,
+            // https://stackoverflow.com/questions/47075661/error-user-limit-of-inotify-watches-reached-extreact-build
+            LOG_ERROR(_logger, "Out of memory or limit number of inotify watches reached. The latter can be raised by the user.");
+            return {ExitCode::SystemError, ExitCause::NotEnoughINotifyWatches};
```
ClementKunz, line 211: So the error might be not enough memory but the returned cause might lead us on a wrong track while debugging.
Maybe something like `ExitCause::NotEnoughMemoryOrINotifyWatches`

luc-guyot-infomaniak: I opted for a clearer separation of errors: 5b723f166eec759f5347d7b9b0bb5b8b34b3adc4.

ClementKunz (approved):

PR #856, opened 2025-06-13
Remove Fedora from the supported version in the README

ClementKunz:
Remove Fedora from the supported version in the README

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,7 +33,7 @@ The macOS extension is made in **Objective-C/C++**
 ### Minimum Requirements
 | System | With LiteSync | Without LiteSync | ARM
 |---|---|---|---|
-| Linux | No LiteSync support | KDE / Gnome<br>Ubuntu 20.04, Fedora 39 | KDE / Gnome<br>Ubuntu 20.04, Fedora 39
+| Linux | No LiteSync support | KDE / Gnome<br>Ubuntu 20.04 | KDE / Gnome<br>Ubuntu 20.04
 | macOS | macOS 10.15 | macOS 10.15 | ARM Friendly since 3.3.3<br>no more Rosetta 2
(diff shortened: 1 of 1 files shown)


luc-guyot-infomaniak (approved):

README.md
```diff
@@ -33,7 +33,7 @@ The macOS extension is made in **Objective-C/C++**
 ### Minimum Requirements
 | System | With LiteSync | Without LiteSync | ARM
 |---|---|---|---|
-| Linux | No LiteSync support | KDE / Gnome<br>Ubuntu 20.04, Fedora 39 | KDE / Gnome<br>Ubuntu 20.04, Fedora 39
+| Linux | No LiteSync support | KDE / Gnome<br>Ubuntu 20.04 | KDE / Gnome<br>Ubuntu 20.04
```
ChristopheLarchier, line 36: It would be good to repeat exactly what is in the FAQ:
https://www.infomaniak.com/fr/support/faq/2374/synchroniser-kdrive-entre-differents-appareils

Système | Avec Lite Sync | Sans Lite Sync | ARM
-- | -- | -- | --
Linux* | Pas de Lite Sync | Ubuntu 20.04 | OK
macOS | macOS 10.15 | macOS 10.15 | OK
Windows | Windows 10 1709 | Windows 10 | n/a

ChristopheLarchier (approved):

emanuc:
Why? Does this mean that we Fedora users won't receive support? Or that in the long term you'll support Flatpak, making it distribution-agnostic? Thanks in advance for any clarification.

JorisBodin:
The application is only tested on ubuntu. Support is trained only for ubuntu and will be able to answer technical questions only on that platform.
As a result, we cannot guarantee optimal operation on the many other distributions, nor will support be able to properly answer questions specific to these other distributions.
This is just a clarification, as it's already been like this for several years.
So if the application works well on your fedora, there's no reason why it shouldn't work in the future :)

https://faq.infomaniak.com/617

PR #863, opened 2025-06-17
Kdesktop 1752 error if file has remote move and local edit operations

herve-er:


src/libsyncengine/propagation/operation_sorter/operationsorterworker.cpp
```diff
@@ -326,14 +326,17 @@ bool OperationSorterWorker::hasParentWithHigherIndex(const std::unordered_map<Un
...
+                std::find_if(opList.begin(), opList.end(), [](const SyncOpPtr &op) { return op->type() == OperationType::Move; });
 
-        // Since in case of move op, the node already contains the new name
-        // we always want to execute move operation first
-        moveFirstAfterSecond(editOp, moveOp);
+        for (auto &op: opList) {
+            if (op->type() == OperationType::Edit) {
+                moveFirstAfterSecond(op, *moveOpIt);
+            }
+        }
```
ClementKunz, line 339: So we move every Edit operations, that are affected by a Move operation, after all Move operations?

PR #869, opened 2025-06-18
Creation date should not be updated to the current date when it is 0.

herve-er:
This PR addresses the following issue:

When a local edit occurs, an upload job is created. However, this upload job does **not** send the local file's creation date to the backend. As a result, the backend returns its own `creationDate` once the upload is complete. (We don't want to upload the local creation date on edit as it might not be consistent, especially on linux).

If the creation date returned by the backend differs from the one on the local filesystem, the upload job attempts to update the local file's metadata to match the backend's value. Then, the executor updates the database with this new creation date, i.e., the one provided by the backend.

This behavior leads to issues, particularly on Linux where it’s not possible to set a file's creation date. Therefore, during the next sync cycle, a false-positive (omitted) edit is detected and propagated to the database (because the creation date in db differ from the local one). This doesn’t break functionality but causes unnecessary sync operations.

A similar issue occurs on Windows. Previously, if a file's creation date was `0`, we would set it to the current date. This meant the database would store `0`, but the local file would have the current timestamp — again resulting in a mismatch and a spurious sync edit.

### Fixes:

* **Windows**: We now allow setting the local file’s creation date to `0`, avoiding the mismatch (as it is already done in macOS)
* **Linux**: The upload job now always returns the file system’s creation date instead of the one returned by the backend.


src/libsyncengine/propagation/executor/executorworker.cpp
```diff
@@ -1646,6 +1646,13 @@ ExitInfo ExecutorWorker::propagateChangeToDbAndTree(SyncOpPtr syncOp, std::share
             if (syncOp->type() == OperationType::Create) {
                 return propagateCreateToDbAndTree(syncOp, nodeId, newCreationTime, newModificationTime, node, newSize);
             } else {
+                if (syncOp->targetSide() == ReplicaSide::Local) {
+                    // Upload job will try to update the local file creation time if it differs from the remote one.
```
luc-guyot-infomaniak, line 1650: My understanding is that `syncOp` has type `Edit` and that we land in this `else` clause when an edit operation occurred on the remote side. So, is it rather the `DownloadJob` that tries to update the local creation time?
If so, it tries to do so with the remote creation date, right?

src/libsyncengine/propagation/executor/executorworker.cpp
```diff
@@ -1646,6 +1646,13 @@ ExitInfo ExecutorWorker::propagateChangeToDbAndTree(SyncOpPtr syncOp, std::share
...
+                    // However, it might not succeed (e.g. Linux does not support setting creation time). As the creation time in
+                    // the DB refers to the local file, we do not update it in this case. It will be updated on the next sync if
+                    // needed (omitted local edit operation).
+                    newCreationTime = syncOp->correspondingNode()->createdAt().value_or(0);
```
luc-guyot-infomaniak, line 1654: Provided my understanding is correct (see above comment), I would suggest: `const auto &remoteNode = syncOp->correspondingNode();` followed by `newCreationTime = syncOp->localNode->createdAt().value_or(0)`.

herve-er: The logic could actually be simplified. I’ve done that here: https://github.com/Infomaniak/desktop-kDrive/commit/9bd06b51d09fdfa777d4ef0b2f04784cd855ec29
Let me know if you have any questions.

PR #870, opened 2025-06-18
Manage CPPUnit with Conan

R-Gld:
Depends on #918 for the dependency artifacts.

Depends on #933 for the mock fix implementation.


R-Gld:
Depends on https://github.com/Infomaniak/desktop-kDrive/pull/918 for the dependency artifacts.

PR #878, opened 2025-06-20
Name must be set before looking for parent node

ClementKunz:
This bug was discovered while implementing an integration test on cycle breaking:

- Initial state A/AA/AAA
- Rename A/AA/AAA*
- Create A/AA/AAA
- Move A/AA/AAA* -> A/AA/AAA/AAA*

An error occur in the `UpdateTreeWorker`:

Screenshot 2025-06-20 at 13 27 58

This error occurs because in this call: `const auto exitCode = getOrCreateNodeFromExistingPath(moveOp->destinationPath().parent_path(), parentNode);`, the name of the parent (the newly created node) is the same as the name of the currently evaluated node (the moved node).

The correction consists in renaming the moved node before retrieving its parent.


src/libsyncengine/update_detection/update_detector/updatetreeworker.cpp
```diff
@@ -890,30 +890,31 @@ ExitCode UpdateTreeWorker::createMoveNodes(const NodeType &nodeType) {
...
-
             currentNode->setMoveOriginInfos({moveOp->path(), moveOriginParentId.value()});
             currentNode->insertChangeEvent(OperationType::Move);
             currentNode->setCreatedAt(moveOp->createdAt());
             currentNode->setLastModified(moveOp->lastModified());
             currentNode->setSize(moveOp->size());
+            currentNode->setName(moveOp->destinationPath().filename().native());
             currentNode->setIsTmp(false);
 
+            // create parent node if not exist
```
luc-guyot-infomaniak, line 910:
```suggestion
            // Create the parent node if it does not exist
```

ClementKunz:
> If it is possible to write, in a reasonable amount of time, a unit test for that would fail before this change and pass after it, that would be awesome.

It will be the case in integration tests when this [PR](https://github.com/Infomaniak/desktop-kDrive/pull/887) is done

Issue #885, opened 2025-06-23
kDrive cannot sync when analytics.infomaniak.com is blocked or not responding


nvandamme:
I usually blocked any DNS request towards tracking and analytics.
kDrive handle this very badly : when analytics.infomaniak.com is blocked (return 0.0.0.0 upon dns requests), kDrive pause itself at sync stage 1 without continuing.

Please, remove this limitiation as a WebDav client should *not* block because of this *feature* (and avoid collecting data without telling the user also).

R-Gld:
Hello,

The recent implementation of [Matomo](https://matomo.org/) was done using the Qt module [`qt-piwik-tracker`](https://github.com/Infomaniak/desktop-kDrive/tree/develop/src/3rdparty#:~:text=qt%2Dpiwik%2Dtracker%20%40%20c705cfd).

This module relies on Qt's network request management but does not explicitly define a timeout. Since Qt doesn't set a default timeout either, the request ends up using the operating system's default, which can last several tens of seconds. During that time, the sync process may appear frozen. 

As a temporary workaround, you might consider configuring your DNS to return an `NXDOMAIN` error instead of `0.0.0.0`. This would trigger an immediate failure, preventing any TCP socket from being opened and avoiding the OS-level timeout delay.

In a future release, we may add an environment variable to entirely disable connections to Matomo, and of course, set a timeout for all Matomo-related requests.

Have you been able to confirm that the issue is specifically caused by Matomo being blocked? For example, does syncing work correctly when tracking is enabled, and stall when it’s blocked?

Also, just to clarify: kDrive does **not** use WebDAV anymore since version 3.5.0 (~September 2023).

nvandamme:
Thanks.

Changing to NXDOMAIN can largely impact blocking, especially rogue and bad implemented services, hence responding to any filtered DNS request with 0.0.0.0 (default behavior of many DNS filtering servers such as AdGuard DNS, Pihole, NextDNS...).

Qt sure handle setting and handling networktimeout tough : https://doc.qt.io/archives/qt-5.15/qnetworkrequest.html#setTransferTimeout

To confirm, changing to NXDOMAIN seems to fix the issue and not blocking the worker queue.

R-Gld:
Yes, Qt handles setting a timeout; we will include this in a future release.

You can return NXDOMAIN only for our domain using a rule like this:
```
||analytics.infomaniak.com^$dnsrewrite=NXDOMAIN
``` 
[...]

PR #892, opened 2025-06-23
fix(Matomo): handle timeouts, and auto-disable on failures, rename enums, add logging

R-Gld:
### Purpose

This PR enhances the robustness and maintainability of the Matomo integration.

### Main Changes

- **Timeout handling**  
  Added timeout management for all Matomo requests .
- **Automatic Matomo disabling on repeated failures**  
  If several consecutive failures (timeout, connection refusal, etc.) occur when sending data to Matomo, the Matomo feature is automatically disabled. A clear log entry is generated each time this occurs.

- **Enum renaming**  
  Enums used by Matomo have been renamed for better clarity and code readability.

- **Detailed logging**  
  Comprehensive logs have been added: every key step (sending, timeout, failure, disabling) is now traced to make debugging and telemetry monitoring easier.


src/libcommongui/matomoclient.cpp
```diff
@@ -48,6 +50,39 @@ MatomoClient &MatomoClient::instance(const QString &clientId) {
 MatomoClient::MatomoClient(QCoreApplication *app, const QString &clientId) :
     PiwikTracker(app, QUrl(MATOMO_URL), MATOMO_SITE_ID, clientId) {
     initNameFieldMap();
+    qDebug(lcMatomoClient) << "MatomoClient initialized with URL:" << MATOMO_URL
+                      << "and site ID:" << MATOMO_SITE_ID;
+
+    const auto children = this->findChildren<QNetworkAccessManager*>();
+    if (!children.isEmpty()) {
+        _piwikNAM = children.first();
```
ClementKunz, line 58: Seems overly complicated. Why simply checking if `_piwikNAM` is not `nullptr`?

R-Gld: The `QNetworkAccessManager` pointer defined in the `PiwikTracker` class has `private` visibility (see [`qt-piwik-tracker.h` L45-46](https://github.com/pbek/qt-piwik-tracker/blob/c705cfd34109545e859b9915ddcc23f52f594258/piwiktracker.h#L45-L46)).  
The simplest solution I found to access it from the `MatomoClient` class, in order to apply the `transferTimeout` and connect its `finished` signal, is to use Qt's `findChildren` method.

src/libcommongui/matomoclient.h
```diff
@@ -94,6 +92,13 @@ inline Q_LOGGING_CATEGORY(lcMatomoClient, "gui.matomo", QtInfoMsg)
             std::unordered_map<MatomoNameField, std::pair<QString, QString>> _nameFieldMap;
             void initNameFieldMap();
             void getPathAndAction(MatomoNameField name, QString &path, QString &action) const;
+
+            QNetworkAccessManager* _piwikNAM = nullptr;
+
+            static constexpr int _matomoTimeout = 10000;  // Timeout for Matomo requests in milliseconds
+            int _matomoTimeoutCounter = 0;              // Counter to avoid sending too many Matomo requests in a short time
+            static constexpr int _matomoTimeoutMax = 2; // Maximum number of Matomo requests allowed
```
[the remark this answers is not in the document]
R-Gld: Personally, I don't have a preference. If you prefer it for readability or maintainability reasons, I can move them into the `.cpp` file.

Note that since C++17, in this case, `constexpr` implies `inline` (see [`constexpr` specifier - cppreference](https://en.cppreference.com/w/cpp/language/constexpr.html#:~:text=A%20constexpr%20specifier%20used%20in%20the%20first%20declaration%20of%20a%20function%20or%20static%20data%20member(since%20C%2B%2B17)%20implies%20inline.%20If%20any%20declaration%20of%20a%20function%20or%20function%20template%20has%20a%20constexpr%20specifier%2C%20then%20every%20declaration%20must%20contain%20that%20specifier.)), which means these values have no additional memory cost.

That said, such micro-optimizations should never take precedence over readability or maintainability, especially if placing the definitions in the `.cpp` is a conscious design choice for clarity.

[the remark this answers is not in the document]
ClementKunz: My feeling is that we should remain consistent everywhere in the code.
Defining these values inside the source file ensure that they will never be accessed from anything else than the methods defined in this specific source file.

src/libcommongui/matomoclient.cpp
R-Gld: Scope reduced: https://github.com/Infomaniak/desktop-kDrive/pull/892/commits/ff826a9f60f19148c948efb4aae0c2ca1d5845f5 ✅

R-Gld:
The link to the issue https://github.com/Infomaniak/desktop-kDrive/issues/885 was removed to prevent the issue from being closed when merging this PR, since the problems mentioned are no longer related to the fixes applied in this PR.

ClementKunz (approved):

PR #903, opened 2025-06-25
chore(build-scripts): use build via Podman as the default Linux Amd64 release build

luc-guyot-infomaniak:


ClementKunz:
Should be reviewed? All tasks are not completed?

luc-guyot-infomaniak:
> Should be reviewed? All tasks are not completed?

My hope was that one reviewer could perform the missing test on Linux Amd64, because I cannot.

infomaniak-build-tools/linux/build-release-appimage.sh
```diff
@@ -0,0 +1,223 @@
...
+  if [ ! -f "$conan_toolchain_file" ]; then
+    echo "Conan toolchain file not found: $conan_toolchain_file"
+    exit 1
+  fi
+
+  cd "$build_folder"
+
+  cmake -DCMAKE_PREFIX_PATH="$QT_BASE_DIR" \
+      -DCMAKE_INSTALL_PREFIX=/usr \
+      -DQT_FEATURE_neon=ON \
```
R-Gld, line 82: This CMake variable depends on the architecture: `QT_FEATURE_neon` must be set to `ON` on ARM and `OFF` on AMD.

When trying the script on AMD, we get this error:
```
CMake Error at /opt/qt6.2.3/lib/cmake/Qt6/QtFeature.cmake:1135 (message):
  Feature neon is already defined to be "ON" and should now be set to "OFF"                                                                                                                                
  when importing features from Qt6::Core.                                                                                                                                                                  
Call Stack (most recent call first):                                                                                                                                                                       
  /opt/qt6.2.3/lib/cmake/Qt6Core/Qt6CoreConfig.cmake:118 (qt_make_features_available)                                                                                                                      
  /opt/qt6.2.3/lib/cmake/Qt6/Qt6Config.cmake:177 (find_package)                                                                                                                                            
  src/CMakeLists.txt:1 (find_package) 
```

infomaniak-build-tools/linux/build-release-via-podman.sh
```diff
@@ -18,7 +18,51 @@
...
-set -ex
+program_name="$(basename "$0")"
+
+function display_help {
+  echo "$program_name [-h] [-a architecture]"
+  echo "  Build the Linux release AppImage for desktop-kDrive via Podman."
+  echo "where:"
+  echo "-h  Show this help text."
+  echo "-a <architecture>"
+  echo "  Set the target architecture. Either 'arm64' or 'amd64'."
```
R-Gld, line 29: Wouldn't it be nice to just force the chosen architecture with `uname -m`, since the script wouldn't even work if someone on ARM tried to build the AMD version, and vice versa?

R-Gld (changes requested):
I tried to execute the script on the Intel MacBook Pro, and it resulted in an ld error with undefined references to log4cplus after setting `QT_FEATURE_neon` to `OFF` manually.

```
...
/usr/bin/ld: /root/.conan2/p/b/log4cc7cbf822484d8/p/lib/liblog4cplus.so: undefined reference to `lstat@GLIBC_2.33'
...
```
You can find the full log [here](https://kpaste.infomaniak.com/JqnJr8mFjiHhJUvtd_bti3GV7CbK-pl7#AP69PwyqtUxkeAre7HhxrooDkiZwTup11HArthJ6qvue).

But these errors could come from the Conan cache, I'll check further tomorrow.

infomaniak-build-tools/linux/build-release-appimage.sh
luc-guyot-infomaniak: Thanks for the checks and the solution! Fixed with cae98408a5cb15355a260543dad1cfd1730a640b.

infomaniak-build-tools/linux/build-release-via-podman.sh
luc-guyot-infomaniak: Done with cae98408a5cb15355a260543dad1cfd1730a640b. The architecture argument becomes optional and defaults to the host architecture.

PR #908, opened 2025-06-26
Override auto startup file

ClementKunz:
Override auto startup file on each application start to make sure it points to the correct executable.


src/server/appserver.cpp
```diff
@@ -233,6 +233,12 @@ void AppServer::init() {
     // For access to keyring in order to promt authentication popup
     KeyChainManager::instance()->writeDummyTest();
     KeyChainManager::instance()->clearDummyTest();
+
+    // Update the auto startup file
```
herve-er, line 237: Minor comment:
Could this logic be moved to line 298 with the rest of the autostart logic?

PR #914, opened 2025-06-30
fix(parms-db-migration): upgrade columns before any other DB request

luc-guyot-infomaniak:
This PR makes sure to insert missing columns in the `Parameters` database (as part of the `upgrade` method) before any request involving the modified columns. 

Unit tests are added to prevent regressions.

### DIFF
--- a/src/libparms/db/parmsdb.cpp
+++ b/src/libparms/db/parmsdb.cpp
@@ -1117,14 +1117,17 @@ bool ParmsDb::upgradeColumns() {
         return false;
     }
 
-    LOG_INFO(_logger, "Columns upgrade in << dbType()  << successfully completed.");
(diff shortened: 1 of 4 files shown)


src/libparms/db/parmsdb.cpp
```diff
@@ -1061,22 +1061,7 @@ bool ParmsDb::prepare() {
...
-#ifdef _WIN32
-        if (!replaceShortDbPathsWithLongPaths()) {
-            LOG_WARN(_logger, "Failed to replace short DB paths with long ones.");
-        }
-#endif
-    } else {
-        LOG_INFO(_logger, "Apply generic upgrade fixes to " << dbType() << " DB version " << fromVersion);
-    }
-
+bool ParmsDb::upgradeColumns() {
```
ClementKunz, line 1064: Maybe we can find better name since this functions might also create new tables

PR #925, opened 2025-07-03
fix(executor): Update creation time source in upload session

herve-er:
Changed the source of the `newCreationTime` variable to retrieve the creation time from `uploadSessionJob` instead of `uploadJob`.


src/libsyncengine/propagation/executor/executorworker.cpp
```diff
@@ -1638,7 +1638,7 @@ ExitInfo ExecutorWorker::propagateChangeToDbAndTree(SyncOpPtr syncOp, std::share
                     auto uploadSessionJob(std::dynamic_pointer_cast<DriveUploadSession>(job));
                     if (uploadSessionJob) {
                         nodeId = uploadSessionJob->nodeId();
-                        newCreationTime = uploadJob->creationTime();
+                        newCreationTime = uploadSessionJob->creationTime();
```
ClementKunz, line 1641: To avoid further errors, we should reduce the scope of the variable `uploadJob` by declaring it inside the if statement

PR #926, opened 2025-07-03
fix(executor): update creation time source in upload session

herve-er:
Changed the source of the `newCreationTime` variable to retrieve the creation time from `uploadSessionJob` instead of `uploadJob`.


src/libsyncengine/propagation/executor/executorworker.cpp
```diff
@@ -1632,7 +1632,7 @@ ExitInfo ExecutorWorker::propagateChangeToDbAndTree(SyncOpPtr syncOp, std::share
                     auto uploadSessionJob(std::dynamic_pointer_cast<DriveUploadSession>(job));
                     if (uploadSessionJob) {
                         nodeId = uploadSessionJob->nodeId();
-                        newCreationTime = uploadJob->creationTime();
+                        newCreationTime = uploadSessionJob->creationTime();
```
ClementKunz, line 1635: To avoid further errors, we should reduce the scope of variable `uploadJob` by declaring it inside the if statement

herve-er: I tried it, but what is declared in the `if` statement is also available in the `else` statement

luc-guyot-infomaniak: Is there an integration test that performs an upload session?

ClementKunz: Yes, of course. Sorry for that 😅

PR #961, opened 2025-07-18
Modification date on FAT filesystems

ClementKunz:
Resolution for the modification time is 2s on FAT file systems: https://learn.microsoft.com/en-us/windows/win32/api/minwinbase/ns-minwinbase-filetime
Therefore, on FAT system only, the time difference to trigger an EDIT operation must be strictly higher than 1s now.

Depends on #964


src/libcommonserver/utility/utility.cpp
```diff
@@ -24,6 +24,8 @@
 
 #include "utility/utility_base.h"
 
+#include "utility/utility_base.h"
```
luc-guyot-infomaniak, line 27: Included twice.

src/libsyncengine/update_detection/file_system_observer/computefsoperationworker.cpp
```diff
@@ -307,17 +311,17 @@ ExitCode ComputeFSOperationWorker::inferChangeFromDbNode(const ReplicaSide side,
     }
 
     // Detect EDIT
-    const SyncTime snapshotLastModified = snapshot->lastModified(nodeId);
+    const SyncTime snapshotModificationTime = snapshot->lastModified(nodeId);
+    const bool modifiedTimeDiffIsEnough = abs(snapshotModificationTime - dbModificationTime) > _timeDifferenceThresholdForEdit;
```
luc-guyot-infomaniak, line 315: A description of the issue that is being fixed would be helpful. Here, it looks as if the application has detected an EDIT event on a file but will ignore it if the modification date is too close from the modification date stored in the database for this file.

src/libcommonserver/utility/utility.cpp
ClementKunz: Done in 1c2b886a202b3b104e7e9a7b9fbb762fbe07dee2

src/libsyncengine/update_detection/file_system_observer/computefsoperationworker.cpp
ClementKunz: Yes, it is exactly what is done here.
Description added in 1c2b886a202b3b104e7e9a7b9fbb762fbe07dee2

luc-guyot-infomaniak: Thanks, got it: we indeed ignore a modification on a FAT system when it happens shortly (less than two seconds) after the last synchronization of the file.

PR #963, opened 2025-07-21
Fix run tests.sh

R-Gld:
This pull request updates the `./infomaniak-build-tools/run-tests.sh` script to add support for detecting the operating system and sourcing the appropriate configuration file for macOS or Linux. It also includes error handling for unsupported operating systems.

Operating system detection and configuration:

* Added logic to detect the operating system using `uname -s` and source the corresponding `conanrun.sh` file for macOS or Linux.

You can see the error that was generated [here](https://github.com/Infomaniak/desktop-kDrive/actions/runs/16412289543/job/46369956325?pr=956#step:7:18).

### DIFF
--- a/infomaniak-build-tools/run-test.sh
+++ b/infomaniak-build-tools/run-test.sh
@@ -38,14 +38,6 @@ fi
 dir="$1"
 tester="$2"
 
-conanrun=$(find . -type f -name "conanrun.sh" | head -n 1)
-if [[ -n "$conanrun" ]]; then
-    source "$conanrun"
-else
-    echo -e "${RED}conanrun.sh not found${NC}"
-    exit 1
-fi
-
 echo "${YELLOW}---------- Running $tester ----------${NC}"
 pushd "$dir" 1>/dev/null
 


R-Gld:
> Shouldn't this change be applied to `run-test.sh` as well?
> _@herve-er_ 

The script [`run-test.sh`](https://github.com/Infomaniak/desktop-kDrive/blob/fix-run-tests.sh/infomaniak-build-tools/run-test.sh) does not source `conanrun.sh`. I assumed that `conanrun.sh` was sourced outside the script, but I can add the source command to it.

R-Gld:
> > Shouldn't this change be applied to `run-test.sh` as well?
> > _@herve-er_
> 
> The script [`run-test.sh`](https://github.com/Infomaniak/desktop-kDrive/blob/fix-run-tests.sh/infomaniak-build-tools/run-test.sh) does not source `conanrun.sh`. I assumed that `conanrun.sh` was sourced outside the script, but I can add the source command to it.

The `run-test.sh` script is currently used on the test machines, and dependencies are downloaded via GitHub artifacts. The `conanrun.sh` script is not available and does not need to be transferred, as the dependencies are already available alongside the test executable in the current CI workflow.

herve-er (approved):

PR #984, opened 2025-07-30
feat(winui3): Add localization

herve-er:


ClementKunz:
Side remarks:
- The project is located in "src/gui4/windows/kDrive client/kDrive client". Is it really necessary to have 2 nested folders named "kDrive client"
- Did you had a chance to check compatibility with [Loco](https://localise.biz/)? I do not "resw" format in the ones available for export.
Screenshot 2025-07-30 at 17 03 45

herve-er:
> Side remarks:
> 
> * The project is located in "src/gui4/windows/kDrive client/kDrive client". Is it really necessary to have 2 nested folders named "kDrive client"
> * Did you had a chance to check compatibility with [Loco](https://localise.biz/)? I do not "resw" format in the ones available for export.
> 
> Screenshot 2025-07-30 at 17 03 45

The two nested folders are due to how Visual Studio organizes its structure: the first *kDrive client* folder is the **Solution** folder, while the second one is the **Project** folder.
If we later add unit tests for the GUI, they will go in a new Project. For example, the path could be:
`src/gui4/windows/kDrive client/kDrive client_test`.

Yes, I took the time to check, and it is compatible (and very useful, it will save us a lot of time!).
You need to select **XML**, then choose **resX** (which is the format used in `.resw` files).
I invited you on my test project if you want to take a look: https://localise.biz/herve-er/untitled-project#!l=1

ClementKunz (approved):

PR #998, opened 2025-08-06
fix(sync-node-update): remove user-blacklisted nodes from the sync_node table when they are known to have been deleted remotely

luc-guyot-infomaniak:
The missing updates of user-blacklisted nodes (a.k.a user-excluded nodes) in the `sync_node` table can cause the GUI (specifically, the folder tree item widget) to become unresponsive when those blacklisted nodes are numerous. 

Indeed, the desktop application performs `HTTP GET` requests for every entry of the `sync_node` table in order to display the remote drive arborescence. Thus the nodes which were deleted remotely, but which weren't removed from the database table, will trigger useless requests with a `404` return status. 

The proposed changes ensure that any user-blacklisted folder is removed from `SyncDb`'s `sync_node` table when a `HTTP GET` request has revealed that it has been suppressed from the remote replica.

### DIFF
--- a/test/libsyncengine/test.cpp
+++ b/test/libsyncengine/test.cpp
@@ -46,6 +46,8 @@
 #include "jobs/testjobmanager.h"
 #include "propagation/executor/testfilerescuer.h"
 #include "requests/testexclusiontemplatecache.h"
+#include "requests/testsyncnodecache.h"
+
 #include "update_detection/update_detector/benchupdatetreeworker.h"
 
 #if defined(KD_MACOS)


src/libsyncengine/propagation/executor/executorworker.cpp
```diff
@@ -171,41 +178,34 @@ void ExecutorWorker::execute() {
...
 
-    _syncPal->vfs()->cleanUpStatuses();
+            _syncPal->vfs()->cleanUpStatuses();
 
-    setExitCause(executorExitInfo.cause());
-    LOG_SYNCPAL_DEBUG(_logger, "Worker stopped: name=" << name() << " " << executorExitInfo);
-    setDone(executorExitInfo.code());
+            setExitCause(executorExitInfo.cause());
+            LOG_SYNCPAL_DEBUG(_logger, "Worker stopped: name=" << name() << " " << executorExitInfo);
+            setDone(executorExitInfo.code());
```
ClementKunz, line 206: Indentation and brackets looks odd. Especially this part should not be inside the while loop but should be called once the worker has terminated

src/libsyncengine/requests/syncnodecache.cpp
```diff
@@ -40,43 +40,93 @@ std::shared_ptr<SyncNodeCache> SyncNodeCache::instance() {
...
         LOG_WARN(Log::instance()->getLogger(), "Sync not found in syncNodes map for syncDbId=" << syncDbId);
         return ExitCode::DataError;
     }
 
-    if (_syncNodesMap[syncDbId].find(type) == _syncNodesMap[syncDbId].end()) {
+    return ExitCode::Ok;
+}
+
+
+ExitCode SyncNodeCache::checkIfTypeExists(const int syncDbId, const SyncNodeType type) const {
```
ClementKunz, line 58: Can we change the name? It looks like we provided an invalid enum value.
Something like `checkIfSyncNodeListExists`?

src/libsyncengine/requests/syncnodecache.cpp
```diff
@@ -40,43 +40,93 @@ std::shared_ptr<SyncNodeCache> SyncNodeCache::instance() {
...
+bool SyncNodeCache::contains(const int syncDbId, const SyncNodeType type, const NodeId &nodeId) const noexcept {
+    if (auto exitCode = checkIfSyncExists(syncDbId); exitCode != ExitCode::Ok) return false;
+    if (auto exitCode = checkIfTypeExists(syncDbId, type); exitCode != ExitCode::Ok) return false;
+
+    return _syncNodesMap.at(syncDbId).at(type).contains(nodeId);
+}
+bool SyncNodeCache::contains(const int syncDbId, const NodeId &nodeId) const noexcept {
+    if (auto exitCode = checkIfSyncExists(syncDbId); exitCode != ExitCode::Ok) return false;
+
+    for (auto typeInt = toInt(SyncNodeType::BlackList); typeInt <= toInt(SyncNodeType::TmpLocalBlacklist); ++typeInt) {
```
ClementKunz, line 90:
```suggestion
    for (auto typeInt = toInt(SyncNodeType::BlackList); typeInt < toInt(SyncNodeType::EnumEnd); ++typeInt) {
```

test/libsyncengine/test.cpp
```diff
@@ -55,45 +56,11 @@
...
-// CPPUNIT_TEST_SUITE_REGISTRATION(TestOldSyncDb); // Needs a pre 3.3.4 DB
-CPPUNIT_TEST_SUITE_REGISTRATION(TestExecutorWorker);
-CPPUNIT_TEST_SUITE_REGISTRATION(TestFileRescuer);
-CPPUNIT_TEST_SUITE_REGISTRATION(TestSyncPal);
-CPPUNIT_TEST_SUITE_REGISTRATION(TestSyncPalWorker);
-CPPUNIT_TEST_SUITE_REGISTRATION(TestIntegration);
+CPPUNIT_TEST_SUITE_REGISTRATION(TestSyncDb);
+CPPUNIT_TEST_SUITE_REGISTRATION(TestSyncNodeCache);
 
-// CPPUNIT_TEST_SUITE_REGISTRATION(BenchmarkParallelJobs);
```
ClementKunz, line 96: Perhaps other tests are usefull too 😅

src/libsyncengine/requests/syncnodecache.cpp
luc-guyot-infomaniak: Done with 1997e6ce2b988165f6d2269fbec1a537d98b8ac0.

src/libsyncengine/propagation/executor/executorworker.cpp
ClementKunz: Still strange. All this last step (from line 192) should not be done inside the while loop.

luc-guyot-infomaniak: Sorry! Indeed, I had forgotten to submit my latest changes.

ClementKunz (approved):

PR #1001, opened 2025-08-07
[TRELLO-1025] fix(update-tree-corruption): handle the case of a non-empty newly created folder re-using the ID of a deleted item

luc-guyot-infomaniak:


luc-guyot-infomaniak:
Superseded by https://github.com/Infomaniak/desktop-kDrive/pull/1003 because the fix envisioned with this PR comes with major changes in the architecture of the update tree (multiple nodes with the same ID would be needed).

PR #1003, opened 2025-08-08
fix(compute-fs-operations): postpone the remote creation of newly created local items if they reuse the identifiers of deleted local items

luc-guyot-infomaniak:
The following events were not handled properly by the `desktop-kDrive` application:
- A local item is deleted.
- A non-empty local folder is created and reuses the identifier of the deleted item.

Such a situation leads to a corruption of the local update tree: the local update tree which contains a temporary node for the folder created locally but this node has no event of type `Create`.

The proposed changes address this situation in postponing the propagation creation of the local folder and all its descendants to the next synchronization. To this end, the local identifiers that are reused by the new items created locally are inserted into the ComputeFSOperationsWorker::_localReusedIds" and the `Create` operations detected on them or on any of their descendants are removed from the `SyncPal` operation list.

### DIFF
--- a/src/libsyncengine/update_detection/file_system_observer/computefsoperationworker.cpp
+++ b/src/libsyncengine/update_detection/file_system_observer/computefsoperationworker.cpp
@@ -44,7 +44,8 @@ ComputeFSOperationWorker::ComputeFSOperationWorker(SyncDbReadOnlyCache &testSync
 
 void ComputeFSOperationWorker::postponeCreateOperationsOnReusedIds() {
     for (const auto &localId: _localReusedIds) {
-        deleteLocalDescendantOps(localId);
+        _syncPal->removeLocalOperation(localId, OperationType::Create);
+        deleteLocalDescendantCreateOps(localId);
         SyncPath localPath;
         bool ignore = false;
         (void) _syncPal->snapshot(ReplicaSide::Local)->path(localId, localPath, ignore);
@@ -952,7 +953,7 @@ void ComputeFSOperationWorker::deleteChildOpRecursively(const std::shared_ptr<co
     }
 }
 
-void ComputeFSOperationWorker::deleteLocalDescendantOps(const NodeId &localNodeId) {
+void ComputeFSOperationWorker::deleteLocalDescendantCreateOps(const NodeId &localNodeId) {
     const auto localSnapshot = _syncPal->snapshot(ReplicaSide::Local);
     const auto descendantIds = localSnapshot->getDescendantIds(localNodeId);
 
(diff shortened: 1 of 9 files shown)


src/libsyncengine/update_detection/file_system_observer/computefsoperationworker.cpp
```diff
@@ -42,6 +42,21 @@ ComputeFSOperationWorker::ComputeFSOperationWorker(SyncDbReadOnlyCache &testSync
...
+        bool ignore = false;
+        (void) _syncPal->snapshot(ReplicaSide::Local)->path(localId, localPath, ignore);
+        LOGW_SYNCPAL_DEBUG(_logger, L"Postponing the creation of local item with id='"
+                                            << CommonUtility::s2ws(localId) << L"' and " << Utility::formatSyncPath(localPath)
+                                            << L" and its descendants because this item, or one of its ancestors, has reused the "
+                                               L"identifier of a deleted item.");
+    }
+
+    _localReusedIds.clear();
+}
```
ClementKunz, line 58: Shouldn't we call `_syncPal->removeLocalOperation` on `localId` here?

luc-guyot-infomaniak: This is a very good question, and this is a tricky point: actually, no `Create` operation is generated for the items that re-use IDs of deleted one. But `Create` operations are generated for their descendants (so a newly created empty folder won't cause any issue).

I see two possibilities:
- Either I add a comment explaining this, 
- or I call `_syncPal->removeLocalOperation(localId)` although it has currently no effect, but this might be more robust on the long run.

test/libsyncengine/integration/testintegration.cpp
```diff
@@ -599,33 +599,42 @@ void TestIntegration::testNodeIdReuseFile2DirAndDir2File() {
...
+    // Replace the directory with a file on the local side with the same ID.
     _syncPal->pause();
     while (!_syncPal->isPaused()) {
         Utility::msleep(100);
     }
     IoHelper::deleteItem(absoluteLocalWorkingDir / "testNodeIdReuseDir", ioError);
     CPPUNIT_ASSERT_EQUAL(IoError::Success, ioError);
-    { const std::ofstream file((absoluteLocalWorkingDir / "testNodeIdReuseFile").string()); }
+    { const std::ofstream file(absoluteLocalWorkingDir / "testNodeIdReuseFile"); }
     CPPUNIT_ASSERT_EQUAL(IoError::Success, ioError);
```
ClementKunz, line 632: Not sure to understand this test... Why would `ioError` value have changed?

src/libsyncengine/update_detection/file_system_observer/computefsoperationworker.cpp
luc-guyot-infomaniak: I opted for the second solution: f25148dc3682fe167ea62f2e073fa0f8e50e27bf

test/libsyncengine/integration/testintegration.cpp
luc-guyot-infomaniak: Good catch, fixed with f25148dc3682fe167ea62f2e073fa0f8e50e27bf.

ClementKunz (approved):

PR #1004, opened 2025-08-11
Bring back tray icon

ClementKunz:
App icon is not visible anymore on macOS 26 (beta)

It seems that the system tray icon now must be instantiated from the main process.


ClementKunz:
Icon display fixed in macOS 26 beta 6

PR #1006, opened 2025-08-11
refactor(cmake): avoid duplicate files inside the AppImage.

R-Gld:
On Linux, inside the AppImage filesystem (SquashFS), two filenames are duplicated:
```bash
$ find squashfs-root/ \( -iname "libsentry.so" -o -iname "sync-exclude.lst" \)
squashfs-root/kDrive/libsentry.so
squashfs-root/kDrive/sync-exclude.lst
squashfs-root/usr/bin/sync-exclude.lst
squashfs-root/usr/lib/libsentry.so
```
This PR keeps only one copy of each.

You can also find any other duplicated files with the following commands:
```bash
./kDrive.AppImage --appimage-extract
find ./squashfs-root -type f -printf "%f\n" 2>/dev/null | sort | uniq -d | while read -r filename; do
  echo "duplicate: $filename"
  find ./squashfs-root -type f -name "$filename"
  echo
done
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -293,14 +293,11 @@ elseif(BUILD_CLIENT)
         if(CMAKE_BUILD_TYPE STREQUAL "Debug")
             file(COPY sync-exclude-linux.lst DESTINATION ${CMAKE_BINARY_DIR}/bin)
             file(RENAME ${CMAKE_BINARY_DIR}/bin/sync-exclude-linux.lst ${CMAKE_BINARY_DIR}/bin/sync-exclude.lst)
-        else()
-            install(FILES sync-exclude-linux.lst DESTINATION ${SYSCONFDIR}/${APPLICATION_SHORTNAME} RENAME sync-exclude.lst)
         endif()
 
         # Sentry
         find_library(SENTRY_SHARED_LIBRARY NAMES sentry NO_CACHE)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -293,14 +293,11 @@ elseif(BUILD_CLIENT)
         if(CMAKE_BUILD_TYPE STREQUAL "Debug")
             file(COPY sync-exclude-linux.lst DESTINATION ${CMAKE_BINARY_DIR}/bin)
             file(RENAME ${CMAKE_BINARY_DIR}/bin/sync-exclude-linux.lst ${CMAKE_BINARY_DIR}/bin/sync-exclude.lst)
-        else()
-            install(FILES sync-exclude-linux.lst DESTINATION ${SYSCONFDIR}/${APPLICATION_SHORTNAME} RENAME sync-exclude.lst)
```
luc-guyot-infomaniak, line 297: Wouldn't it be simpler to remove the installation of `sync-exclude.lst` from the `bash` building scripts? (Note however that `CMake` does not seem to install it at the right location).

This way `CMake` would be responsible for the installation of this exclusion file for both the `Release` and the `Debug` build.

R-Gld: You’re right, this seems like a better idea.

PR #1008, opened 2025-08-12
Deactivate 50 file limitation in blacklist

ClementKunz:


src/gui/guiutility.cpp
```diff
@@ -633,15 +633,17 @@ QString GuiUtility::getDateForCurrentLanguage(const QDateTime &dateTime, const Q
 }
 
 bool GuiUtility::checkBlacklistSize(const size_t blacklistSize, QWidget *parent) {
```
luc-guyot-infomaniak, line 635: Is it going to break som integration test?

luc-guyot-infomaniak (approved):

PR #1010, opened 2025-08-13
Blacklist not updated correctly

ClementKunz:
If a folder was partially checked in the tree view and the user check it (or uncheck it) without developing the tree, the children were kept into the blacklist.

The children of the checked or unchecked nodes are now removed from the blacklist.

### DIFF
--- a/src/gui/foldertreeitemwidget.cpp
+++ b/src/gui/foldertreeitemwidget.cpp
@@ -196,16 +196,15 @@ ExitCode FolderTreeItemWidget::updateBlackUndecidedSet() {
 }
 
 void FolderTreeItemWidget::updateBlacklistPathMap() {
-    for (int i = 0; i < 2; i++) {
+    for (auto i = 0; i < 2; i++) {
         for (const QString &nodeId: i == 0 ? _oldBlackList : _oldUndecidedList) {
             QString path;
-            ExitCode exitCode = GuiRequests::getNodePath(_syncDbId, nodeId, path);
-            if (exitCode != ExitCode::Ok) {
+            if (const auto exitCode = GuiRequests::getNodePath(_syncDbId, nodeId, path); exitCode != ExitCode::Ok) {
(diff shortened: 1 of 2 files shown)


src/gui/foldertreeitemwidget.cpp
```diff
@@ -272,27 +271,43 @@ void FolderTreeItemWidget::createBlackSet(QTreeWidgetItem *parentItem, QSet<QStr
...
-        for (int i = 0; i < parentItem->childCount(); ++i) {
+        for (auto i = 0; i < parentItem->childCount(); ++i) {
             createBlackSet(parentItem->child(i), blackset);
         }
     }
 }
 
+void FolderTreeItemWidget::removeChildNodeFromSet(const QString &parentId, QSet<QString> &blackset) {
+    const auto parentPath = _blacklistCache[parentId];
+    for (auto it = _blacklistCache.begin(); it != _blacklistCache.end(); it++) {
```
luc-guyot-infomaniak, line 301:
```suggestion
    for (const auto &[id, path] : _blacklistCache) {
```

src/gui/foldertreeitemwidget.h
```diff
@@ -77,7 +77,7 @@ class FolderTreeItemWidget : public QTreeWidget {
         static const QColor _sizeTextColor;
         bool _inserting;
         QHash<QString, QTreeWidgetItem *> _subFoldersMap;
-        QHash<QString, QString> _blacklistCache;
+        QHash<QString, QString> _blacklistCache; // nodeId / path
```
luc-guyot-infomaniak, line 80:
```suggestion
        using NodeId = QString;
        using Path = QString;
        QHash<NodeId, Path> _blacklistCache;
```
Then re-use `NodeId` and `Path` in  the class wherever they can clarify something.

ClementKunz: Not a fan to use `NodeId` since it is already a used type in the project.
What do you think about fd5879180005c11782f9c30413635a4dd43451dc

PR #1014, opened 2025-08-14
Add poco to conan

R-Gld:


CMakeLists.txt
```diff
@@ -327,6 +326,7 @@ if(UNIX AND NOT APPLE)
         get_library_dirs("OpenSSL" "openssl")
         get_library_dirs("xxHash" "xxhash")
         get_library_dirs("sentry" "sentry")
+        get_library_dirs("Poco" "poco") # poco_LIB_DIRS_RELWITHDEBINFO
```
ClementKunz, line 329: why `RELWITHDEBINFO` here? Shouldn't we start using only `Release` version of the 3rd party libs?

PR #1025, opened 2025-08-20
fix(action.yml): use -a flag for cp commands to preserve file attributes

R-Gld:


R-Gld:
Note: GitHub stores artifacts as .zip archives and automatically dereferences symlinks. As a result, we can’t rely on symlinks for the `.so` files to reduce the artifact size.

PR #1032, opened 2025-08-25
HandleError takes a decompressed string as argument

ClementKunz:
The reply of the `/listing/full` request (and potentially others) is compressed, even in case of errors. However, the app decompresses the body only if the request is successful, leading to JSON parsing errors when the backend returns an HTTP error.
This PR aims to decompress the reply's body when the header `content-encoding` has value `gzip`.

Depends on #1031


ClementKunz:
> It would be nice to have some unit tests.

Done in fc8041b435cd759f8af7ef3bf1b8e77913fa5545

luc-guyot-infomaniak (dismissed):

luc-guyot-infomaniak (approved):

PR #1048, opened 2025-09-04
[TRELLO-1091] Do not issue file not found error when download is cancelled

luc-guyot-infomaniak:


src/libsyncengine/jobs/network/kDrive_API/downloadjob.cpp
```diff
@@ -87,11 +87,14 @@ DownloadJob::~DownloadJob() {
         }
 
         // TODO: usefull ?
-        if (const ExitInfo exitInfo = _vfs->forceStatus(_localpath, VfsStatus()); !exitInfo) {
-            LOGW_WARN(_logger, L"Error in vfsForceStatus: " << Utility::formatSyncPath(_localpath) << L": " << exitInfo);
-        }
+        bool exists = false;
```
ChristopheLarchier, line 90: Why is setPinState not handled as forceStatus?

herve-er (commented):
Suggestion:
It seems that modifying the condition in `downloadjob` at line 84 from
`if (_responseHandlingCanceled) {`
to
`if (_responseHandlingCanceled && !_isCreate) {`

should be sufficient and also prevent attempting to set the pin state and call `_vfs->cancelHydrate(_localpath);` on a file that does not exist.

src/libsyncengine/jobs/network/kDrive_API/downloadjob.cpp
```diff
@@ -89,7 +89,8 @@ DownloadJob::~DownloadJob() {
         // TODO: usefull ?
         bool exists = false;
         if (auto ioError = IoError::Unknown; IoHelper::checkIfPathExists(_localpath, exists, ioError) && exists) {
-            if (const ExitInfo exitInfo = _vfs->forceStatus(_localpath, VfsStatus()); !exitInfo) {
+            if (const ExitInfo exitInfo = _vfs->forceStatus(_localpath, VfsStatus());
+                !exitInfo && exitInfo.cause() != ExitCause::NotFound) {
                 LOGW_WARN(_logger, L"Error in vfsForceStatus: " << Utility::formatSyncPath(_localpath) << L": " << exitInfo);
             }
```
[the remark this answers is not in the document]
luc-guyot-infomaniak: Sure, but `Vfs::checkIfPathIsValid` expects the file indicated by the input path to exist and returns an error if doesn't. The intent of these changes is to avoid propagating the error when non-existence is expected, which is likely to happen after a download cancellation (it does happen in some reported cases).

luc-guyot-infomaniak:
> Suggestion: It seems that modifying the condition in `downloadjob` at line 84 from `if (_responseHandlingCanceled) {` to `if (_responseHandlingCanceled && !_isCreate) {`
> 
> should be sufficient and also prevent attempting to set the pin state and call `_vfs->cancelHydrate(_localpath);` on a file that does not exist.

The flag `_isCreate` indicates whether the intent of the download is to create a local file (as opposed to editing the content of an existing file). So the operations on the file status done in the `DownloadJob` destructor should be skipped only if the job has been cancelled, `_isCreate` is `true` and the file does not exist. Is this what you suggested?

herve-er:
> > Suggestion: It seems that modifying the condition in `downloadjob` at line 84 from `if (_responseHandlingCanceled) {` to `if (_responseHandlingCanceled && !_isCreate) {`
> > should be sufficient and also prevent attempting to set the pin state and call `_vfs->cancelHydrate(_localpath);` on a file that does not exist.
> 
> The flag `_isCreate` indicates whether the intent of the download is to create a local file (as opposed to editing the content of an existing file). So the operations on the file status done in the `DownloadJob` destructor should be skipped only if the job has been cancelled, `_isCreate` is `true` and the file does not exist. Is this what you suggested?

It is. This way, we keep the “file not found” error only for cases where we actually expect the file to exist. Otherwise, in the case of a creation, we simply don’t try to do anything with the file.

https://github.com/Infomaniak/desktop-kDrive/pull/1048/files/822a56e03ad20956bf17dd17b7c5359e616bdd06..86e98b101770e40ea31f284df7a1f2944c313813#r2386844609

src/libsyncengine/jobs/network/kDrive_API/downloadjob.cpp
herve-er, line 84:
```suggestion
    if (_responseHandlingCanceled) {
        if(_isCreate) return;
```

luc-guyot-infomaniak: I have written something similar here f8502fccf610ac25560ae83ef196bebd0b09107d.

PR #1052, opened 2025-09-09
Trello 1097 Socket commServer/Channel

herve-er:


test/server/comm/testsocketcomm.cpp
```diff
@@ -0,0 +1,224 @@
+﻿/*
```
herve-er, line 1: hese tests will be moved to `testGuiComm/testExtComm`, which now handle `sendMessage`/`readMessage`.

PR #1062, opened 2025-09-19
[TRELLO-1107] Handle and log every error raised in `Utility::moveItemToTrash` (Windows)

luc-guyot-infomaniak:
This pull request follows-up with reported failures when moving items to trash that 
- do not raise any error,
- do not leave any trace in logs.

The proposed changes make sure the remaining possible errors are handled and their messages are logged.

### DIFF
--- a/test/libcommonserver/utility/testutility.cpp
+++ b/test/libcommonserver/utility/testutility.cpp
@@ -148,7 +148,6 @@ void TestUtility::testMoveItemToTrash(void) {
     CPPUNIT_ASSERT(!std::filesystem::exists(dirPath));
 
     // A regular file within a subdirectory that misses owner exec permission:
-    // - No issue on Windows
     const SyncPath subdir = tempDir.path() / "permission_less_subdirectory";
     (void) std::filesystem::create_directory(subdir);
     path = subdir / "file.txt";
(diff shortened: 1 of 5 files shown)


test/libcommonserver/utility/testutility.cpp
```diff
@@ -145,11 +145,33 @@ void TestUtility::testMoveItemToTrash(void) {
...
     CPPUNIT_ASSERT(Utility::moveItemToTrash(dirPath));
     CPPUNIT_ASSERT(!std::filesystem::exists(dirPath));
+
+    // A regular file within a subdirectory that misses owner exec permission:
+    // - No issue on Windows
+    const SyncPath subdir = tempDir.path() / "permission_less_subdirectory";
+    (void) std::filesystem::create_directory(subdir);
+    path = subdir / "file.txt";
+    { std::ofstream ofs(path); }
+    std::filesystem::permissions(subdir, std::filesystem::perms::owner_exec, std::filesystem::perm_options::remove);
```
herve-er, line 155: `std::filesystem::permissions` isn’t working as expected on Windows (NTFS). It only changes the read-only flag of the directory, which is purely indicative and doesn’t modify the file’s ACL.

You can use `IoHelper::setRights` instead, which tries to match ACLs as closely as possible to the Unix permission system.

ClementKunz: Following Herve's comment, you used `IoHelper::setRights` only on Windows, why? Why not using only `IoHelper::setRights`?

luc-guyot-infomaniak: @ClementKunz Thanks, done with 29b4003542b802d9f2f126a418b4f5e88def1509.

ClementKunz (approved):

PR #1039, opened 2025-08-28
feat(macOS UI): SplitView Sidebar

valentinperignon:
Depends on #1036

PR #1069, opened 2025-09-24
feat(macOS UI): Open Settings Window

valentinperignon:
Depends on #1039 

https://github.com/user-attachments/assets/7199d119-f761-4b1a-91d3-41f37c6e49c2

### DIFF
--- a/src/front/macOS/kDrive/Controllers/PreferencesWindow/SplitView/PreferencesSidebarViewController.swift
+++ b/src/front/macOS/kDrive/Controllers/PreferencesWindow/SplitView/PreferencesSidebarViewController.swift
@@ -26,7 +26,7 @@ extension SidebarItem {
 }
 
 class PreferencesSidebarViewController: NSViewController {
-    static let navigationCellIdentifier = NSUserInterfaceItemIdentifier("NavigationSidebarCell")
+    static let navigationCellIdentifier = NSUserInterfaceItemIdentifier(String(describing: SidebarTableCellView.self))
 
     weak var delegate: NavigableSidebarViewControllerDelegate?
 
--- a/src/front/macOS/kDrive/Controllers/MainWindow/SplitView/MainSidebarViewController.swift
+++ b/src/front/macOS/kDrive/Controllers/MainWindow/SplitView/MainSidebarViewController.swift
@@ -41,7 +41,7 @@ extension SidebarItem {
 }
 
 final class MainSidebarViewController: NSViewController {
-    static let navigationCellIdentifier = NSUserInterfaceItemIdentifier("NavigationSidebarCell")
+    static let navigationCellIdentifier = NSUserInterfaceItemIdentifier(String(describing: SidebarTableCellView.self))
 
     weak var delegate: NavigableSidebarViewControllerDelegate?
 


src/front/macOS/kDrive/Resources/Generated/Strings+Generated.swift
```diff
@@ -14,8 +14,12 @@ internal enum KDriveLocalizable {
   internal static let buttonOpenInBrowser = KDriveLocalizable.tr("Localizable", "buttonOpenInBrowser", fallback: "Open in browser")
   /// loco:68b04d5a63085b7ab90cea63
   internal static let buttonOpenInFinder = KDriveLocalizable.tr("Localizable", "buttonOpenInFinder", fallback: "Open in Finder")
-  /// loco:68a7228730d0f594cc03e4b3
-  internal static let menuItemSearch = KDriveLocalizable.tr("Localizable", "menuItemSearch", fallback: "Search")
+  /// loco:68cd386502633dee14000352
```
luc-guyot-infomaniak, line 17: It would be nice to document how these `loco` identifiers have been generated.

src/front/macOS/kDrive/Views/MainMenu.swift
```diff
@@ -0,0 +1,185 @@
...
+    }
+
+    private func setupMenuItems() {
+        let applicationItem = addItem(withTitle: "Application", action: nil, keyEquivalent: "")
+        setSubmenu(createApplicationMenu(), for: applicationItem)
+
+        let fileItem = addItem(withTitle: "File", action: nil, keyEquivalent: "")
+        setSubmenu(createFileMenu(), for: fileItem)
+
+        let editItem = addItem(withTitle: "File", action: nil, keyEquivalent: "")
```
luc-guyot-infomaniak, line 51: Is the title the expected one?

luc-guyot-infomaniak (commented):
Nice to have: screenshots added to the the PR description that exemplify the features under development.

src/front/macOS/kDrive/Resources/Generated/Strings+Generated.swift
valentinperignon: That's an internal ID for Loco, and this file is 100% generated by Loco, so there is not real point in documenting those IDs.

src/front/macOS/kDrive/Controllers/MainWindow/MainWindowController.swift
```diff
@@ -0,0 +1,47 @@
...
+
+    init() {
+        let window = NSWindow(
+            contentRect: NSRect(x: 0, y: 0, width: 900, height: 600),
+            styleMask: [.titled, .closable, .resizable, .miniaturizable, .fullSizeContentView],
+            backing: .buffered,
+            defer: false
+        )
+
+        window.title = "kDrive"
```
adrien-coye, line 32: Do we want to i18n this or have it in a `Constants.swift` file  ?

src/front/macOS/kDrive/Controllers/PreferencesWindow/SplitView/PreferencesSidebarViewController.swift
```diff
@@ -0,0 +1,116 @@
...
+import kDriveCoreUI
+
+extension SidebarItem {
+    static let general = SidebarItem(icon: NSImage(resource: .settings), title: KDriveLocalizable.sidebarItemGeneral)
+    static let accounts = SidebarItem(icon: NSImage(resource: .persons), title: KDriveLocalizable.sidebarItemAccounts)
+    static let advanced = SidebarItem(icon: NSImage(resource: .hammerWrench), title: KDriveLocalizable.sidebarItemAdvanced)
+}
+
+class PreferencesSidebarViewController: NSViewController {
+    static let navigationCellIdentifier = NSUserInterfaceItemIdentifier("NavigationSidebarCell")
```
adrien-coye, line 29: I think I prefer the following, as the compilation will fail on a partial rename. If you agree you can apply this pattern on each of your `NSUserInterfaceItemIdentifier`

```suggestion
    static let navigationCellIdentifier = NSUserInterfaceItemIdentifier(String(describing: NavigationSidebarCell.self))
```

src/front/macOS/kDrive/Controllers/PreferencesWindow/PreferencesWindowController.swift
```diff
@@ -0,0 +1,48 @@
...
+ */
+
+import Cocoa
+
+final class PreferencesWindowController: NSWindowController {
+    private var preferencesViewController: PreferencesSplitViewController!
+
+    init() {
+        let window = NSWindow(
+            contentRect: NSRect(x: 0, y: 0, width: 700, height: 400),
```
adrien-coye, line 26: Same comment about consts for window sizes.

src/front/macOS/kDrive/Views/MainMenu.swift
```diff
@@ -0,0 +1,185 @@
...
+
+import Cocoa
+
+final class MainMenu: NSMenu {
+    private(set) var servicesMenu: NSMenu?
+    private(set) var windowMenu: NSMenu?
+    private(set) var helpMenu: NSMenu?
+
+    private var applicationName: String {
+        return Bundle.main.object(forInfoDictionaryKey: "CFBundleName") as? String ?? "kDrive"
```
adrien-coye, line 27: Instead of the hardcoded "kDrive" string you use for a window name, you could use this as well.

src/front/macOS/kDrive/Views/MainMenu.swift
```diff
@@ -0,0 +1,185 @@
...
+    }
+
+    convenience init() {
+        self.init(title: "MainMenu")
```
adrien-coye, line 31: Do we want to i18n this ? (this goes as well for the other ones like this)

src/front/macOS/kDriveCoreUI/Utils/SidebarItem.swift
```diff
@@ -0,0 +1,41 @@
...
+import Foundation
+import Cocoa
+
+public struct SidebarItem: Sendable, Equatable, Hashable {
+    public enum ItemType: Sendable, Equatable {
+        case navigation
+        case menu
+    }
+
+    public let icon: NSImage
```
adrien-coye, line 28: There is an argument about preferring a `class` to a `struct` when we encapsulate reference types (objects) in one. I'd like to discuss this with you.

src/front/macOS/kDrive/Controllers/MainWindow/MainWindowController.swift
valentinperignon: I eventually removed the window title in a later PR. The window is named after the title of the selected SplitView item.
I will directly delete it in this PR.

src/front/macOS/kDrive/Views/MainMenu.swift
valentinperignon: You're right, we should add l8n to the menu.
However, I need to explore the best way to implement it, so I’ll handle it later. For example, there are open-source tools that can generate the localization files.

src/front/macOS/kDriveCoreUI/Utils/SidebarItem.swift
valentinperignon: Interesting, let's discuss together 😊

adrien-coye (approved):
Thanks for the feedback. All good to me.

luc-guyot-infomaniak (commented):
I am willing to approve. I would be even more convinced if screenshots were added to the PR description in order to illustrate the features under development.

valentinperignon:
@luc-guyot-infomaniak There is now a video on the PR description

luc-guyot-infomaniak (approved):

PR #1085, opened 2025-10-01
feat: XPC shared object library

adrien-coye:
### Abstract

To be able to exchange objects via XPC between processes, a simple library with exchange objects that implement `NSSecureCoding` can be useful.

### Details

This will be written and tested in Swift.

Objects in here will also provide a view and init with the JSON format the kext uses for the sake of use simplicity.

This library will also be tested with standard issue SPM test target.


adrien-coye:
As discussed with @PhilippeWeidmann we drop this library to work with JSON like the Windows app.

PR #1096, opened 2025-10-07
[TRELLO-1117] Fixes incorrect update of the folders excluded by the user via GUI

luc-guyot-infomaniak:
Accessing non-existing items in the excluded folder paths cache had the side effect to remove folders that were purposely excluded by the user.

### DIFF
--- a/src/gui/foldertreeitemwidget.cpp
+++ b/src/gui/foldertreeitemwidget.cpp
@@ -157,18 +157,19 @@ ExitCode FolderTreeItemWidget::updateBlackUndecidedSet() {
     if (!_syncDbId) return ExitCode::Ok;
 
     bool userConnected = false;
-    if (auto userInfoIt = _gui->userInfoMap().find(_userDbId); userInfoIt != _gui->userInfoMap().end()) {
+    if (const auto userInfoIt = _gui->userInfoMap().find(_userDbId); userInfoIt != _gui->userInfoMap().end()) {
         userConnected = userInfoIt->second.connected();
     }
 
     if (!userConnected) return ExitCode::Ok;
 
-    if (auto exitCode = GuiRequests::getSyncIdSet(_syncDbId, SyncNodeType::BlackList, _oldBlackList); exitCode != ExitCode::Ok) {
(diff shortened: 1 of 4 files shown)


src/gui/foldertreeitemwidget.cpp
```diff
@@ -304,14 +280,16 @@ void FolderTreeItemWidget::createBlackSet(const QTreeWidgetItem *parentItem, QSe
 }
 
 void FolderTreeItemWidget::removeChildNodeFromSet(const QString &parentId, QSet<QString> &blackset) {
+    assert(_blacklistCache.contains(parentId) && "parentId was not found in black list cache.");
```
luc-guyot-infomaniak: Will prevent regression.

src/gui/foldertreeitemwidget.cpp
```diff
@@ -287,7 +264,6 @@ void FolderTreeItemWidget::createBlackSet(const QTreeWidgetItem *parentItem, QSe
             return;
         }
         case Qt::Checked:
-            removeChildNodeFromSet(parentNodeId, blackset);
```
luc-guyot-infomaniak, line 290: This is the fix: If `parentNodeId`is not in the cache, valid excluded folders are removed as a harmful side effect of  `removeChildNodeFromSet`. Note that checked children are removed automatically through recursion.

ChristopheLarchier (dismissed):

src/gui/foldertreeitemwidget.cpp
```diff
@@ -262,56 +245,60 @@ QSet<QString> FolderTreeItemWidget::createBlackSet() {
...
-            (void) blackset.insert(parentNodeId);
-            removeChildNodeFromSet(parentNodeId, blackset);
+            removeChildNodeFromSet(path, blackset);
             return;
         }
-        case Qt::Checked:
-            removeChildNodeFromSet(parentNodeId, blackset);
-            [[fallthrough]];
+        case Qt::Checked: {
+            const QString path = getPath(parentNodeId);
```
luc-guyot-infomaniak, line 278: This is the fix: ensures that we always have a valid path for `parentNodeId` before removing its children.

ClementKunz (approved):

PR #1101, opened 2025-10-08
Check cycles only if no move operations

ClementKunz:
Cycles might occur only if at least 1 MOVE operation exists. Therefore, a simple optimization consists in **not** checking for cycles only if there are no move operations to propagate. This might considerably reduce the duration of the `OperationSorter` step for the very first synchronization of big drives.

### DIFF
--- a/src/libsyncengine/propagation/operation_sorter/operationsorterworker.cpp
+++ b/src/libsyncengine/propagation/operation_sorter/operationsorterworker.cpp
@@ -69,12 +69,18 @@ void OperationSorterWorker::sortOperations() {
         fixEditBeforeMove();
         fixMoveBeforeMoveHierarchyFlip();
 
-        CycleFinder cycleFinder(_reorderings);
-        cycleFinder.findCompleteCycle();
-        if (cycleFinder.hasCompleteCycle()) {
-            completeCycle = cycleFinder.completeCycle();
-            cycleFound = true;
-            break;
+        // Cycles can occur only in presence of at least 1 Move operation.
+        if (!_syncPal->_syncOps->opListIdByType(OperationType::Move).empty() ||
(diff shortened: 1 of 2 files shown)


src/libsyncengine/propagation/operation_sorter/operationsorterworker.cpp
```diff
@@ -69,12 +69,18 @@ void OperationSorterWorker::sortOperations() {
...
         fixMoveBeforeMoveHierarchyFlip();
 
-        CycleFinder cycleFinder(_reorderings);
-        cycleFinder.findCompleteCycle();
-        if (cycleFinder.hasCompleteCycle()) {
-            completeCycle = cycleFinder.completeCycle();
-            cycleFound = true;
-            break;
+        // Cycles can occur only in presence of at least 1 Move operation.
+        if (!_syncPal->_syncOps->opListIdByType(OperationType::Move).empty() ||
```
luc-guyot-infomaniak: **Suggestion.** Design a helper function as a method or as a lambda that returns `true` if and only if there is at least on Move operation. 

Are there some unit tests that could adapted so that the two scenarios are actually happening (no Move operation, more than on Move operations and an actual cycle)?

ClementKunz: Helper function done in 640afff13a22847c75d0a72a4659d2c84e2a9938

ClementKunz: About a unit test, it is hard to extract the information whether the cycles have been check, without introducing a variable used only for testing. However, an integration test exists on breaking a cycle which already ensure that this change does not cause any regression

ChristopheLarchier (approved):

PR #1102, opened 2025-10-09
WiniUI3 Setup server communication

herve-er:
## ✨ Pull Request Overview

This PR introduces the **communication architecture** for the **WinUI3 client**, enabling structured and reliable interactions with the server.

It establishes a **socket-based communication protocol** and a **dedicated service layer**, ensuring maintainability, scalability, and thread-safe data handling across the application.

---

### 🚀 Key Changes

* **Communication protocol**

  * Implemented a socket-based client for live communication with the server
  * Added a mock implementation for offline testing and development

* **Server communication service layer**

  * Introduced a centralized service for message handling and data synchronization
  * Abstracted server protocol details from higher-level components

* **View model updates**

  * Refactored all view models to use **thread-safe property setters**
  * Improved UI responsiveness and state consistency

depends on #1105


src/server/comm/guijobs/unknownrequestjob.cpp
```diff
@@ -0,0 +1,64 @@
...
+    writeParamValue(outParamsError, CommonUtility::str2CommString(_error));
+    writeParamValue(outParamsErrorDescr, CommonUtility::str2CommString(_errorDescr));
+    return true;
+}
+
+bool UnknownRequestJob::process() {
+    _exitInfo = ExitCode::LogicError;
+    _error = "unknown_request";
+    _errorDescr = "The request is unknown by the server application.";
+    return _exitInfo;
```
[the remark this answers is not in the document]
herve-er: `_exitInfo` can safely be converted to bool

src/gui4/windows/kDrive client/kDrive client/ServerCommunication/Services/MockServerCommProtocol.cs
```diff
@@ -0,0 +1,287 @@
...
+                }
+                ((JsonArray)result["syncInfo"]!).Add(syncData);
+
+            }
+
+            return new CommData
+            {
+                Type = CommMessageType.Request,
+                Id = (int)NextId,
+                RequestNum = RequestNum.AccountInfoList,
```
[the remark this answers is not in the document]
herve-er: Done

src/gui4/windows/kDrive client/kDrive client/ServerCommunication/Services/SocketServerCommProtocol.cs
```diff
@@ -0,0 +1,308 @@
...
+                if (_pendingRequests.TryGetValue(requestId, out var reply))
+                {
+                    if (reply != null)
+                    {
+                        _pendingRequests.Remove(requestId, out _);
+                        return reply;
+                    }
+                    else
+                    {
+                        Logger.Log(Logger.Level.Debug, $"Found request(Id) {requestId}, but the respons is not available yet null");
```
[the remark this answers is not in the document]
herve-er: Done

src/server/comm/guijobs/unknownrequestjob.cpp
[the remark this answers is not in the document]
herve-er: Done

src/gui4/windows/kDrive client/kDrive client/ServerCommunication/Services/SocketServerCommProtocol.cs
```diff
@@ -0,0 +1,308 @@
...
+                    }
+                }
+                else
+                {
+                    Logger.Log(Logger.Level.Debug, $"No request(Id) found for {requestId}");
+                }
+                await Task.Delay(10).ConfigureAwait(false);
+                if (canceletionToken.IsCancellationRequested)
```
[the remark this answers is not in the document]
herve-er: Done

src/gui4/windows/kDrive client/kDrive client/Utility/UISafeObservableObject.cs
```diff
@@ -0,0 +1,74 @@
...
+                return false;
+            }
+            OnPropertyChangedInUIThread(propertyName);
+            field = newValue;
+            OnPropertyChangingInUIThread(propertyName);
+            return true;
+
+        }
+
+        protected void OnPropertyChangedInUIThread([CallerMemberName] string? propertyName = null)
```
ClementKunz, line 57: Couldn't simply overload `OnPropertyChanged` in order to avoid changing all the call to this method everywhere?

src/gui4/windows/kDrive client/kDrive client/Utility/UISafeObservableObject.cs
```diff
@@ -0,0 +1,74 @@
...
+using System.Diagnostics.CodeAnalysis;
+using System.Linq;
+using System.Runtime.CompilerServices;
+using System.Text;
+using System.Threading.Tasks;
+using System.Windows.Threading;
+
+namespace Infomaniak.kDrive
+{
+    public class UISafeObservableObject : ObservableObject
```
ClementKunz: Can you briefly explain why you needed to overload this class?
Why those properties are changed in UI thread (causing some concurrency issues if I understand correctly)?

src/gui4/windows/kDrive client/kDrive client/ViewModels/App/User.cs
```diff
@@ -17,132 +17,141 @@
...
         {
             get => _avatar;
-            set => SetProperty(ref _avatar, value);
+            set
+            {
+                if (SetPropertyInUIThread(ref _avatar, value))
+                {
+                    AppModel.UIThreadDispatcher.TryEnqueue(async () =>
+                    {
+                        _avatarImageSource = await ByteArrayToImageSource(_avatar); // ByteArrayToImageSource need to be run in the UI thread
```
ClementKunz, line 105: Why? This sounds like a potentially heavy task that could lock temporary the IU thread?

src/gui4/windows/kDrive client/kDrive client/Utility/UISafeObservableObject.cs
herve-er: `OnPropertyChanged` isn’t virtual and therefore cannot be properly overridden. At best, it could be hidden by another method with the same name, which would generate a warning. Even though this warning can be suppressed using the `new` keyword, I don’t think it’s good practice to hide such a common function.

I prefer to keep it explicit for the caller whether they’re invoking the standard `OnPropertyChanged` method or the version that runs on the UI thread. Some private properties are not related to the UI and therefore don’t need the overhead of `OnPropertyChangedInUIThread`; they can safely be changed from any thread.

src/gui4/windows/kDrive client/kDrive client/Utility/UISafeObservableObject.cs
herve-er: Comment added in code base ->

```
/// <summary>
/// Provides a thread-safe version of <see cref="ObservableObject"/> that ensures
/// property change notifications are always raised on the UI thread when using
/// <see cref="SetPropertyInUIThread"/>.
///
/// This class is particularly useful for view models whose properties are bound
/// to UI elements but may be updated from background threads. Since UI elements
/// can only be modified from the UI thread, this class helps prevent cross-thread
/// access issues and simplifies synchronization logic.
///
/// By using <see cref="SetPropertyInUIThread"/>, you avoid the need to manually
/// dispatch property change notifications to the UI thread in every background
/// update scenario.
///
/// For properties that are not related to the UI, the regular
/// <see cref="ObservableObject.SetProperty"/> method can still be used to avoid
/// unnecessary thread marshalling overhead.
/// </summary>
```

src/gui4/windows/kDrive client/kDrive client/ViewModels/App/User.cs
herve-er: ImageSource can only be constructed on the UI thread.

However, the lambda is async and `ByteArrayToImageSource` is awaited, so it will run on a different thread, keeping the UI thread free to do other work and preventing it from freezing.

ClementKunz (approved):

PR #1110, opened 2025-10-13
Do not merge 2 real nodes

ClementKunz:
A client had a recurring error on step 3 (UpdateTreeWorker) in `updateNodeWithDb`. After analyzing his logs, we figured out that in some cases, 2 real nodes might be merged into 1 node (with both Move and Delete operations associated).

The fix is [here](https://github.com/Infomaniak/desktop-kDrive/pull/1110#discussion_r2438682342), where we previously did not check if a node was temporary before merging it with another real node.

### DIFF
--- a/src/libsyncengine/propagation/operation_sorter/operationsorterworker.cpp
+++ b/src/libsyncengine/propagation/operation_sorter/operationsorterworker.cpp
@@ -102,8 +102,11 @@ void OperationSorterWorker::fixDeleteBeforeMove() {
 
         const auto deleteNode = deleteOp->affectedNode();
         LOG_IF_FAIL(deleteNode)
+        const auto parentPath = deleteNode->parentNode()->hasChangeEvent(OperationType::Move)
+                                        ? deleteNode->parentNode()->moveOriginInfos().path()
+                                        : deleteNode->parentNode()->getPath();
         NodeId deleteNodeParentId;
-        if (!getIdFromDb(deleteNode->side(), deleteNode->getPath().parent_path(), deleteNodeParentId)) {
+        if (!getIdFromDb(deleteNode->side(), parentPath, deleteNodeParentId)) {
             continue;
         }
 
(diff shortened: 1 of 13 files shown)


src/libsyncengine/propagation/operation_sorter/operationsorterworker.cpp
```diff
@@ -188,8 +191,11 @@ void OperationSorterWorker::fixDeleteBeforeCreate() {
 
         const auto deleteNode = deleteOp->affectedNode();
         LOG_IF_FAIL(deleteNode)
+        const auto parentPath = deleteNode->parentNode()->hasChangeEvent(OperationType::Move)
```
luc-guyot-infomaniak, line 194: This is the fix, isn't it? (The description hints toward the update tree).

src/libsyncengine/update_detection/update_detector/updatetreeworker.cpp
```diff
@@ -207,53 +208,54 @@ ExitCode UpdateTreeWorker::step3DeleteDirectory() {
             }
 
             // Check if parentNode has got a child with the same name
-            std::shared_ptr<Node> newNode = parentNode->findChildren(deleteOp->path().filename(), deleteOp->nodeId());
-            if (newNode != nullptr) {
+            std::shared_ptr<Node> existingNode = parentNode->findChildren(deleteOp->path().filename(), deleteOp->nodeId());
+            if (existingNode && existingNode->isTmp()) {
```
ClementKunz, line 212: This is the fix

src/libsyncengine/propagation/operation_sorter/operationsorterworker.cpp
ClementKunz: No, not directly. The fix is [here](https://github.com/Infomaniak/desktop-kDrive/pull/1110#discussion_r2438682342).

This change fix a related bug I discovered while running the new integration test. Basically, in operation sorter, the wrong path was retrieved if the parent of a deleted node has been moved.

luc-guyot-infomaniak: It would be nice to mention it in the description of this PR. This would help track bug fixes.

luc-guyot-infomaniak (approved):

PR #1115, opened 2025-10-15
fix(extended-tests): Fix extended tests failures

luc-guyot-infomaniak:


src/server/comm/pipecommserver.cpp
```diff
@@ -411,18 +411,6 @@ SyncPath PipeCommServer::createPipe() {
 
     const SyncPath pipePath = SyncPath(R"(\\.\pipe\)") / Str2SyncName(name);
 
-    // Delete/create pipe file
```
luc-guyot-infomaniak, line 414: This code is not required (named pipes on Windows are not created on the normal file system) an causes troubles, at least in  extended tests.

ClementKunz (approved):

PR #1116, opened 2025-10-16
Check access to tmp folder

ClementKunz:
On app launch, check if the app have access to the temp folder. If it is not the case, do not start the syncs and show a didecated error message


src/libcommonserver/utility/utility.cpp
```diff
@@ -667,4 +668,63 @@ bool Utility::isError500(const Poco::Net::HTTPResponse::HTTPStatus httpErrorCode
     }
 }
 
+IoError Utility::tryCreateTmpDir(const SyncName &name /*= "testDir"*/) {
+#if defined(KD_MACOS)
+    // Add random suffix to make sure the file does not already exist.
+    const auto filename = name + SyncName(CommonUtility::generateRandomStringAlphaNum());
```
ChristopheLarchier: This can be a problem if the maximum name size is reached.
Wouldn't it be better to test the return code of create_directory?

src/libsyncengine/update_detection/file_system_observer/remotefilesystemobserverworker.cpp
```diff
@@ -341,7 +341,10 @@ ExitInfo RemoteFileSystemObserverWorker::getItemsInDir(const NodeId &dirId, cons
         }
 
         // Check unsupported characters
-        if (hasUnsupportedCharacters(item.name(), item.id(), item.type())) {
+        if (const auto ioError = checkForUnsupportedCharacters(item.name(), item.id(), item.type()); !ioError) {
```
luc-guyot-infomaniak:
```suggestion
        if (const auto exitInfo = checkForUnsupportedCharacters(item.name(), item.id(), item.type()); !ioError) {
```

src/libsyncengine/update_detection/file_system_observer/remotefilesystemobserverworker.cpp
```diff
@@ -480,8 +483,12 @@ ExitInfo RemoteFileSystemObserverWorker::processActions(Poco::JSON::Array::Ptr a
         }
 
         // Check unsupported characters
-        if (hasUnsupportedCharacters(actionInfo.snapshotItem.name(), actionInfo.snapshotItem.id(),
-                                     actionInfo.snapshotItem.type())) {
+        if (const auto ioError = checkForUnsupportedCharacters(actionInfo.snapshotItem.name(), actionInfo.snapshotItem.id(),
```
luc-guyot-infomaniak:
```suggestion
        if (const auto exitInfo = checkForUnsupportedCharacters(actionInfo.snapshotItem.name(), actionInfo.snapshotItem.id(),
```

src/libcommonserver/utility/utility.cpp
```diff
@@ -667,4 +668,63 @@ bool Utility::isError500(const Poco::Net::HTTPResponse::HTTPStatus httpErrorCode
...
+}
+
+IoError Utility::tryCreateTmpFile(const SyncName &name /*= "testFile"*/) {
+    // Add random suffix to make sure the file does not already exist.
+    SyncPath tmpDirPath;
+    if (auto ioError = IoError::Unknown; !IoHelper::tempDirectoryPath(tmpDirPath, ioError)) {
+        return ioError;
+    }
+
+    const auto filename = name + SyncName(CommonUtility::generateRandomStringAlphaNum());
```
ChristopheLarchier: This can be a problem if the maximum name size is reached.

test/libcommonserver/utility/testutility.cpp
```diff
@@ -537,4 +537,16 @@ void TestUtility::testUserName() {
 #endif
 }
 
+void TestUtility::testTryCreateTmpDir() {
```
luc-guyot-infomaniak, line 540: It would be nice to also have failure scenarios that revolve around missing permissions (the initial problem).

src/gui/parametersdialog.cpp
ChristopheLarchier: You could move those 4 cases after SystemError.

src/libsyncengine/update_detection/file_system_observer/remotefilesystemobserverworker.cpp
ClementKunz: Done in 53ad14b019d7776aec6ae5777c3cbba673277de9

src/libsyncengine/update_detection/file_system_observer/remotefilesystemobserverworker.cpp
ClementKunz: Done in 53ad14b019d7776aec6ae5777c3cbba673277de9

src/gui/parametersdialog.cpp
ClementKunz: Done in 0920db299f332bab7d374e48b76caa8895eb7de5

src/libcommonserver/utility/utility.cpp
ClementKunz: Done in bf4d53879f0f19133a05b22e8d39e0b6652dc700.
Return code of `create_directory` is checked below

src/libcommonserver/utility/utility.cpp
ClementKunz: Done in bf4d53879f0f19133a05b22e8d39e0b6652dc700
I just worried that a file with same name might remain in the tmp directory leading to a wrong error "UNICODE ...". What do you think? The file creation using `std::ofstream` does provide any feedback.

ChristopheLarchier: It is quickly removed but...
At worst, if 2 jobs try to create the same file at the same time, one will error out and the file will be excluded from the synchronization. It will be synced next time.

test/libcommonserver/utility/testutility.cpp
ClementKunz: Done in adb3d701263ce0326992c120113d4bd805958673

src/libcommonserver/utility/utility.cpp
ClementKunz: Yes, not a big deal indeed 👍

ChristopheLarchier (dismissed):

ChristopheLarchier (approved):

PR #1126, opened 2025-10-22
Windows - Package signature check

ClementKunz:
Check the signature of the downloaded package before proceeding with the update. If the signature is considered as invalid, the update is aborted.

- [x] Send a sentry alert


src/libcommonserver/utility/digitalsignaturechecker_win.cpp
```diff
@@ -0,0 +1,307 @@
...
+
+#pragma comment(lib, "crypt32.lib")
+
+#define ENCODING (X509_ASN_ENCODING | PKCS_7_ASN_ENCODING)
+
+namespace KDC {
+
+DigitalSignatureChecker_win::DigitalSignatureChecker_win(const SyncPath &packageAbsolutePath) :
+    _packageAbsolutePath(packageAbsolutePath) {
+    // std::wstring errorMsg;
```
luc-guyot-infomaniak: To be removed?

src/libcommonserver/utility/digitalsignaturechecker_win.cpp
```diff
@@ -0,0 +1,307 @@
...
+namespace {
+using SPROG_PUBLISHERINFO = struct {
+        LPWSTR lpszProgramName;
+        LPWSTR lpszPublisherLink;
+        LPWSTR lpszMoreInfoLink;
+};
+using PSPROG_PUBLISHERINFO = SPROG_PUBLISHERINFO *;
+
+BOOL extractCertificateInfo(const PCCERT_CONTEXT pCertContext, DigitalSignatureInfo &signatureInfo, SourceLocation &srcLoc) {
+    LPTSTR szName = nullptr;
```
luc-guyot-infomaniak: We may move this declaration where the variable is initialized in a non-trivial way.

src/libcommonserver/utility/digitalsignaturechecker_win.cpp
```diff
@@ -0,0 +1,307 @@
...
+    DWORD dwData = 0;
```
luc-guyot-infomaniak: We may move this declaration right before the first use of the variable.

src/libcommonserver/utility/digitalsignaturechecker_win.cpp
```diff
@@ -0,0 +1,307 @@
...
+    // Get Serial Number.
+    std::wstringstream ss;
+    for (DWORD i = 0; i < pCertContext->pCertInfo->SerialNumber.cbData; i++) {
+        ss << std::hex << std::uppercase << (int) pCertContext->pCertInfo->SerialNumber.pbData[i] << L" ";
+    }
+    signatureInfo._serialNumber = ss.str();
+
+    // Get Issuer name size.
+    if (!(dwData = CertGetNameString(pCertContext, CERT_NAME_SIMPLE_DISPLAY_TYPE, CERT_NAME_ISSUER_FLAG, nullptr, nullptr, 0))) {
+        if (szName != nullptr) (void) LocalFree(szName);
```
luc-guyot-infomaniak: No allocation has been done so far. So this line does not seem useful.

src/libcommonserver/utility/digitalsignaturechecker_win.cpp
```diff
@@ -0,0 +1,307 @@
...
+
+    // Get Serial Number.
+    std::wstringstream ss;
+    for (DWORD i = 0; i < pCertContext->pCertInfo->SerialNumber.cbData; i++) {
+        ss << std::hex << std::uppercase << (int) pCertContext->pCertInfo->SerialNumber.pbData[i] << L" ";
+    }
+    signatureInfo._serialNumber = ss.str();
+
+    // Get Issuer name size.
+    if (!(dwData = CertGetNameString(pCertContext, CERT_NAME_SIMPLE_DISPLAY_TYPE, CERT_NAME_ISSUER_FLAG, nullptr, nullptr, 0))) {
```
luc-guyot-infomaniak, line 64: Is it the correct function call? The only output parameter is the second to last argument which is set with `nullptr`.

src/libcommonserver/utility/digitalsignaturechecker_win.cpp
```diff
@@ -0,0 +1,307 @@
...
+    if (!(CertGetNameString(pCertContext, CERT_NAME_SIMPLE_DISPLAY_TYPE, CERT_NAME_ISSUER_FLAG, nullptr, szName, dwData))) {
+        if (szName != nullptr) (void) LocalFree(szName);
+        return FALSE;
+    }
+    signatureInfo._issuerName = szName;
+    (void) LocalFree(szName);
+    szName = nullptr;
+
+    // Get Subject name size.
+    if (!(dwData = CertGetNameString(pCertContext, CERT_NAME_SIMPLE_DISPLAY_TYPE, 0, nullptr, nullptr, 0))) {
```
luc-guyot-infomaniak, line 85: Is it the correct call? The variable `szName` is not modified by this function call.

src/libcommonserver/utility/digitalsignaturechecker_win.cpp
```diff
@@ -0,0 +1,307 @@
...
+        if (szName != nullptr) (void) LocalFree(szName);
+        return FALSE;
+    }
+
+    // Allocate memory for subject name.
+    szName = (LPTSTR) LocalAlloc(LPTR, dwData * sizeof(TCHAR));
```
luc-guyot-infomaniak, line 90: If the call to `CertGetNameString` is successful, the allocation is already done.

test/libcommonserver/utility/testdigitalsignaturecheck_win.cpp
```diff
@@ -0,0 +1,39 @@
...
+void TestDigitalSignatureCheck_win::testIsSignatureValid() {
+    // Empty path.
+    CPPUNIT_ASSERT(!DigitalSignatureChecker_win({}).isSignatureValid());
+    // Path to non existing file.
+    CPPUNIT_ASSERT(!DigitalSignatureChecker_win(SyncPath("A/B/C")).isSignatureValid());
+    // Path to existing file but not signed.
+    LocalTemporaryDirectory tmpDir;
+    const auto testPath = tmpDir.path() / "testSignature.txt";
+    testhelpers::generateOrEditTestFile(testPath);
+    CPPUNIT_ASSERT(!DigitalSignatureChecker_win(SyncPath(testPath)).isSignatureValid());
```
luc-guyot-infomaniak: Would it be possible to test a scenario with a positive outcome, mocking some parts if needed?

src/libcommonserver/utility/digitalsignaturechecker_win.cpp
ClementKunz: I took this code from Microsoft example: https://learn.microsoft.com/en-us/previous-versions/troubleshoot/windows/win32/get-information-authenticode-signed-executables

And the output parameter is the returned value

src/libcommonserver/utility/digitalsignaturechecker_win.cpp
ClementKunz: same answer as above. And it looks like that the aarguments `pszNameString` is an output parameter.

Screenshot 2025-10-27 171024

src/libcommonserver/utility/digitalsignaturechecker_win.cpp
ClementKunz: `CertGetNameString` does not allocate memory for `szName`.

test/libcommonserver/utility/testdigitalsignaturecheck_win.cpp
ClementKunz: I would but I do not know how to do it. I do not want to make `_signatureInfo` a protected member just to make it accessible by the mock class.
Do you have any suggestions?

src/libcommonserver/utility/digitalsignaturechecker_win.cpp
luc-guyot-infomaniak: I see, thanks. But my comment is, I believe, still relevant: if the call does not modify `szName`, why deallocating this pointer's memory in case of error? It would have made sense if it had been passed as the fifth argument of the call.

src/libcommonserver/utility/digitalsignaturechecker_win.cpp
luc-guyot-infomaniak: As the call returns only the name size and does not allocate any memory to `szName`, my previous comment is still relevant : why deallocating a pointer that hasn't been allocated any memory?

src/libcommonserver/utility/digitalsignaturechecker_win.cpp
luc-guyot-infomaniak: Yes, you are right, as long as you provide the size as an argument of the call. But this point was not clear to me, because you inspected the value of `szName` in the earlier call as if it was an output argument.

test/libcommonserver/utility/testdigitalsignaturecheck_win.cpp
luc-guyot-infomaniak: We could write instead an extended test using a file with a valid signature that is stored on the remote test drive.

src/libcommonserver/utility/digitalsignaturechecker_win.cpp
ClementKunz: Done in 07d568c69e51f24da4a3d391c0b5d109b44bc304

src/libcommonserver/utility/digitalsignaturechecker_win.cpp
ClementKunz: Done in 6e5619d4c5986950e8ef44c0aec9c5f0797d83ef

src/libcommonserver/utility/digitalsignaturechecker_win.cpp
ClementKunz: Done in 6e5619d4c5986950e8ef44c0aec9c5f0797d83ef

src/libcommonserver/utility/digitalsignaturechecker_win.cpp
ClementKunz: Done in 6e5619d4c5986950e8ef44c0aec9c5f0797d83ef

src/libcommonserver/utility/digitalsignaturechecker_win.cpp
ClementKunz: Oh ok, so it is about the deallocation of memory on next line? About that you are 100% right and I have changed it

src/libcommonserver/utility/digitalsignaturechecker_win.cpp
ClementKunz: Done in https://github.com/Infomaniak/desktop-kDrive/commit/6e5619d4c5986950e8ef44c0aec9c5f0797d83ef

test/libcommonserver/utility/testdigitalsignaturecheck_win.cpp
ClementKunz: Done in edbe346eb88a8d6f7f12256897cffd528d451337

luc-guyot-infomaniak (approved):

PR #1136, opened 2025-10-27
Filter out empty node ID

ClementKunz:
Empty node IDs should never be sent in the `without_ids` body parameter


ChristopheLarchier (approved):

luc-guyot-infomaniak (commented):
Looks good. Would it be possible to add unit test checking this filtering?

ClementKunz:
Done in https://github.com/Infomaniak/desktop-kDrive/pull/1140

PR #1144, opened 2025-10-28
[TRELLO-1132] Define the constants of the Windows installer project at build time

luc-guyot-infomaniak:
The goal of this PR is to ensure that the file _"OverlayConstants.h"_ remains the only source of true regarding the values of the overlay GUI constants. Duplicate values have been removed from the Windows installer project. The overlay constants are now retrevied at the time of bootstrapping the Windows installer project.

### DIFF
--- a/installer/windows/Bootstraper/fetchVersion.ps1
+++ b/installer/windows/Bootstraper/fetchVersion.ps1
@@ -28,12 +28,15 @@ function Get-Overlay-Guid {
         [string] $keyType
     )
 
-    $guid = (Select-String -Path "$path\extensions\windows\standard\KDOverlays\OverlayConstants.h" $keyType |  Select-Object -ExpandProperty Line)
-    $guid = $guid.Split(" ")[-1] # Extract last word.
-    $guid = $guid.Substring(1) # Trim the initial 'L' character.
-
-    if ($LASTEXITCODE -ne 0) { exit $LASTEXITCODE }
-
+    try {
+        $guid = (Select-String -Path "$path\extensions\windows\standard\KDOverlays\OverlayConstants.h" $keyType |  Select-Object -ExpandProperty Line)
+        $guid = $guid.Split(" ")[-1] # Extract last word.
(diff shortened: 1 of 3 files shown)


installer/windows/Bootstraper/fetchVersion.ps1
```diff
@@ -21,6 +21,46 @@ param(
...
+    param (
+        [string] $path,
+        [string] $keyType
+    )
+
+    $guid = (Select-String -Path "$path\extensions\windows\standard\KDOverlays\OverlayConstants.h" $keyType |  Select-Object -ExpandProperty Line)
+    $guid = $guid.Split(" ")[-1] # Extract last word.
+    $guid = $guid.Substring(1) # Trim the initial 'L' character.
+
+    if ($LASTEXITCODE -ne 0) { exit $LASTEXITCODE }
```
herve-er: During testing, the script exited prematurely because `Select-String` doesn’t set `$LASTEXITCODE` — leaving it undefined (and thus not equal to 0), which triggered an unintended exit.

```suggestion
    try{
        $guid = (Select-String -Path "$path\extensions\windows\standard\KDOverlays\OverlayCnstants.h" $keyType |  Select-Object -ExpandProperty Line)
        $guid = $guid.Split(" ")[-1] # Extract last word.
        $guid = $guid.Substring(1) # Trim the initial 'L' character.
    }
    catch {
        Write-Error "Failed to retrieve overlay GUID for $keyType from OverlayCnstants.h"
        exit 1
    }
```

herve-er (changes requested):
Even with the suggested fix, building the project twice duplicates all keys, causing a build failure.

installer/windows/Bootstraper/fetchVersion.ps1
luc-guyot-infomaniak: Thanks for testing and spotting these problems (I am guilty of untested last minute changes).

Your concerns have been addressed with cd26f6a0ab330f374bddf143a9251ddc0d98ca4a.

herve-er (approved):

PR #1156, opened 2025-11-04
Handle 401 errors better

ClementKunz:
In some cases, we could try to refresh the token even if we do not have a refresh token available anymore.
This PR aims to bypass the refresh token process if the refresh token is empty. In that case, the request is retried once more.

### DIFF
--- a/src/libsyncengine/jobs/network/abstracttokennetworkjob.cpp
+++ b/src/libsyncengine/jobs/network/abstracttokennetworkjob.cpp
@@ -39,8 +39,8 @@ namespace KDC {
 std::unordered_map<int, std::pair<std::shared_ptr<Login>, int>> AbstractTokenNetworkJob::_userToApiKeyMap;
 std::unordered_map<int, std::pair<int, int>> AbstractTokenNetworkJob::_driveToApiKeyMap;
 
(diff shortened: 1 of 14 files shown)


src/libsyncengine/jobs/network/abstracttokennetworkjob.cpp
```diff
@@ -123,29 +122,26 @@ std::string AbstractTokenNetworkJob::getSpecificUrl() {
...
         _accessTokenAlreadyRefreshed = false;
-        _token = token;
-        addRawHeader("Authorization", "Bearer " + _token);
+        _apiToken = apiToken;
+        addRawHeader("Authorization", "Bearer " + _apiToken.accessToken());
         return ExitCode::TokenRefreshed;
     }
 
-    if (!_accessTokenAlreadyRefreshed || tokenUpdateDurationFromNow() > TOKEN_LIFETIME) {
+    if (!_apiToken.refreshToken().empty() && (!_accessTokenAlreadyRefreshed || tokenUpdateDurationFromNow() > TOKEN_LIFETIME)) {
```
ClementKunz, line 133: This is the fix

ClementKunz:
The reduction of code duplication would require a deep refactoring of the `AbstractTokenNetworkJob` class. This should be the topic of a dedicated PR.

ChristopheLarchier (approved):

luc-guyot-infomaniak (approved):

PR #1165, opened 2025-11-06
Crash serveur - commandSetThumbnail

ChristopheLarchier:
https://sentry-desktop.infomaniak.com/organizations/sentry/issues/216772/events/16d9412692874301b9b989c5f51b70cb/

Not sure this is the issue but access to _syncPalMap & _vfsMap should be mutexed as they are now used in jobs executed in threads.


src/server/appserver.h
```diff
@@ -53,6 +53,12 @@ class AppServer : public SharedTools::QtSingleApplication {
         Q_OBJECT
 
     public:
+        static SyncPalMap _syncPalMap;
```
luc-guyot-infomaniak: Names of public member variables should not be prefixed with `_`.

ClementKunz:
Could we address Sonar's remarks, at least on the newly created code such as in `ExtensionJob`

src/server/appserver.h
ChristopheLarchier: [127b6d7](https://github.com/Infomaniak/desktop-kDrive/pull/1165/commits/127b6d725df6572c2648705c4bf7ecbac02f0e13)

ChristopheLarchier:
> Could we address Sonar's remarks, at least on the newly created code such as in `ExtensionJob`

Some comments addressed.
[c0962f1](https://github.com/Infomaniak/desktop-kDrive/pull/1165/commits/c0962f1103b1221ed482ad574cd0ec66eb284229)

src/server/appserver.cpp
```diff
@@ -3824,17 +3949,21 @@ ExitInfo AppServer::setSupportsVirtualFiles(int syncDbId, bool value) {
...
             return exitInfo;
         }
-        _syncPalMap[sync.dbId()]->setVfs(vfsPtr);
+        syncPalMapIt->second->setVfs(vfs);
 
         QTimer::singleShot(100, this, [=, this]() {
             if (newMode != VirtualFileMode::Off) {
                 // Clear file system
-                _vfsMap[sync.dbId()]->convertDirContentToPlaceholder(SyncName2QStr(sync.localPath()), true);
+                const std::scoped_lock lock3(vfsMapMutex);
```
ClementKunz, line 3971: Scoped lock seems required at every access to the `vfsMap` and `syncPalMap`. This looks like something easily forgotten in a future development. Can we imagine a dedicated class managing those maps (or 2 classes, on for each)? This class would be responsible for the thread management, and it will have the positive side effect to decrease code duplication.

ClementKunz (approved):

PR #1177, opened 2025-11-07
Fix job manager logic for parallel download jobs

ClementKunz:
If too many syncs are running according to the number of CPU cores on the machine, the thread pool would be filled up with `LongPollJob` jobs. Since the download of a big file (> 100MB) start only if more than 50% of the job pool is available, they might never start.

This PR modify the logic to run max 3 big download jobs in parallel, without considering the number of CPU cores.


test/libsyncengine/jobs/testsyncjobmanagersingleton.cpp
```diff
@@ -394,9 +394,11 @@ void TestSyncJobManagerSingleton::testCanRunjob() {
                 break;
             }
             SyncJobManagerSingleton::instance()->queueAsyncJob(job, Poco::Thread::PRIO_NORMAL);
+            counter++;
             Utility::msleep(100);
         }
         CPPUNIT_ASSERT_EQUAL(true, noMoreRun);
+        CPPUNIT_ASSERT_EQUAL(static_cast<uint64_t>(3), counter);
```
luc-guyot-infomaniak:
```suggestion
        CPPUNIT_ASSERT_EQUAL(uint64_t{3}, counter);
```
Shorter.

luc-guyot-infomaniak (approved):

PR #1185, opened 2025-11-11
Node id exists on both replica

ClementKunz:
In the rare cases where a same node ID exist on both replica, the sequence of operations could be altered in a way that makes it impossible to propagate them. Especially, an operation on a child item could be executed before creating its parent folder, leading the sync to fail.

The correction involve checking that the operation list retrieve from a node ID contains only operation for a given replica side.

### DIFF
--- a/src/libsyncengine/reconciliation/syncoperation.h
+++ b/src/libsyncengine/reconciliation/syncoperation.h
@@ -158,7 +158,7 @@ class SyncOperationList : public SharedObject {
         std::unordered_map<UniqueId, SyncOpPtr> _allOps;
         std::list<UniqueId> _opSortedList;
         std::unordered_map<OperationType, std::unordered_set<UniqueId>> _opListByType;
-        std::unordered_map<NodeId, std::list<UniqueId>> _node2op;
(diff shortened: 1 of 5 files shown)


src/libsyncengine/reconciliation/syncoperation.h
```diff
@@ -133,7 +133,14 @@ class SyncOperationList : public SharedObject {
         SyncOpPtr getOp(UniqueId id);
         [[nodiscard]] const std::list<UniqueId> &opSortedList() const { return _opSortedList; }
         const std::unordered_set<UniqueId> &opListIdByType(const OperationType type) { return _opListByType[type]; }
-        const std::list<UniqueId> &getOpIdsFromNodeId(const NodeId &nodeId) { return _node2op[nodeId]; }
+        /**
+         * @brief Get the list of operation IDs related to the given node. The side must be provided as well in the case where the
+         * same ID would exist both on local and on remote replica.
```
herve-er:
```suggestion
         * @brief Get the list of operation IDs related to the given node. The side must be provided to distinguish between identical
         *  IDs on the local and remote replicas.
```

luc-guyot-infomaniak (dismissed):

herve-er (dismissed):

ChristopheLarchier (approved):

PR #1193, opened 2025-11-12
Crash server - Blacklist item during sync restart

ChristopheLarchier:
During a sync restart after a DataError, the localFileSystemObserverWorker continues to run and in case of AccessDenied, the handleAccessDeniedItem method is called which accesses the updateTree while it is destroyed/recreated by the sync restart.


src/libsyncengine/syncpal/syncpal.h
```diff
@@ -127,6 +127,8 @@ struct SyncProgress {
 
 class SYNCENGINE_EXPORT SyncPal : public std::enable_shared_from_this<SyncPal> {
     public:
+        static std::recursive_mutex updateTreesMutex;
```
ClementKunz: I am not comfortable with the idea of a mutex being accessible anywhere in the code. This seems to have high risk of deadlock and/or impacting performance.
Moreover, it seems to me that only the running module should be accessing and modifying the update trees. Therefor, no concurrent access should occur. Maybe the algorithm is to be refactored somewhere instead?

ChristopheLarchier: The LocalFileSystemObserverWorker updates the update trees and runs in parallel with the other modules:
LocalFileSystemObserverWorker::sendAccessDeniedError => SyncPal::handleAccessDeniedItem => UpdateTree::deleteNode

ChristopheLarchier: Changed as discussed.
https://github.com/Infomaniak/desktop-kDrive/pull/1193/commits/bff0b852b37970b97009daff14b72b27c65aa366

src/server/comm/extensionjob.cpp
```diff
@@ -196,6 +196,13 @@ void ExtensionJob::commandGetMenuItems(const CommString &argument, std::shared_p
                     }
                 }
 
+                {
+                    const std::scoped_lock lock(_dehydrationMutex);
```
ClementKunz: Variable name already used (see Sonar's comment)

test/libsyncengine/update_detection/update_detector/testupdatetreeworker.cpp
```diff
@@ -1199,16 +1200,18 @@ void TestUpdateTreeWorker::testIntegrityCheck() {
...
+        CPPUNIT_ASSERT(!_localUpdateTreeWorker->integrityCheck());
+        newNode->setChangeEvents(OperationType::Create | OperationType::Edit);
+        CPPUNIT_ASSERT(!_localUpdateTreeWorker->integrityCheck());
+        newNode->setChangeEvents(OperationType::Delete | OperationType::Move);
+        CPPUNIT_ASSERT(!_localUpdateTreeWorker->integrityCheck());
+        newNode->setChangeEvents(OperationType::Delete | OperationType::Edit);
+        CPPUNIT_ASSERT(!_localUpdateTreeWorker->integrityCheck());
+        newNode->setChangeEvents(OperationType::Delete | OperationType::Create);
+        CPPUNIT_ASSERT(!_localUpdateTreeWorker->integrityCheck());
+    }
```
ClementKunz: Why those checks are done only on CI?
Are they really pertinent? Can we completely remove them?

src/server/comm/extensionjob.cpp
ChristopheLarchier: [5248381](https://github.com/Infomaniak/desktop-kDrive/pull/1193/commits/524838172bf93b19c83c8f79d45c4a62c14b9e0f)

test/libsyncengine/update_detection/update_detector/testupdatetreeworker.cpp
ChristopheLarchier: [5248381](https://github.com/Infomaniak/desktop-kDrive/pull/1193/commits/524838172bf93b19c83c8f79d45c4a62c14b9e0f)

ClementKunz (approved):

PR #1197, opened 2025-11-13
[Windows] Crash serveur - Exception in the LocalFileSystemObserverWorker

ChristopheLarchier:
Possibly in GetFileStat or GetNodeId.

GetProcAddress can throw an exception if the first parameter, GetModuleHandle(L”ntdll.dll”), is null, which can happen in case of OS corruption or if the function is not found (ERROR_PROC_NOT_FOUND).

### DIFF
--- a/src/libcommonserver/io/iohelper_win.cpp
+++ b/src/libcommonserver/io/iohelper_win.cpp
@@ -170,26 +170,26 @@ bool IoHelper::getNodeId(const SyncPath &path, NodeId &nodeId) noexcept {
     LONGLONG fileInfo[(MAX_PATH_LENGTH_WIN_LONG + sizeof(FILE_ID_FULL_DIR_INFORMATION)) / sizeof(LONGLONG)];
     PFILE_ID_FULL_DIR_INFORMATION pFileInfo = (PFILE_ID_FULL_DIR_INFORMATION) fileInfo;
 
-    IoError ioError;
+    IoError ioError = IoError::Success;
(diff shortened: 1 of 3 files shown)


src/libcommonserver/io/iohelper_win.cpp
```diff
@@ -1162,4 +1159,33 @@ bool IoHelper::moveItemToTrash(const SyncPath &itemPath) {
...
+    if (hModule == nullptr) {
+        LOG_WARN(Log::instance()->getLogger(), "Error in GetModuleHandle for ntdll.dll");
+        ioError = dWordError2ioError(GetLastError(), Log::instance()->getLogger());
+        return nullptr;
+    }
+
+    try {
+        pzwQueryDirectoryFile = (PZW_QUERY_DIRECTORY_FILE) GetProcAddress(hModule, "ZwQueryDirectoryFile");
+    } catch (const std::exception &e) {
+        LOG_WARN(Log::instance()->getLogger(), "Exception in GetProcAddress: err=" << e.what());
```
ClementKunz, line 1178: Maybe force the value of `ioError` to something (ex: `Unknown`). If I understand the code correctly, `ioError` is still equal to `Success` here and the log message on line 176 might be misleading.

ChristopheLarchier: Done
[c71e523](https://github.com/Infomaniak/desktop-kDrive/pull/1197/commits/c71e5234797423e404f73f11f8e388ddb26d076a)

luc-guyot-infomaniak (approved):

PR #1200, opened 2025-11-13
Remove hardcoded value

ClementKunz:


test/libsyncengine/jobs/testsyncjobmanagersingleton.cpp
```diff
@@ -398,7 +398,7 @@ void TestSyncJobManagerSingleton::testCanRunjob() {
             Utility::msleep(100);
         }
         CPPUNIT_ASSERT_EQUAL(true, noMoreRun);
-        CPPUNIT_ASSERT_EQUAL(uint64_t{3}, counter);
+        CPPUNIT_ASSERT_EQUAL(uint64_t{maxNumberParallelBigDownloads}, counter);
```
ChristopheLarchier: Is it mandatory to keep the "uint64_t{}" ?

src/libsyncengine/jobs/network/networkjobsparams.h
```diff
@@ -29,6 +29,7 @@ static const uint64_t chunkMinSize = 10 * 1024 * 1024; // 10MB
 static const int64_t bigFileThreshold = 100 * 1024 * 1024; // if file size > 100MB -> start upload session
 static const uint64_t optimalTotalChunks = 200;
 static const uint64_t maxTotalChunks = 10000; // Theoretical max. file size 10'000 * 100MB = 1TB
+static const uint64_t maxNumberParallelBigDownloads = 3; // Download max. 3 big files in parallel
```
ChristopheLarchier: For numeric values, wouldn't it be better to use constexpr?

test/libsyncengine/jobs/testsyncjobmanagersingleton.cpp
ClementKunz: Indeed, it is not mandatory anymore. Removed in 5139ee209a734a7676b9745c049f36e5db88f528

src/libsyncengine/jobs/network/networkjobsparams.h
ClementKunz: Those values are not needed to be initialized at compile time, but I do not see any problem with making them `constexpr`. Done in 5139ee209a734a7676b9745c049f36e5db88f528.

ChristopheLarchier (approved):

PR #1201, opened 2025-11-14
feat: re-download the installer if absent

ClementKunz:
Check if the installer is still available in the temp folder before trying to start it. If not available, try to download it again.

### DIFF
--- a/src/server/updater/windowsupdater.cpp
+++ b/src/server/updater/windowsupdater.cpp
@@ -66,7 +66,7 @@ void WindowsUpdater::startInstaller() {
         return;
     }
     if (!std::filesystem::exists(filepath)) {
-        LOG_WARN(Log::instance()->getLogger(), "Installer file not found.");
+        LOGW_WARN(Log::instance()->getLogger(), L"Installer file not found at: " << Utility::formatSyncPath(filepath));
         if (_autoUpdate) {
             LOG_ERROR(Log::instance()->getLogger(), "Already tried to re-download the installer before.");
             setState(UpdateState::UpdateError);
@@ -132,7 +132,7 @@ void WindowsUpdater::downloadFinished(const UniqueId jobId) {
         return;
     }
     if (!std::filesystem::exists(filepath)) {
-        LOG_ERROR(Log::instance()->getLogger(), "Installer file not found.");
+        LOGW_WARN(Log::instance()->getLogger(), L"Installer file not found at: " << Utility::formatSyncPath(filepath));
         downloadUpdate();
         return;
     }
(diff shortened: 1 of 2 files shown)


src/server/updater/windowsupdater.cpp
```diff
@@ -119,10 +132,8 @@ void WindowsUpdater::downloadFinished(const UniqueId jobId) {
         return;
     }
     if (!std::filesystem::exists(filepath)) {
-        const auto error = "Installer file not found.";
-        sentry::Handler::captureMessage(sentry::Level::Warning, "WindowsUpdater::downloadFinished", error);
-        LOG_ERROR(Log::instance()->getLogger(), error);
-        setState(UpdateState::DownloadError);
+        LOG_ERROR(Log::instance()->getLogger(), "Installer file not found.");
```
luc-guyot-infomaniak: It would be nice to display the installer path in the log message.

src/server/updater/windowsupdater.cpp
```diff
@@ -65,6 +65,19 @@ void WindowsUpdater::startInstaller() {
         setState(UpdateState::DownloadError);
         return;
     }
+    if (!std::filesystem::exists(filepath)) {
```
luc-guyot-infomaniak: It would be safer to use one of the two signatures which are `noexcept`.

src/server/updater/windowsupdater.cpp
```diff
@@ -65,6 +65,19 @@ void WindowsUpdater::startInstaller() {
...
+        LOG_WARN(Log::instance()->getLogger(), "Installer file not found.");
```
ChristopheLarchier: Why "LOG_WARN", isn't it the normal case?

src/server/updater/windowsupdater.cpp
ClementKunz: Done in 5ea1727537041196ee1f15a71f2b7860a1686e45

src/server/updater/windowsupdater.cpp
ClementKunz: Done in 5ea1727537041196ee1f15a71f2b7860a1686e45

src/server/updater/windowsupdater.cpp
ClementKunz: At this point, we are trying to start the installer, it should be available. That's why I put a warning here, even though we might be able to re-download and start it without the user noticing anything.

herve-er (approved):

ChristopheLarchier (approved):

luc-guyot-infomaniak (approved):

PR #1206, opened 2025-11-20
feat(comm-server): Implements `SyncGetPrivateLinkUrl` as a GUI job

luc-guyot-infomaniak:


src/server/comm/guijobs/syncgetprivatelinkurljob.h
```diff
@@ -0,0 +1,45 @@
...
+        // Input parameters
+        int _driveDbId = 0;
+        CommString _fileId;
+
+        // Output parameters
+        std::string _linkUrl;
+
+        ExitInfo deserializeInputParms() override;
+        ExitInfo serializeOutputParms() override;
+        ExitInfo process() override;
```
ClementKunz, line 40: In most classes, methods are declared before attributes in header file. We should stay coherent everywhere.

ClementKunz: Ok apparently not in `guijobs`.
Forget about it

ClementKunz (approved):

PR #1226, opened 2025-11-25
Fix nightly builds

ClementKunz:
The `TestAppServer` is locked because of `QtSingleApplication`. Those tests have been disabled for now.


ClementKunz:
Fix implemented in https://github.com/Infomaniak/desktop-kDrive/pull/1243

PR #1228, opened 2025-11-25
feat(gui-jobs): Implements `NodeCreateMissingFolders` as a GUI job

luc-guyot-infomaniak:


src/server/comm/guijobs/nodecreatemissingfoldersjob.cpp
```diff
@@ -0,0 +1,140 @@
...
+    return ExitCode::Ok;
+}
+
+ExitInfo NodeCreateMissingFoldersJob::serializeOutputParms() {
+    writeParamValue(outParamsParentNodeId, _parentNodeId);
+
+    return ExitCode::Ok;
+}
+
+ExitInfo NodeCreateMissingFoldersJob::process() {
```
ClementKunz, line 92: Maybe we should handle such Sonar issues right now in order to avoid accumulating more issues

src/server/comm/guijobs/nodecreatemissingfoldersjob.cpp
```diff
@@ -0,0 +1,140 @@
...
+    // Pause all syncs of the drive
+    std::vector<int> pausedSyncs;
+
+    const std::scoped_lock lock(AppServer::syncPalMapMutex);
```
ClementKunz, line 96: Not to be handled in this PR, but I am still not a fan of having a public static mutex accessible everywhere in the code...

luc-guyot-infomaniak: One solution would be to implement ad hoc public methods in `AppServer` and to make a call to it in `process` here.

src/server/comm/guijobs/nodecreatemissingfoldersjob.cpp
luc-guyot-infomaniak: Addressed here: 382c274ab65d548645079508a938fa9dd50db35b

ClementKunz (approved):

PR #1237, opened 2025-11-27
Handle files finishing with a point or space on Windows

ClementKunz:
Some clients have files finishing by a '.' or a space on the local Windows file system, which is normally forbidden by the OS.
This creates a sync loop because the path could not be inserted in `_createFileOperationSet` in `UpdateTreeWorker`, leading to the reconstruction of the local snapshot and the restart of the sync.

The fix consists in rebuilding the snapshot in the case a of failed insertion in the map, but only if the path exists in it. If the path does not exist in the map, we simply continue the synchronization and the items ending with `.` or a space will be simply ignored because they will not appear in the `UpdateTree`.

It is hard to create a dedicated unit test because this is a situation that is difficult to reproduce locally.

### DIFF
--- a/src/libsyncengine/update_detection/update_detector/updatetreeworker.cpp
+++ b/src/libsyncengine/update_detection/update_detector/updatetreeworker.cpp
@@ -279,31 +279,38 @@ ExitCode UpdateTreeWorker::handleCreateOperationsWithSamePath() {
         }
 
         FSOpPtr createOp;
-        _operationSet->getOp(createOpId, createOp);
+        (void) _operationSet->getOp(createOpId, createOp);
 
-        std::pair<FSOpPtrMap::iterator, bool> insertionResult;
+        bool insertionFailedBecausePathExist = false;
+        SyncPath path;
         switch (createOp->objectType()) {
             case NodeType::File: {
-                SyncPath normalizedPath;
-                if (!Utility::normalizedSyncPath(createOp->path(), normalizedPath)) {
-                    normalizedPath = createOp->path();
+                if (!Utility::normalizedSyncPath(createOp->path(), path)) {
+                    path = createOp->path();
                     LOGW_SYNCPAL_WARN(_logger, L"Failed to normalize: " << Utility::formatSyncPath(createOp->path()));
                 }
-                insertionResult = _createFileOperationSet.try_emplace(normalizedPath, createOp);
+                if (!_createFileOperationSet.try_emplace(path, createOp).second) {
+                    insertionFailedBecausePathExist =
+                            _createFileOperationSet.contains(path); // The insertion might have fail because of unsupported
+                                                                    // character (e.g.: name finishing by a space or .)
+                }
                 break;
             }
             case NodeType::Directory:
-                insertionResult = createDirectoryOperationSet.try_emplace(createOp->path(), createOp);
+                if (!createDirectoryOperationSet.try_emplace(createOp->path(), createOp).second) {
+                    insertionFailedBecausePathExist =
+                            createDirectoryOperationSet.contains(path); // The insertion might have fail because of unsupported
+                                                                        // character (e.g.: name finishing by a space or .)
+                }
                 break;
             default:
                 break;
         }
 
-        if (!insertionResult.second) {
+        if (insertionFailedBecausePathExist) {
             // Failed to insert Create operation. A full rebuild of the snapshot is required.
             // The following issue has been identified: the operating system missed a delete operation, in which case a
             // liveSnapshot rebuild is both required and sufficient.
-
             LOGW_SYNCPAL_WARN(_logger, _side << L" update tree: Operation Create already exists on item with "
                                              << Utility::formatSyncPath(createOp->path()) << L", ID: "
                                              << CommonUtility::s2ws(createOp->nodeId()) << L", type " << createOp->objectType()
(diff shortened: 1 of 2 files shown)


src/libsyncengine/update_detection/update_detector/updatetreeworker.cpp
```diff
@@ -279,31 +279,38 @@ ExitCode UpdateTreeWorker::handleCreateOperationsWithSamePath() {
         }
 
         FSOpPtr createOp;
-        _operationSet->getOp(createOpId, createOp);
+        (void) _operationSet->getOp(createOpId, createOp);
 
-        std::pair<FSOpPtrMap::iterator, bool> insertionResult;
+        bool insertionFailedBecausePathExist = false;
```
luc-guyot-infomaniak:
```suggestion
        bool insertionFailedBecausePathExists = false;
```

src/libsyncengine/update_detection/update_detector/updatetreeworker.cpp
```diff
@@ -279,31 +279,38 @@ ExitCode UpdateTreeWorker::handleCreateOperationsWithSamePath() {
...
-                if (!Utility::normalizedSyncPath(createOp->path(), normalizedPath)) {
-                    normalizedPath = createOp->path();
+                if (!Utility::normalizedSyncPath(createOp->path(), path)) {
+                    path = createOp->path();
                     LOGW_SYNCPAL_WARN(_logger, L"Failed to normalize: " << Utility::formatSyncPath(createOp->path()));
                 }
-                insertionResult = _createFileOperationSet.try_emplace(normalizedPath, createOp);
+                if (!_createFileOperationSet.try_emplace(path, createOp).second) {
+                    insertionFailedBecausePathExist =
+                            _createFileOperationSet.contains(path); // The insertion might have fail because of unsupported
```
luc-guyot-infomaniak:
```suggestion
                            _createFileOperationSet.contains(path); // The insertion might have failed because of unsupported
```

src/libsyncengine/update_detection/update_detector/updatetreeworker.cpp
```diff
@@ -279,31 +279,38 @@ ExitCode UpdateTreeWorker::handleCreateOperationsWithSamePath() {
...
+                                                                    // character (e.g.: name finishing by a space or .)
```
luc-guyot-infomaniak:
```suggestion
                                                                    // characters in `path` (e.g.: a file name finishing by a space or a dot).
```

src/libsyncengine/update_detection/update_detector/updatetreeworker.cpp
```diff
@@ -279,31 +279,38 @@ ExitCode UpdateTreeWorker::handleCreateOperationsWithSamePath() {
...
+                }
                 break;
             }
             case NodeType::Directory:
-                insertionResult = createDirectoryOperationSet.try_emplace(createOp->path(), createOp);
+                if (!createDirectoryOperationSet.try_emplace(createOp->path(), createOp).second) {
+                    insertionFailedBecausePathExist =
+                            createDirectoryOperationSet.contains(path); // The insertion might have fail because of unsupported
```
luc-guyot-infomaniak:
```suggestion
                            createDirectoryOperationSet.contains(path); // The insertion might have failed because of unsupported
```

src/libsyncengine/update_detection/update_detector/updatetreeworker.cpp
```diff
@@ -279,31 +279,38 @@ ExitCode UpdateTreeWorker::handleCreateOperationsWithSamePath() {
...
+                                                                        // character (e.g.: name finishing by a space or .)
```
luc-guyot-infomaniak:
```suggestion
                                                                        // characters (e.g.: a folder name finishing by a space or a dot).
```

src/libsyncengine/update_detection/update_detector/updatetreeworker.cpp
```diff
@@ -279,31 +279,38 @@ ExitCode UpdateTreeWorker::handleCreateOperationsWithSamePath() {
...
-                SyncPath normalizedPath;
-                if (!Utility::normalizedSyncPath(createOp->path(), normalizedPath)) {
-                    normalizedPath = createOp->path();
+                if (!Utility::normalizedSyncPath(createOp->path(), path)) {
+                    path = createOp->path();
                     LOGW_SYNCPAL_WARN(_logger, L"Failed to normalize: " << Utility::formatSyncPath(createOp->path()));
                 }
-                insertionResult = _createFileOperationSet.try_emplace(normalizedPath, createOp);
+                if (!_createFileOperationSet.try_emplace(path, createOp).second) {
+                    insertionFailedBecausePathExist =
```
ChristopheLarchier: To avoid code duplication, would it be possible to keep the previous code and check in  "if (!insertionResult.second)..." that "_createFileOperationSet.contains(path)"?

ClementKunz: I wanted to do it but since it is not the same map that is checked (`_createFileOperationSet` vs `createDirectoryOperationSet`), we would be forced to check again if the object type is file or a directory which would result into more line of code than that.

ChristopheLarchier (approved):

PR #1240, opened 2025-11-27
feat(gui-jobs): implements `ExclTemplGetExcluded` and the other exclusion template related tasks as a GUI jobs

luc-guyot-infomaniak:


src/libsyncengine/propagation/executor/executorworker.cpp
```diff
@@ -286,7 +286,8 @@ ExitInfo ExecutorWorker::handleCreateOp(SyncOpPtr syncOp, std::shared_ptr<SyncJo
     if (isLiteSyncActivated() && !syncOp->omit()) {
         bool isDehydratedPlaceholder = false;
         if (ExitInfo exitInfo = checkLiteSyncInfoForCreate(syncOp, absoluteLocalFilePath, isDehydratedPlaceholder); !exitInfo) {
-            LOG_SYNCPAL_WARN(_logger, "Error in checkLiteSyncInfoForCreate" << " " << exitInfo);
+            LOG_SYNCPAL_WARN(_logger, "Error in checkLiteSyncInfoForCreate"
+                                              << " " << exitInfo);
```
ClementKunz, line 290: Strange that this line is re-formatted while the one above is longer

ClementKunz:
There are a few of Sonar's issues that could be resolved easily, especially those about unused return values and catching more specific exceptions.

src/libsyncengine/propagation/executor/executorworker.cpp
luc-guyot-infomaniak: The formatting script is doing this (I have checked out the `origin/develop` version of this file and then commit it, the result is the same).

luc-guyot-infomaniak:
> There are a few of Sonar's issues that could be resolved easily, especially those about unused return values and catching more specific exceptions.

Done with 698c0ac4f4e47fd3ba32aac27eb4987b123e3eb8.

ClementKunz (approved):

PR #1220, opened 2025-11-24
 feat: XPC GUI comm jobs 6

adrien-coye:
### Abstract

This PR aims to develop the `CoherentCache` to parity with all the jobs added with #1198
(full tested observation will be available in XPC-7)

### Note

Followed by https://github.com/Infomaniak/desktop-kDrive/pull/1247
Depends on https://github.com/Infomaniak/desktop-kDrive/pull/1198

PR #1247, opened 2025-12-01
feat: XPC GUI comm jobs 7

adrien-coye:
### Abstract

This PR aims to finish observation of the server cache.
Observation is now also fully tested

Note that now observation can send nil value if the object is removed.
You can guard on your observation to make sure it will not nilify your local values.

### Note

Depends on https://github.com/Infomaniak/desktop-kDrive/pull/1220


adrien-coye (commented):
We discussed this with @valentinperignon

valentinperignon (approved):

PR #1253, opened 2025-12-03
Fix lite sync ext may start after app

ChristopheLarchier:


src/libcommonserver/utility/utility_mac.mm
```diff
@@ -128,6 +128,14 @@ void runKillCommand(pid_t pid) {
     }
 }
 
+bool Utility::isLiteSyncExtRunning() {
+    NSString *bundleID = NSBundle.mainBundle.bundleIdentifier;
+    NSString *processName = [NSString stringWithFormat:@"%@.LiteSyncExt", bundleID];
+    NSMutableArray *pids = getPIDsForProcessName(processName);
+    assert(pids.count <= 1);
```
ClementKunz, line 135: Is this assert really necessary? Nothing forbid to call this utility function when a sync is running without LiteSync, which result in a crash in debug mode.
It should be the caller's responsibility to assert if the result is not the one expected.

ChristopheLarchier: When there is no sync running with LiteSync, pids.count == 0 so <= 1

ClementKunz (approved):

PR #1259, opened 2025-12-04
WinUI3 -  Add WinUI3 client to the release process

herve-er:
**Depends-On: #1238** 

This pull request adds full support for building and packaging the new Windows GUI client ("gui4") for kDrive, while also improving the build scripts and installer workflow. The new client is now built, archived, signed, and installed correctly, and protocol registration is automatically handled at launch.
The build system has been refactored to better manage output directories and streamline installer and package creation.

As part of this PR, the main CMake client target now builds both the legacy client and the new WinUI3 client.
The WinUI3 client output is located in **`bin/client`**.

### Developer Requirements

Every developer must:

* Update their Conan profile at
  **`%USERPROFILE%/.conan2/profiles/debug_vars.cmake`**
  to include:

  ```cmake
  set(CMAKE_BUILD_TYPE "Debug")
  ```

  👉 Visual Studio does not set `CMAKE_BUILD_TYPE` by default when building CMake projects, which previously caused various build inconsistencies.

* Ensure that **Visual Studio 2022 or 2025** is installed with the following workloads/packages:

  * **.NET Desktop Development**
  * **WinUI app development**
  * **.NET 9.0 Runtime**
---

# Installer Build

Several improvements have been made to the `build-drive.ps1` script.

* The script **now always produces an installer**, even when the `-upload` flag is not provided.
  👉 This allows to easily generate *debug installers*.

* A new `-newGUI` flag has been added.
  When enabled, it bundles the new GUI (alongside with the legacy one) into the installer and signs it.
  ⚠️ This flag is temporary and will be removed once the new GUI is officially released.
  (Note: the release build script does **not** make use of this flag yet.)

---

# Using the New GUI

For now, the new GUI is launched automatically when the following conditions are met:

* It is available at runtime
  (either because it was installed via an installer built with `-newGUI`, or because it is built from the IDE).
* The server triggers the startup of a client  (either a Release build, or `KDRIVE_DEBUG_RUN_CLIENT` is set).
* The application is in the **Internal Release Channel**  (the change takes effect after the first app restart once the user selects the channel).

When the new GUI is active, the legacy interface can still be accessed at any time by clicking the **kDrive logo in the top-left corner** of the new GUI.


infomaniak-build-tools/windows/build-drive.ps1
```diff
@@ -521,6 +526,23 @@ function Prepare-Archive {
...
+    if ($newGui) {
+        # Copy client files
+        Write-Host "Copying new client files ($newGuiDir) to the archive ..."
+        Remove-Item -Path "$archivePath/client" -Recurse -Force -ErrorAction SilentlyContinue
+        Copy-Item -Path "$newGuiDir/." -Destination "$archivePath/client" -Recurse
+
+        # Sign all the .exe, .dll and .xbf that have no signature yet
+        $filesToSign = Get-ChildItem -Path "$archivePath/client" -Recurse -Include *.exe, *.dll, *.xbf | Where-Object {
+            $signature = Get-AuthenticodeSignature $_.FullName
+            $signature.Status -eq 'NotSigned'
```
ClementKunz, line 538: What is this test about? We do not keep or check the result of this test?

src/gui4/windows/kDrive client/kDrive client/App.xaml.cs
```diff
@@ -103,6 +109,28 @@ protected override async void OnLaunched(Microsoft.UI.Xaml.LaunchActivatedEventA
 
         }
 
+        private void RegisterProtocol()
```
ClementKunz:
```suggestion
        private void RegisterOAuthProtocol()
```

infomaniak-build-tools/windows/build-drive.ps1
herve-er: It is inside the `Where-Object { ... }`. All files that do not pass this test are not included in the `$filesToSign` list. The goal is to avoid signing DLLs from the Windows SDK that are already signed by Microsoft.

ClementKunz: Sorry, I did not see the bracket 🙃

src/gui4/windows/kDrive client/kDrive client/App.xaml.cs
herve-er: Done here: https://github.com/Infomaniak/desktop-kDrive/pull/1259/commits/705650711d90d7cd6c81b997293d372c5d7a7701

ClementKunz (approved):

PR #1260, opened 2025-12-04
feat(gui-jobs): implements `UpdaterChangeChannel` and other updater tasks as GUI jobs

luc-guyot-infomaniak:


src/server/appserver.cpp
```diff
@@ -818,6 +818,40 @@ bool AppServer::areMacVfsAuthsOk() const {
...
+    if (_noUpdate) return UpdateState::Unknown;
+
+    assert(_updateManager && "The update manager is not set.");
+
+    return _updateManager->state();
+}
+
+void AppServer::startInstaller() {
+    assert(_updateManager && "The update manager is not set.");
+    _updateManager->startInstaller();
```
herve-er: To avoid a crash ->
```suggestion
    LOG_IF_FAIL(_logger, _updateManager && "The update manager is not set.");
    if(_updateManager) _updateManager->startInstaller();
```

src/server/appserver.cpp
```diff
@@ -818,6 +818,40 @@ bool AppServer::areMacVfsAuthsOk() const {
...
+        return versionInfo;
+    }
+
+    assert(_updateManager && "The update manager is not set.");
+
+    return _updateManager->versionInfo(versionChannel);
+}
+
+UpdateState AppServer::getUpdateState() const {
+    if (_noUpdate) return UpdateState::Unknown;
```
ClementKunz: Maybe we should add a "Deactivated" state. Waht do you think?

src/server/comm/guijobs/updaterchangechanneljob.cpp
herve-er, line 1: I think this job is redundant with UpdateParmsInfo, which also sets a channel field. I’d be in favor of a solution that avoids having to call two separate paths to achieve the same result, this also reduces the risk of inconsistency between ParmsCache and the updater.

luc-guyot-infomaniak: Agreed. If it is feasible (I feel it totally is), I'll remove the job and update the  `UpdateParmsInfoJob` to make sure it encompasses the update of the channel value.

src/server/appserver.cpp
herve-er: -> https://github.com/Infomaniak/desktop-kDrive/pull/1260/changes/6b01dbe09e20917e0f47904fb79b7d2a622043f1

src/server/appserver.cpp
luc-guyot-infomaniak: Addressed with 9e83ddae7d1591aae3f95f971e67671cb1c3ad5b.

src/server/comm/guijobs/updaterchangechanneljob.cpp
luc-guyot-infomaniak: Done here: 42527667f9aa063ed4c92676c2da4ceb9eafe4cc

ClementKunz (approved):

PR #1268, opened 2025-12-08
fix(client-logging): do not log any message in `appclient.cpp` before logging setup is complete

luc-guyot-infomaniak:


src/gui/appclient.cpp
```diff
@@ -112,6 +112,8 @@ AppClient::AppClient(int &argc, char **argv) :
         return;
     }
 
+    setupLogging();
```
herve-er, line 115: `setupLogging();` will probably crash if called before `connectToServer()`.  
`setupLogging();` requires the `ParametersCache`, which depends on the server connection<  
If the serverconnection is not set up, `ParametersCache::instance()` will return `nullptr` and cause a crash.

luc-guyot-infomaniak: Thanks, you are absolutely right! Actually the Qt debug macros redirect the log messages to the standard output as long as the logger is not set up. So it does seem useful to change anything here.

luc-guyot-infomaniak:
Abandoned as it is not only useless but also harmful.

PR #1270, opened 2025-12-08
Cherry pick fixes about network disconnection

ClementKunz:
Backport fixes from PR #1267 and #1265

### DIFF
--- a/src/libsyncengine/jobs/network/kDrive_API/downloadjob.cpp
+++ b/src/libsyncengine/jobs/network/kDrive_API/downloadjob.cpp
@@ -44,6 +44,7 @@ namespace KDC {
 #define TRIALS 5
 #define READ_PAUSE_SLEEP_PERIOD 100 // 0.1 s
 #define READ_RETRIES 10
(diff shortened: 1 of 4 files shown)


herve-er (dismissed):

src/libsyncengine/jobs/network/kDrive_API/downloadjob.cpp
```diff
@@ -649,12 +649,28 @@ ExitInfo DownloadJob::createTmpFile(std::optional<std::reference_wrapper<std::is
                         retryCount = 0;
                     }
 
+                    if (readSize == 0 && !istr->get().eof()) {
+                        if (retryCount < READ_RETRIES) {
```
herve-er: Could we increase the max retries value?
Currently, we abort after one second without receiving data, which is very likely to happen on an unstable network.

ClementKunz: Done in 6792341817707273d4e1ce53178b5580a6b161f3

herve-er (approved):

PR #1276, opened 2025-12-09
fix(parmsdb): bind filter param only when needed

herve-er:
Previously, the filter parameter was always bound to the SQL query, even when no filter was applied. Now, the filter value is only bound if it is different from "None", preventing LOG_IF_FAIL false positive

### DIFF
--- a/src/libparms/db/parmsdb.cpp
+++ b/src/libparms/db/parmsdb.cpp
@@ -2960,7 +2960,9 @@ bool ParmsDb::selectConflicts(int syncDbId, ConflictType filter, std::vector<Err
 
     LOG_IF_FAIL(queryResetAndClearBindings(requestId));
     LOG_IF_FAIL(queryBindValue(requestId, 1, syncDbId));
-    if (filter != ConflictType::None) LOG_IF_FAIL(queryBindValue(requestId, 2, std::to_string(toInt(filter))));
+    if (filter != ConflictType::None) {
+        LOG_IF_FAIL(queryBindValue(requestId, 2, std::to_string(toInt(filter))));
+    }
 
     bool found = false;
     for (;;) {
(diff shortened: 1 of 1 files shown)


src/libparms/db/parmsdb.cpp
```diff
@@ -2960,7 +2960,9 @@ bool ParmsDb::selectConflicts(int syncDbId, ConflictType filter, std::vector<Err
 
     LOG_IF_FAIL(queryResetAndClearBindings(requestId));
     LOG_IF_FAIL(queryBindValue(requestId, 1, syncDbId));
-    LOG_IF_FAIL(queryBindValue(requestId, 2, std::to_string(toInt(filter))));
```
ClementKunz, line 2963: Sorry but I do not understand why it triggers `LOG_IF_FAIL`, can you explain?
We convert `ConflictType::None` (enum value) -> 0 (int) -> "0" (string). The bind fail if we supply "0"?

herve-er: Line 2958 we selected which db request we want to use (`SELECT_ALL_CONFLICTS_BY_SYNCDBID_REQUEST_ID` or `SELECT_FILTERED_CONFLICTS_BY_SYNCDBID_REQUEST_ID`).

If `ConflictType` is `ConflictType::None`, we chose `SELECT_ALL_CONFLICTS_BY_SYNCDBID_REQUEST_ID` which is:

```cpp
#define SELECT_ALL_CONFLICTS_BY_SYNCDBID_REQUEST                                                                            \
    "SELECT dbId, time, "                                                                                                   \
    "functionName, workerName, exitCode, exitCause, "                                                                       \
    "localNodeId, remoteNodeId, nodeType, path, status, conflictType, inconsistencyType, cancelType, destinationPath FROM " \
    "error "                                                                                                                \
    "WHERE syncDbId=?1 AND conflictType <> 0;"
```

We can see that this request only expects one parameter.

But later (line 2963) we are passing a second parameter anyway.
`queryBindValue` fails to **bind** the provided value with the request (wich expect only one) and returns false.

ClementKunz: Of course thanks for the explaination

ClementKunz (approved):

PR #1302, opened 2025-12-12
feat(gui-jobs): implements `SignalUtilityQuit` and other utility jobs as a GUI jobs

luc-guyot-infomaniak:


src/server/comm/guijobs/signalutilityshownotificationjob.cpp
```diff
@@ -0,0 +1,42 @@
...
+
+// Output parameters keys
+static const auto outParamsTitle = "title";
+static const auto outParamsMessage = "message";
+
+namespace KDC {
+
+SignalUtilityShowNotificationJob::SignalUtilityShowNotificationJob(CommString title, CommString message) :
+    _title(std::move(title)),
+    _message(std::move(message)) {
```
ClementKunz, line 31: All other jobs take the parameters as const ref. Is there really a gain to pass by value and use `std::move` here?

src/server/appserver.cpp
```diff
@@ -75,6 +75,33 @@
...
+#include "server/comm/guijobs/signalsyncprogressinfo.h"
+#include "server/comm/guijobs/signalsynccompleteditem.h"
+#include "server/comm/guijobs/signalsyncupdatedjob.h"
+#include "server/comm/guijobs/signalerroraddedjob.h"
+#include "server/comm/guijobs/signalerrorremovedjob.h"
+#include "server/comm/guijobs/signalutilityshownotificationjob.h"
+#include "server/comm/guijobs/signalutilityshowsettingsjob.h"
+#include "server/comm/guijobs/signalutilityshowsynthesisjob.h"
+#include "server/comm/guijobs/signalutilityloguploadstatejob.h"
+#include "server/comm/guijobs/signalutilityquitjob.h"
```
ClementKunz: This looks like a lot of include for only 5 new classes. Are sure they are all mandatory?
The app seems to compile (on Windows) without most them.

src/server/comm/guijobs/signalutilityshownotificationjob.cpp
luc-guyot-infomaniak: This was suggested by SonarCloud.

src/server/appserver.cpp
luc-guyot-infomaniak: Thanks, that's an accidental change committed mistakenly while solving merge conflicts. Addressed with 5e48f1728296ee5137dbc682e4eb2deddf3a2ace.

ClementKunz (approved):

PR #1303, opened 2025-12-12
feat(socket): add async reconnect loop and splash screen

herve-er:
Refactor socket connection logic to use an asynchronous reconnect loop, improving resilience to connection loss. The client now continuously attempts to reconnect if the server is unavailable or the connection drops, with a 2s delay between attempts. SendRequestAsync now waits for a connection before sending requests. Also includes code cleanup, better resource management with Dispose, and adds a destructor to ensure proper cleanup. 

## New Splash screen
(This is a demo, it is faster in real case usage)

https://github.com/user-attachments/assets/57644c78-82fd-421a-81e7-bbd6bf59fb99


src/gui4/windows/kDrive client/kDrive client/ServerCommunication/Services/SocketServerCommProtocol.cs
```diff
@@ -65,72 +65,85 @@ public SocketServerCommProtocol()
...
-            if (_client != null && _client.Connected)
+        ~SocketServerCommProtocol()
+        {
+            _client?.Dispose();
+            _stopRequested = true;
+        }
+
+        private async Task ReconnectLoop()
+        {
+            string homePath = Path.Combine(
```
ClementKunz: Maybe a more explicit name will nice. `commPortFilePath` or something lik that?

herve-er: https://github.com/Infomaniak/desktop-kDrive/pull/1303/commits/709c95fd92570eca3560b111e58e37bf8763d85d

ClementKunz (approved):

PR #1311, opened 2025-12-15
WinUI3 add special errors pages

herve-er:
https://github.com/user-attachments/assets/afe268af-138e-4a55-95f7-04dd0d6afd1d


src/gui4/windows/kDrive client/kDrive client/Pages/SpecialErrorsPages/SpecialErroBasePage.cs
```diff
@@ -0,0 +1,87 @@
...
+            DetachHandlers();
+        }
+
+        protected void DetachHandlers()
+        {
+            ViewModel.SelectedSyncChanged -= OnSelectedSyncChanged;
+
+            if (ViewModel.SelectedSync is not null)
+                ViewModel.SelectedSync.PropertyChanged -= OnSelectedSyncPropertyChanged;
+        }
```
ClementKunz, line 59: Duplicated method. Is it possible to implement it in a parent class?

herve-er: I have to add `DetachHandlers` to all classes that subscribe to events to avoid memory leaks.
At the moment, both pages subscribe to the same events, which makes the function identical, but it is highly likely that they will quickly diverge and therefore require different detach handler methods (and it is certain that when I add the `DetachHandlers` method to other classes, they will have different implementations).
I’m available to discuss this in person if needed

ClementKunz (approved):

PR #1316, opened 2025-12-16
Implement exists request

ClementKunz:
A lighter request exists in the kDrive API to check file existance. This PR aims at using it to check file existance in `LocalDeleteJob` and fallback to the existing method if the file still exist on remote replica.

Depends on #1333


src/libsyncengine/jobs/network/kDrive_API/fileexistsjob.cpp
```diff
@@ -0,0 +1,99 @@
...
+    for (const auto &[id, exists]: _itemInfo) {
+        (void) ids.add(id);
+    }
+    (void) obj.set("ids", ids);
+
+    std::stringstream ss;
+    try {
+        obj.stringify(ss);
+    } catch (Poco::Exception &e) {
+        // Not an issue
```
herve-er:
```suggestion
```

src/libsyncengine/jobs/local/localdeletejob.cpp
```diff
@@ -126,30 +124,31 @@ ExitInfo LocalDeleteJob::canRun() {
     }
 
     // Check if the item we want to delete locally has a remote counterpart.
-
-    SyncPath remoteRelativePath;
-    const bool remoteItemIsFound = findRemoteItem(remoteRelativePath);
-    if (!remoteItemIsFound) return ExitCode::Ok; // Safe deletion.
+    FileExistsJob existsJob(_syncInfo.driveDbId, {_remoteNodeId});
+    existsJob.runSynchronously();
+    if (!existsJob.exists(_remoteNodeId)) return ExitCode::Ok; // Safe deletion.
```
herve-er: It seems that `FileExistsJob::exists` currently returns a `FileExistsJob::Status`:

```cpp
enum Status {
    EXISTS = 0,
    NOT_FOUND,
    UNHANDLED
};
```

If the item is found, it returns `FileExistsJob::Status::EXISTS`, which equals 0. Consequently, using it in a condition like `!existsJob.exists(_remoteNodeId)` will evaluate to `true`, causing an early exit. This appears to be the opposite of the intended behavior.

To simplify the logic and make it consistent with other parts of the code, we could consider changing the signature of `exists` to either:

```cpp
ExitInfo FileExistsJob::exists(bool &exist);
```

or simply:

```cpp
ExitInfo FileExistsJob::exists();
```

where it would return:

* `ExitCode::OK` if the item exists,
* `ExitCode::NotFound` if it doesn’t,
* `ExitCode::LogicError` if the ID isn’t managed by the job.

This approach would make the intent clearer and avoid the counterintuitive boolean logic.

src/libsyncengine/jobs/network/kDrive_API/fileexistsjob.h
```diff
@@ -18,28 +18,28 @@
 
 #pragma once
 
-#include "server/comm/guijobs/abstractguijob.h"
+#include "jobs/network/abstracttokennetworkjob.h"
 
 namespace KDC {
 
-class NodePathJob : public AbstractGuiJob {
+class FileExistsJob : public AbstractTokenNetworkJob {
```
herve-er: Suggestion:
To make the intent of the job clearer, you could consider renaming `FileExistsJob` to `ItemsExistJob`. This would emphasize that it can check multiple items and that these items could be either files or directories, rather than only files.

src/libsyncengine/jobs/network/kDrive_API/fileexistsjob.cpp
ClementKunz: Done in d1748c02c3e55d65ae97de3b7998f4d1a51de86a

src/libsyncengine/jobs/local/localdeletejob.cpp
ClementKunz: Indeed, silly mistake. Your solution is indeed better, and avoid adding this new enum type. Done in d1748c02c3e55d65ae97de3b7998f4d1a51de86a

src/libsyncengine/jobs/network/kDrive_API/fileexistsjob.h
ClementKunz: Done in d1748c02c3e55d65ae97de3b7998f4d1a51de86a

src/libsyncengine/jobs/network/kDrive_API/itemsexistjob.h
```diff
@@ -0,0 +1,40 @@
...
+
+        ExitInfo exists(const NodeId &id);
+
+    private:
+        ExitInfo setData() override;
+        virtual std::string getSpecificUrl() override;
+
+        ExitInfo handleResponse(std::istream &is) override;
+
+        std::unordered_map<NodeId, bool, StringHashFunction, std::equal_to<>> _itemInfo;
```
luc-guyot-infomaniak: You may use `NodeSet` from `types.h` instead.

ClementKunz: No because I need a map here (key: NodeId, value: boolean), not a set

luc-guyot-infomaniak: An unordered map with boolean values implements an unordered set.

ClementKunz: `_itemInfo` is used both as input and output. In the constructor, we fill `_itemInfo` with all the IDs provided by the caller. Then, the boolean value is updated in the map according to the result received from the backend.
I choose this approach so, when calling the `exists` method, we can differentiate the values that have been sent to the backend (all the values stored in `_itemInfo`) and really do not exist from the ones that were never sent. The unit tests explicitly show this behavior.
However, perhaps it is overcomplicated, and just a set would be enough. In that case, `exists` would return false for both items that really do not exist from the ones that have not been checked.

src/libsyncengine/jobs/network/kDrive_API/itemsexistjob.cpp
```diff
@@ -0,0 +1,99 @@
...
+    _httpMethod = Poco::Net::HTTPRequest::HTTP_POST;
+    for (const auto &id: ids) {
+        _itemInfo.emplace(id, false);
+    }
+}
+
+ExitInfo ItemsExistJob::exists(const NodeId &id) {
+    if (!_itemInfo.contains(id)) return ExitCode::LogicError;
+    if (_itemInfo[id]) return ExitCode::Ok;
+    return {ExitCode::SystemError, ExitCause::NotFound};
```
herve-er: I cannot find a better exit code, so it is fine for me. However, SystemError is generally tied to the local system.

herve-er (dismissed):

src/libsyncengine/jobs/network/kDrive_API/itemsexistjob.cpp
herve-er: (You can self resolve this issue 👍🏾 )

ClementKunz: I have changed `ExitCode::SystemError` for `ExitCode::BackError`. Not ideal either, but at least it reflects that the item is absent from the server, not the local machine.
Done in 8375120d33d29dc4e0727259f0385f0d1bdfa733

src/libsyncengine/jobs/network/abstractnetworkjob.cpp
```diff
@@ -643,12 +643,6 @@ ExitInfo AbstractNetworkJob::extractJson(const std::string &replyBody, Poco::JSO
         Utility::logGenericServerError(_logger, "Request error", replyBody, httpResponse());
         return {ExitCode::BackError, ExitCause::ApiErr};
     }
-
```
ChristopheLarchier, line 646: Rather remove the logging of the reply in AbstractTokenNetworkJob::handleResponse.

ClementKunz: Done in 103b75cf9a9116effd80f1685b93386b481cfa57

src/libsyncengine/jobs/network/kDrive_API/itemsexistjob.cpp
ChristopheLarchier: Usually, Exitinfo allows you to know if the function encountered an execution problem.
You should add a bool& parameter to return the result.

src/libsyncengine/jobs/network/kDrive_API/itemsexistjob.h
```diff
@@ -0,0 +1,40 @@
...
+
+        ExitInfo exists(const NodeId &id);
+
+    private:
+        ExitInfo setData() override;
+        std::string getSpecificUrl() override;
+
+        ExitInfo handleResponse(std::istream &is) override;
+
+        std::unordered_map<NodeId, bool, StringHashFunction, std::equal_to<>> _itemInfo;
```
ChristopheLarchier: The variable name doesn't seem very clear to me.
_itemInfo => _nodeExistenceMap ?

ClementKunz: Ok for me : 8b2f2696309c5040812584d4da94e466e3806c42

ChristopheLarchier (approved):

PR #1326, opened 2025-12-18
fix: fix path for icon resource

ClementKunz:
Fix the path for icon resource (`Ordinateur\HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows\CurrentVersion\Explorer\SyncRootManager\<sync root ID>`) on every vfs start

### DIFF
--- a/extensions/windows/cfapi/Vfs/cloudproviderregistrar.cpp
+++ b/extensions/windows/cfapi/Vfs/cloudproviderregistrar.cpp
@@ -35,6 +35,7 @@ using namespace Windows::Security::Cryptography;
 
 #define REGKEY_NAMESPACECLSID L"NamespaceCLSID"
 #define REGKEY_AUMID L"AUMID"
+#define REGKEY_ICONRESOURCE L"IconResource"
 
 std::wstring CloudProviderRegistrar::registerWithShell(ProviderInfo *providerInfo, wchar_t *namespaceCLSID,
                                                        DWORD *namespaceCLSIDSize) {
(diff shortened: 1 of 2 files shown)


extensions/windows/cfapi/Vfs/cloudproviderregistrar.cpp
```diff
@@ -87,6 +88,22 @@ std::wstring CloudProviderRegistrar::registerWithShell(ProviderInfo *providerInf
                     TRACE_ERROR(L"Could not set registry value %s=%s", name.c_str(), value.c_str());
                 }
 
+                // Update IconResource
+                name = REGKEY_ICONRESOURCE;
```
luc-guyot-infomaniak, line 95: Suggestion: To avoid redundancy, create a helper function that takes `name` and `value` as arguments and modifies the registry accordingly.

extensions/windows/cfapi/Vfs/cloudproviderregistrar.cpp
```diff
@@ -87,6 +88,22 @@ std::wstring CloudProviderRegistrar::registerWithShell(ProviderInfo *providerInf
...
+                WCHAR exePath[MAX_FULL_PATH];
+                if (!GetModuleFileNameW(nullptr, exePath, MAX_FULL_PATH)) {
+                    TRACE_ERROR(L"Error in GetModuleFileNameW");
+                    return std::wstring();
```
luc-guyot-infomaniak:
```suggestion
                    return {};
```

herve-er (dismissed):

extensions/windows/cfapi/Vfs/cloudproviderregistrar.cpp
herve-er: I’m not really in favor of this approach, as it would mean having the error log in the helper function. This makes it complicated to trace the origin of an issue when needed.

ClementKunz: Done in 3b390f1a04bd5a28ea4ab1f2e9e5ea83d0b23984

extensions/windows/cfapi/Vfs/cloudproviderregistrar.cpp
ClementKunz: Done in 3b390f1a04bd5a28ea4ab1f2e9e5ea83d0b23984

herve-er (approved):

luc-guyot-infomaniak (approved):

PR #1335, opened 2025-12-26
chore: add more log

ClementKunz:
Add more logs in `PermissionsHolder` to show the error reason.

### DIFF
--- a/src/libcommonserver/io/permissionsholder.cpp
+++ b/src/libcommonserver/io/permissionsholder.cpp
@@ -36,15 +36,11 @@ PermissionsHolder::PermissionsHolder(const SyncPath &path, const log4cplus::Logg
     AccessRightsInfo accessRights;
     if (const auto ioError = IoHelper::getRights(path, accessRights.read, accessRights.write, accessRights.exec);
         ioError != IoError::Success) {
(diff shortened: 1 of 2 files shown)


src/libcommonserver/io/permissionsholder.cpp
```diff
@@ -36,11 +36,15 @@ PermissionsHolder::PermissionsHolder(const SyncPath &path, const log4cplus::Logg
     AccessRightsInfo accessRights;
     if (const auto ioError = IoHelper::getRights(path, accessRights.read, accessRights.write, accessRights.exec);
         ioError != IoError::Success) {
-        log(std::wstringstream() << L"Failed to get rights for " << Utility::formatSyncPath(path), LogLevel::Error);
+        log(std::wstringstream() << L"Failed to get rights for " << Utility::formatSyncPath(path) << L" - error: "
```
ChristopheLarchier: You could use :
static std::wstring formatIoError(const SyncPath &path, IoError ioError);

(In other places too)

ClementKunz: Indeed, you are right. I have also removed the unnecessary `log` method in c984a00de8d8ee29b19580775ebae2d957d97790.

luc-guyot-infomaniak (approved):

herve-er (approved):

PR #1336, opened 2025-12-29
Test file creation in dedicated subfolder

ClementKunz:
In order to make sure that a file name is supported by the system, the app tries to create a file with the same name in the tmp directory and then removes it. To avoid name clashes in the device's tmp directory, we should create those files in a dedicated subdirectory generated by kDrive.


src/libcommonserver/utility/utility.cpp
herve-er, line 702: To make this more robust: in the (rare?) case where the file already exists, such as after a previous app crash between creation and deletion, a user inadvertently moving files into the `kDrive-tmp` directory, or a bug causing files to be saved there, could we add a mechanism to handle this?
For example, we could create a subdirectory with a random name, with a limited number of retries (e.g., a maximum of 3 attempts).

src/libcommonserver/io/iohelper.cpp
```diff
@@ -631,6 +631,18 @@ bool IoHelper::tempDirectoryPath(SyncPath &directoryPath, IoError &ioError) noex
     return ioError == IoError::Success;
 }
 
+bool IoHelper::kDriveTempDirectoryPath(SyncPath &directoryPath, IoError &ioError) noexcept {
```
ChristopheLarchier: I don't really like using the app name for variables or functions; I would prefer "appTempDirectoryPath".

ClementKunz: Ok, done in 99ef012356d379dfa9050ce1cbe66e7f0a866c68

src/libcommonserver/utility/utility.cpp
ClementKunz: Done in 0e1e3eb32f9b025790a2361355ab5f3511bd7c51

herve-er: Same with `tryCreateTmpFile`, with an added log when a retry occurs to make debugging easier

ClementKunz: Done in e8fc688684292729e348d2c8beb53ebd468ac89b

herve-er (approved):

PR #1347, opened 2026-01-06
fix: do not throw if app fail to load user info

ClementKunz:
In the case where the back end is down, returning 503 errors on any request, the app was crashing on startup because of a deliberately uncaught exception. This is not a friendly behavior. The app initialization is now continuing and an error message is displayed.

### DIFF
--- a/src/server/appserver.cpp
+++ b/src/server/appserver.cpp
@@ -372,8 +372,7 @@ void AppServer::init() {
     connect(OldCommServer::instance().get(), &OldCommServer::restartClient, this, &AppServer::onRestartClientReceived);
 
     // Update users/accounts/drives info
-    const auto exitInfo = updateAllUsersInfo();
-    if (exitInfo.code() == ExitCode::InvalidToken) {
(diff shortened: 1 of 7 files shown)


src/server/appserver.cpp
```diff
@@ -372,15 +372,12 @@ void AppServer::init() {
     connect(OldCommServer::instance().get(), &OldCommServer::restartClient, this, &AppServer::onRestartClientReceived);
 
     // Update users/accounts/drives info
-    ExitCode exitCode = updateAllUsersInfo();
-    if (exitCode == ExitCode::InvalidToken) {
+    const auto exitInfo = updateAllUsersInfo();
+    if (exitInfo.code() == ExitCode::InvalidToken) {
```
luc-guyot-infomaniak:
```suggestion
    if (const auto exitInfo = updateAllUsersInfo(); exitInfo.code() == ExitCode::InvalidToken) {
```
The same pattern seems to apply for similar changes.

ClementKunz: Done in 00f9e613f985111812bfeb645bfce38e1f319f84

luc-guyot-infomaniak (approved):

PR #1358, opened 2026-01-08
feat: change icons in activities page

ClementKunz:


src/gui4/windows/kDrive client/kDrive client/Assets/Custom/Icons/Actions/triangle-alert.svg
herve-er, line 1: This icon already exists in: `src\gui4\windows\kDrive client\kDrive client\Assets\DS\Icons\Actions`.

To address the issue of both the outline and full versions having the same name (`triangle-alert`), I suffixed the outline version with `-outline`.

Image

Image

I already discussed this with the UI/UX team to find a way to uniquely name each icon; this is a work in progress on their side.

If you see no issue with this approach, the same logic can be applied to `circle-check`.

ClementKunz: It is done using `ContentControl` and `DataTemplate`

herve-er (approved):

PR #1354, opened 2026-01-07
feat(guijob): add job to validate a sync path for a new sync

herve-er:
This pull request introduces a new utility for validating sync folder paths and simplifies the logic for finding a suitable local sync folder path by removing the dependency on drive IDs. The changes include the addition of a new job to check if a path is valid for a new sync, refactoring of related methods to remove unnecessary parameters, and updates to the path validity checking logic.

depend on #1346

PR #1360, opened 2026-01-08
WinUI3 - Always use lite sync when available

herve-er:
depend on #1354


src/gui4/windows/kDrive client/kDrive client/ServerCommunication/Interfaces/IServerCommService.cs
```diff
@@ -59,6 +58,19 @@ internal interface IServerCommService
         Task RemoveSync(DbId syncDbId, CancellationToken cancellationToken);
         Task<bool> AddSync(NewSync newSync, CancellationToken cancellationToken);
 
+        // localPath must be an absolute path
+        Task<bool?> PathSupportLiteSync(string localPath, CancellationToken cancellationToken);
```
luc-guyot-infomaniak:
```suggestion
        Task<bool?> PathSupportLiteSync(string absoluteLocalPath, CancellationToken cancellationToken);
```

src/gui4/windows/kDrive client/kDrive client/ServerCommunication/Interfaces/IServerCommService.cs
```diff
@@ -59,6 +58,19 @@ internal interface IServerCommService
...
+
+        public struct GetGoodPathResult
+        {
+            public string? GoodPath;
+            public string? ErrorMessage;
+        }        
+        // Returns a valid path for a new sync as close as possible to the desiredPath, if not known, the driveDbId can be set to -1
+
```
luc-guyot-infomaniak: Can this empty line be removed?

src/gui4/windows/kDrive client/kDrive client/ServerCommunication/Services/CommSharedEnums.cs
```diff
@@ -303,4 +302,11 @@ public enum InconsistencyType
         ForbiddenCharOnlySpaces = 0x080, // The name contains only spaces (not supported by back end)
         ForbiddenCharEndWithSpace = 0x100, // The name ends with a space
     };
+    public enum VirtualFileMode
+    {
+        Off,
+        Win,
+        Mac,
+        Suffix
```
luc-guyot-infomaniak, line 310: This value should be removed as it is now unused (dead code).

src/server/appserver.h
```diff
@@ -133,7 +133,7 @@ class AppServer : public SharedTools::QtSingleApplication {
         [[nodiscard]] ExitInfo stopVfs(int syncDbId, bool unregister);
         [[nodiscard]] ExitInfo startSyncs(User &user);
         void stopSyncTask(int syncDbId);
-        [[nodiscard]] ExitInfo setSupportsVirtualFiles(int syncDbId, bool value);
+        [[nodiscard]] ExitInfo setSupportsVirtualFiles(int syncDbId, bool value, bool asyncResponse);
```
luc-guyot-infomaniak: **Suggestion.** Defines two inline methods `setSupportsVirtualFilesSynchronously` and `setSupportsVirtualFilesAsynchronously` that both call `setSupportsVirtualFiles` with `asyncResponse=false` and `asyncResponse=true` respectively.

src/server/requests/serverrequests.cpp
```diff
@@ -412,7 +412,7 @@ ExitCode ServerRequests::findGoodPathForNewSync(const QString &basePath, QString
             return ExitCode::SystemError;
         }
 
-        folder = basePath + QString::number(attempt);
+        folder = basePath + " " + QString::number(attempt);
```
luc-guyot-infomaniak, line 415: **Suggestion.** Replace the empty space by a non-empty separator, e.g., "_". It eases off debugging.

src/server/appserver.h
herve-er: I’m not convinced that triplicating the method will bring much benefit. The AppServer is already quite heavy. I intentionally didn’t set a default value for the `asyncResponse` parameter to ensure the caller is explicitly aware of how the function will behave.

luc-guyot-infomaniak: The issue I am trying to solve: the developer who reads the lines where calls to `setSupportsVirtualFiles` are written keeps wondering about the meaning of the two boolean arguments.

Alternatively, you could define well-named macros before the actual calls (but this would have to be repeated for each new call).

> The AppServer is already quite heavy. 

That's for sure. But this class will be eventually refactored and split once the UI refactoring is over.

> I’m not convinced that triplicating the method will bring much benefit.

I have a different opinion! The base method with three arguments would be kept private, so it's only two public methods with a private implementation.

src/server/requests/serverrequests.cpp
herve-er: We’ve kept spaces (like “Common document”) because they’re already handled well and feel more natural to users. Swapping them for underscores might make debugging easier, but would hurt readability (especially for non tech users)

herve-er: Other sync software also uses a space (or even " : ")
image

src/gui4/windows/kDrive client/kDrive client/ServerCommunication/Interfaces/IServerCommService.cs
herve-er: Done here: https://github.com/Infomaniak/desktop-kDrive/pull/1360/commits/4358d9e43b918dc8ad0bda6bd9dbbbec55cc9a08

src/gui4/windows/kDrive client/kDrive client/ServerCommunication/Interfaces/IServerCommService.cs
herve-er: Done here: https://github.com/Infomaniak/desktop-kDrive/pull/1360/commits/4358d9e43b918dc8ad0bda6bd9dbbbec55cc9a08

src/gui4/windows/kDrive client/kDrive client/ServerCommunication/Services/CommSharedEnums.cs
herve-er: You seem to have already done it here: [[https://github.com/Infomaniak/desktop-kDrive/pull/1378”](https://github.com/Infomaniak/desktop-kDrive/pull/1378%E2%80%9D)](https://github.com/Infomaniak/desktop-kDrive/pull/1378”)

src/server/appserver.h
herve-er: Done here: https://github.com/Infomaniak/desktop-kDrive/pull/1360/commits/27c73f5131fc8183b03946ac489bf0858c704197

luc-guyot-infomaniak: I forgot to suggest the possible use of enum values of the form `SYNCHRONOUS` and `ASYNCHRONOUS` which also addresses the issue of "Boolean Blindness".

luc-guyot-infomaniak (approved):
Approving, but I am worried about the unrelated random but frequent unit tests failures.

PR #1362, opened 2026-01-08
refactor: New sync observation mechanism used with the GUI

adrien-coye:
Making good use of `allSynchrosPublisher()`
More code added than removed as new UI types are created to keep proper separation of concerns (core / UI)

### DIFF
--- a/src/front/macOS/kDrive/UI/Controllers/MainWindow/MainView/MainViewModel.swift
+++ b/src/front/macOS/kDrive/UI/Controllers/MainWindow/MainView/MainViewModel.swift
@@ -38,9 +38,10 @@ final class MainViewModel {
     init() {
         cacheObservable.usersPublisher
             .allSynchrosPublisher()
-            .receive(on: DispatchQueue.main)
-            .receiveOnMain(store: &bindStore) { [weak self] synchros in
-                let uiSynchros = synchros.map { UISynchroContext(synchroContext: $0) }
+            .map { synchros in
+                synchros.map { UISynchroContext(synchroContext: $0) }
+            }
+            .receiveOnMain(store: &bindStore) { [weak self] uiSynchros in
                 self?.handleUpdatedSynchros(uiSynchros)
             }
     }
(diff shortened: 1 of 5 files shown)


src/front/macOS/kDrive/UI/Controllers/MainWindow/MainView/MainViewModel.swift
```diff
@@ -31,15 +31,18 @@ final class MainViewModel {
     @Published private(set) var currentAccount: UIAccount?
     @Published private(set) var currentDrive: UIDrive?
     @Published private(set) var currentSynchro: UISynchro?
-
-    @Published private(set) var availableUsers = [Int: UIUser]()
+    @Published private(set) var currentSynchros = [UISynchroContext]()
```
adrien-coye: Right now just tracking an array is simple enough IMO.
Later I'll probably switch to an `orderedDictionary` if we need to optimise lookup, but to me this in not necessary in this PR.

src/front/macOS/kDrive/UI/Controllers/MainWindow/MainView/MainViewModel.swift
```diff
@@ -31,15 +31,18 @@ final class MainViewModel {
...
 
     private var bindStore = Set<AnyCancellable>()
 
     init() {
         cacheObservable.usersPublisher
-            .receiveOnMain(store: &bindStore) { [weak self] indexedUser in
-                self?.handleUpdatedUsers(indexedUser)
+            .allSynchrosPublisher()
+            .map { synchros in
+                synchros.map { UISynchroContext(synchroContext: $0) }
```
adrien-coye, line 42: Mapping to UI types is now done in background with pure Combine flow style.

adrien-coye (commented):
Left some notes for the reviewer of this branch

src/front/macOS/kDrive/UI/Controllers/MainWindow/MainView/MainViewModel.swift
valentinperignon: The name `currentX` refers to the currently selected item. `currentSynchros` makes no sense in this context, maybe `availableSynchros` is an option?

adrien-coye: sure

valentinperignon (approved):

PR #1365, opened 2026-01-09
feat(macOS UI): Home view

valentinperignon:
CleanShot 2026-01-12 at 13 54 51@2x


src/front/macOS/kDrive/UI/Views/MainWindow/Home/Components/DriveWebShortcutsView.swift
```diff
@@ -0,0 +1,107 @@
...
+            }
+        }
+        .padding(AppPadding.padding16)
+        .background(ColorToken.Surface.secondary.asColor)
+        .clipShape(RoundedRectangle(cornerRadius: AppRadius.radius16))
+    }
+
+    private func didTapFolder(_ folder: WebFolder) {
+        guard let drive else {
+            return
```
adrien-coye, line 97: Do we want to add a sentry or display an error later maybe ?

src/front/macOS/kDriveCore/ServerBridge/Jobs/DTOs/UserResponse.swift
```diff
@@ -38,14 +38,17 @@ public struct UserInfoResponse: Codable, Sendable {
...
-             isConnected: isConnected,
-             isStaff: isStaff)
+        User(
+            dbId: dbId,
+            userId: userId,
+            name: name,
+            email: email,
+            accounts: [:],
+            availableDrives: [:],
+            avatar: avatar,
```
adrien-coye, line 48: good catch 👍

src/front/macOS/kDrive/UI/Views/MainWindow/Home/Components/SynchroStatusView.swift
```diff
@@ -0,0 +1,128 @@
...
+        let title: String
+        let description: String
+        let buttonTitle: String?
+
+        static func == (lhs: State, rhs: State) -> Bool {
+            return lhs.id == rhs.id
+        }
+
+        static let synchroUpToDate = State(
+            id: 1,
```
adrien-coye, line 49: Maybe we can use a swift enum to explicit an id with a name ?

adrien-coye (commented):
I did a review of the code, I'm about to test the PR.

src/front/macOS/kDrive/UI/Views/MainWindow/Home/Components/SynchroStatusView.swift
valentinperignon: I was wondering which form would be best for `state` "object".
Since it encapsulates several properties (4), I thought a struct would be more readable.
The first option seemed more readable to me, let me know what you think!

**Option 1**
```swift
static let synchroInProgress = State(
    id: 2,
    animation: .cloudSync,
    title: KDriveLocalizable.synchroStatusInProgressTitle,
    description: KDriveLocalizable.synchroStatusInProgressDescription,
    buttonTitle: "Voir les activités"
)
```

**Option 2**
```swift
var animation: ThemedAnimation {
    switch {
        case .option1:
            return …
        case option 2:
            return …
    }
}

var title: String {
    switch {
        case .option1:
            return …
        case option 2:
            return …
    }
}

var description: String {
    switch {
        case .option1:
            return …
        case option 2:
            return …
    }
}

etc…
```

src/front/macOS/kDrive/UI/Views/MainWindow/Home/Components/DriveWebShortcutsView.swift
valentinperignon: Currently, this can be a legitimate scenario when you open the app, before the cache has been refreshed.
But you’re definitely right, we should handle this case better! I’ll note it down for a future PR. Nice catch!

adrien-coye (approved):
I did test the UI, once every comment is closed this can be merged

src/front/macOS/kDrive/UI/Views/MainWindow/Home/Components/SynchroStatusView.swift
adrien-coye: I'm not strongly opinionated on whether option 1 or 2 is better 👍 
_maybe_ option 1 is more `kiss` and readable yes


I was hinting at using something like `id: .synchroInProgress` instead of `id: 2` (if you like it)

PR #1368, opened 2026-01-09
test: try to generate core dumps

ClementKunz:


ClementKunz:
Looks like CI works now...

PR #1376, opened 2026-01-12
fix(network-sockets): enable socket reuse in network jobs to prevent risks of defuncted sockets

luc-guyot-infomaniak:


luc-guyot-infomaniak:
Abandoned: as is, the changes bring no value, only issues.

PR #1384, opened 2026-01-14
WinUI3 Avoid Sentry flooding and add log rolling

herve-er:


src/gui4/windows/kDrive client/kDrive client/Logger/Logger.cs
```diff
@@ -42,95 +41,136 @@ public enum Level
...
-            }
-            else
-            {
-                _logFilePath = "";
-                _logStream = null;
-            }
+            if (App.ServiceProvider.GetRequiredService<AppModel>().Settings.SentryEnabled)
+                SentrySdk.AddBreadcrumb(shortLogEntry, level: ToBreadcrumbLevel(level));
+
+            if (LogLevel == Level.None || (LogLevel == Level.Extended && level != Level.Extended))
```
luc-guyot-infomaniak: I don't understand this line. If `Level.None` were the highest level and `Level.Extended` were the lowest, then the line `if (LogLevel > level) return;` would be enough to filter out log messages with non appropriate levels. Wouldn't it?

src/gui4/windows/kDrive client/kDrive client/Logger/Logger.cs
```diff
@@ -42,95 +41,136 @@ public enum Level
...
 
-            string shortLogEntry = $"{Path.GetFileName(filePath)}:{lineNumber} - {memberName}: {message}";
-            string logEntry = $"{DateTime.Now:yyyy-MM-dd HH:mm:ss} [{level}] {shortLogEntry}";
-            SentrySdk.AddBreadcrumb(shortLogEntry, level: LoggerLevelToBreadcrumbLevel(level));
-#if !DEBUG
-            if (LogLevel > level || _logStream is null)
-                return;
-#endif
-            try
+            if (!App.ServiceProvider.GetRequiredService<AppModel>().Settings.SentryEnabled)
```
luc-guyot-infomaniak, line 106: Given the name of the method, I would move up this line at very beginning of the method's body.

src/gui4/windows/kDrive client/kDrive client/Logger/Logger.cs
herve-er: This was the original implementation, but I later moved to having `None` and `Extended` at the end of the enum. This simplifies communication with the server, as the indexes of both `LogLevel` enums (C# side and C++ side) now match, avoiding any risk of misalignment.

src/gui4/windows/kDrive client/kDrive client/Logger/Logger.cs
herve-er: Done here: https://github.com/Infomaniak/desktop-kDrive/pull/1384/commits/322e543c4249339166a4b451f5bb6b137d4d7489

src/gui4/windows/kDrive client/kDrive client/Logger/Logger.cs
luc-guyot-infomaniak: Thanks for your explanation about the current ordering of log levels.

Reasoning with this ordering, I still don't understand the check `(LogLevel == Level.Extended && level != Level.Extended)`.  If `LogLevel == Level.Extended` is `true`, shouldn't any `level` be filtered in? (I am assuming that `level` is never set with the `None` level).

herve-er: You are 100% right, thanks for pointing it out.

Here is the correction: https://github.com/Infomaniak/desktop-kDrive/pull/1384/commits/c9cb8a98bc7506699dd47983985b7e0c6e2ac4d1

src/gui4/windows/kDrive client/kDrive client/Logger/Logger.cs
```diff
@@ -42,95 +41,137 @@ public enum Level
...
+            if (LogLevel == Level.None) 
+                return;
+
+#if !DEBUG
+            if (LogLevel != LogLevel.Extended && LogLevel > level) return;
+#endif
+
+            _logger?.Write(ToSerilogLevel(level), "{SourceContext}: {Message}", sourceContext, message);
+
+#if DEBUG
```
luc-guyot-infomaniak: I would have expected to read `#if !DEBUG` (send Sentry reports only for released applications) but maybe your intent is different. Is it?

herve-er: Done here https://github.com/Infomaniak/desktop-kDrive/pull/1384/commits/8e36a88f0f5a475f51c1d6216d09c63c6c1bca7c

I followed the server-side logic and enabled Sentry for both Debug and Release builds

luc-guyot-infomaniak (dismissed):

PR #1418, opened 2026-01-21
fix(sync): Fix issue with findGoodPathForNewSync

herve-er:
Fix a mismatch in request parms between the server and the client.
Fix an issue in the path existence check.

### DIFF
--- a/src/libcommon/utility/cstypes.h
+++ b/src/libcommon/utility/cstypes.h
@@ -99,6 +99,7 @@ enum class ExitCause {
     ApiErr,
     InvalidSize,
     FileExists,
+    DirExists,
     FileAccessError,
     FileLocked,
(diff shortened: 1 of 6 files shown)


src/libcommon/utility/cstypes.h
```diff
@@ -99,6 +99,7 @@ enum class ExitCause {
     ApiErr,
     InvalidSize,
     FileExists,
+    DirExists,
```
ClementKunz, line 102: Don't you need to also update the `toString` method?

herve-er: Done: https://github.com/Infomaniak/desktop-kDrive/pull/1418/commits/1edaf2006b4387ffef0d68efee1a0b47d8b9eb43

PR #1425, opened 2026-01-22
Fix find good path

herve-er:


src/server/requests/serverrequests.cpp
```diff
@@ -322,33 +322,35 @@ ExitInfo ServerRequests::isPathValidForNewSync(const SyncPath &path, bool &valid
...
     // If the directory exists, check if it is empty
     bool isEmpty = true;
     std::error_code ec;
-    if (std::filesystem::exists(path, ec) && !ec) {
-        if (!std::filesystem::is_directory(path)) {
+    if (alreadyExists) {
+        if (!std::filesystem::is_directory(path, ec)) {
             LOGW_WARN(Log::instance()->getLogger(), L"The path exists but is not a directory: " << Utility::formatSyncPath(path));
-            valid = false;
             return ExitCode::Ok;
```
ClementKunz, line 338: This should return an error, right? If an item exists with the same name, we will not be able to create the sync folder (at least on Windows and macOS)

src/server/requests/serverrequests.cpp
```diff
@@ -2131,11 +2133,28 @@ ExitInfo ServerRequests::checkPathValidityForNewFolder(const std::vector<Sync> &
         return exitCode;
     }
 
-    // check if the local directory isn't used yet in another sync
     Qt::CaseSensitivity cs = Qt::CaseSensitive;
```
ClementKunz:
```suggestion
    const auto cs = Qt::CaseSensitive;
```

src/server/requests/serverrequests.cpp
herve-er: It does. `isValid` is set to true only if all the checks pass, otherwise, it is false

https://github.com/Infomaniak/desktop-kDrive/pull/1425/changes/BASE..4ce99018fe740da18ba2ec97d7584a53307bee28#diff-384ed8e9c749f83ca771fb5c13c84498ecd78bb1138f1fe3eb3a79098177d999R317

src/server/requests/serverrequests.cpp
herve-er: Done

ClementKunz (approved):

PR #1434, opened 2026-01-26
ci: Add Periphery

valentinperignon:


src/front/macOS/kDriveTests/Cache/CoherentCacheTests/CacheData.swift
adrien-coye, line 1: I would like anything under `kDriveTests` to be excluded from periphery, and mock objects unchanged.
Ex. It's not because I do not use all of a mock at some point in time, that the mock _has_ to be refactored.

adrien-coye (commented):
I did review the code. I'm making a case to exclude tests from this.

src/front/macOS/kDriveCore/AutoIncrementIDGenerator.swift
```diff
@@ -27,12 +27,4 @@ public actor AutoIncrementIDGenerator {
         }
         return counter
     }
-
-    var currentID: Int32 {
```
adrien-coye, line 31: Okay I guess I can extend `AutoIncrementIDGenerator` when I'll need it from a test directly later.

adrien-coye (approved):
Okay it works for me.

PR #1459, opened 2026-02-03
macOS - Encoding issue with Finder extension commands

ChristopheLarchier:
The "Free up local space" and "Make available locally" commands applied to folders do not work for files whose names are encoded in NFC.
The problem stems from the QFileInfoList function (ExtensionJob::processFileList), which returns the contents of a folder with NFC encoding.

### DIFF
--- a/src/server/comm/extensionjob.cpp
+++ b/src/server/comm/extensionjob.cpp
@@ -590,6 +590,7 @@ void ExtensionJob::commandGetThumbnail(const CommString &argument, std::shared_p
     channel->sendMessage(response);
 }
 #endif
+
 #if defined(KD_MACOS)
 void ExtensionJob::commandRetrieveFolderStatus(const CommString &argument, std::shared_ptr<AbstractCommChannel> channel) {
     // This command is the same as RETRIEVE_FILE_STATUS
@@ -753,7 +754,7 @@ void ExtensionJob::manageActionsOnSingleFile(std::shared_ptr<AbstractCommChannel
                                              SyncPalMap::const_iterator syncPalMapIt, VfsMap::const_iterator vfsMapIt,
                                              const Sync &sync) {
     bool exists = false;
-    IoError ioError = IoError::Success;
+    auto ioError = IoError::Success;
     if (!IoHelper::checkIfPathExists(path, exists, ioError) || !exists) {
         return;
     }
@@ -924,6 +925,7 @@ bool ExtensionJob::addDownloadJob(const FileData &fileData, const SyncPath &pare
     return true;
 }
 
+#if defined(KD_MACOS)
 bool ExtensionJob::cancelDownloadJobs(int syncDbId, const std::vector<CommString> &fileList) {
     const std::scoped_lock lock(AppServer::syncPalMapMutex);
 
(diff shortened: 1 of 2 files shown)


src/server/comm/extensionjob.cpp
```diff
@@ -1123,40 +1123,46 @@ void ExtensionJob::processFileList(const std::vector<CommString> &inFileList, st
     // Process all files
     for (const auto &path: inFileList) {
         const FileData fileData = FileData::get(path);
-        if (fileData.isValid() && fileData.virtualFileMode == VirtualFileMode::Mac) {
```
luc-guyot-infomaniak, line 1126: Now that we check only the OS and not the virtual file mode anymore, is it expected that this code is run for a non-LiteSync synchronisation on MacOS? (Probably yes, but I ask to be sure).

src/server/comm/extensionjob.cpp
```diff
@@ -1123,40 +1123,46 @@ void ExtensionJob::processFileList(const std::vector<CommString> &inFileList, st
...
-                        LOGW_WARN(Log::instance()->getLogger(),
-                                  L"Error in ExtensionJob::syncFileStatus - " << Utility::formatSyncPath(tmpPath));
-                        continue;
-                    }
+        if (!fileData.isValid()) continue;
 
-                    if (status == SyncFileStatus::Unknown || status == SyncFileStatus::Ignored) {
-                        continue;
-                    }
+#if defined(KD_MACOS)
```
luc-guyot-infomaniak: **Suggestion:** define two implementations, one for MacOS and the other for the non-Mac OSs. This should be easier to read since the non-Mac OS implementation is trivial (non-recursive) while the MacOs-specific implementation is recursive wrt to directory exploration.

src/server/comm/extensionjob.cpp
```diff
@@ -1123,40 +1123,46 @@ void ExtensionJob::processFileList(const std::vector<CommString> &inFileList, st
...
+        IoError ioError = IoError::Success;
```
luc-guyot-infomaniak:
```suggestion
        auto ioError = IoError::Success;
```

ChristopheLarchier: [25d04f7](https://github.com/Infomaniak/desktop-kDrive/pull/1459/commits/25d04f7bc19b3df4624853eea7e848ac6f6da387)

src/server/comm/extensionjob.cpp
ChristopheLarchier: processFileList is only run for a LiteSync sync on MacOS (see commandMakeAvailableLocallyDirect, commandMakeOnlineOnlyDirect and cancelDownloadJobs).
So I have removed the "#ifdef KD_WINDOWS" case and processFileList is now only defined for macOS.
[25d04f7](https://github.com/Infomaniak/desktop-kDrive/pull/1459/commits/25d04f7bc19b3df4624853eea7e848ac6f6da387)

src/server/comm/extensionjob.cpp
ChristopheLarchier: processFileList is only run for a LiteSync sync on MacOS (see commandMakeAvailableLocallyDirect, commandMakeOnlineOnlyDirect and cancelDownloadJobs).
So I have removed the "#ifdef KD_WINDOWS" case and processFileList is now only defined for macOS.
[25d04f7](https://github.com/Infomaniak/desktop-kDrive/pull/1459/commits/25d04f7bc19b3df4624853eea7e848ac6f6da387)

luc-guyot-infomaniak (approved):

PR #1465, opened 2026-02-03
chore(unit-tests): Implements unit tests for `src/libcommon` with the help of OpenCode

luc-guyot-infomaniak:


src/libcommon/CMakeLists.txt
```diff
@@ -156,3 +159,5 @@ if(ZLIB_FOUND)
     target_link_libraries("${libcommon_NAME}"
         ZLIB::ZLIB)
 endif(ZLIB_FOUND)
+
+
```
herve-er:
```suggestion
```

src/libcommonserver/vfs/win
herve-er, line 1: This file shouldn’t be committed.

Suggestion: You might want to add it to the .gitignore.

test/libcommon/info/testaccountinfo.cpp
```diff
@@ -0,0 +1,119 @@
...
+
+namespace KDC {
+
+void TestAccountInfo::testConstructors() {
+    // Test default constructor
+    AccountInfo info1;
+    CPPUNIT_ASSERT_EQUAL(0, info1.dbId());
+    CPPUNIT_ASSERT_EQUAL(0, info1.userDbId());
+    CPPUNIT_ASSERT_EQUAL(-1, info1.accountId());
+
```
herve-er, line 31: The `name` field was added very recently, so there might be conflicts. It should also be reflected here.

test/libcommon/CMakeLists.txt
```diff
@@ -9,33 +9,38 @@ find_package(log4cplus 2.1.2 REQUIRED)
...
-        ../test.cpp
-
-        test.cpp
-        api_token/testapitoken.cpp api_token/testapitoken.h
-        utility/testutility.cpp utility/testutility.h
-        utility/testtypes.cpp utility/testtypes.h
-        utility/testurlhelper.h utility/testurlhelper.cpp
-        utility/testjsonparserutility.h utility/testjsonparserutility.cpp
-        # Log
-        log/sentry/testsentryhandler.h log/sentry/testsentryhandler.cpp
```
herve-er, line 27: Some of the previous tests seem to be missing here. Is there a reason I’m overlooking?

test/libcommon/test.cpp
```diff
@@ -24,8 +24,8 @@
 #include "log/sentry/testsentryhandler.h"
 #include "utility/testurlhelper.h"
 #include "utility/testjsonparserutility.h"
-#include "log/testlog.h"
-#include "io/testio.h"
```
herve-er, line 28: Why were these replaced instead of added? 😓

herve-er (changes requested):
It looks like some of the previous tests are no longer included (see my comments).

#### Regarding the changes in `KeychainManager`:

I totally see the benefit of `IKeychainStore` for testing purposes, but I feel like the keychain handling is becoming a bit overcomplicated.

The current chain (mixing inheritance and members) is:

`IKeychainStore → KeychainStore → KeychainManager → KeychainManagerSingleton`

`KeychainManager` mainly acts as a wrapper around `KeychainStore` methods, adding logging and Sentry reporting. However, logging is already handled everywhere we call `KeychainManager(Singleton)::Instance()->…`, so this feels somewhat redundant. Its added value is storing `_service` and `_package`, but this could probably be handled directly within `KeychainStore`.

`KeychainManagerSingleton` simply holds an instance of `KeychainManager`, effectively acting as a makeshift dependency injection mechanism. I would suggest waiting a proper DI approach instead of relying on these kinds of workarounds.

This is an interesting subject, we can discuss it anytime tomorrow 😄

src/libcommon/CMakeLists.txt
luc-guyot-infomaniak: Done in 30235897295355c42464137740ddf02b34f7593a.

test/libcommon/info/testaccountinfo.cpp
luc-guyot-infomaniak: Addressed in 30235897295355c42464137740ddf02b34f7593a.

test/libcommon/CMakeLists.txt
luc-guyot-infomaniak: Fixed in 30235897295355c42464137740ddf02b34f7593a.

test/libcommon/test.cpp
luc-guyot-infomaniak: Fixed in 30235897295355c42464137740ddf02b34f7593a.

luc-guyot-infomaniak:
> `KeychainManagerSingleton` simply holds an instance of `KeychainManager`, effectively acting as a makeshift dependency injection mechanism. I would suggest waiting a proper DI approach instead of relying on these kinds of workarounds.
> 
> This is an interesting subject, we can discuss it anytime tomorrow 😄

The `KeyManager` was originally a singleton and changing this design has implications which go, in my opinion, much beyond the scope of this PR. With my current proposal, the singleton pattern is preserved while DI is achieved through inheritance by mocking the `IKeychainStore`. 

What is the *proper DI approach* that you envision?

PR #1466, opened 2026-02-03
feat: show disabled icon

ClementKunz:


luc-guyot-infomaniak (commented):
It seems  that the two `iconPath` methods are duplicated. It would be nice implement a single free helper function and use it twice.

ClementKunz:
> It seems that the two `iconPath` methods are duplicated. It would be nice implement a single free helper function and use it twice.

Done in 290ca7c260bd8d372a40bec10cb401e1fa8b45ae

luc-guyot-infomaniak (approved):
Fantastic!

PR #1471, opened 2026-02-04
Win ui3 gracefully handle server errors step7

herve-er:
depends on #1469

### DIFF
--- a/src/gui4/windows/kDrive client/kDrive client/Pages/Settings/SettingsPage.xaml.cs
+++ b/src/gui4/windows/kDrive client/kDrive client/Pages/Settings/SettingsPage.xaml.cs
@@ -22,6 +22,7 @@ using Infomaniak.kDrive.ViewModels;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.UI.Xaml;
 using Microsoft.UI.Xaml.Controls;
+using Microsoft.UI.Xaml.Input;
 using Microsoft.UI.Xaml.Media;
 using Microsoft.UI.Xaml.Navigation;
 using System;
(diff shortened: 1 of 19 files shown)


src/gui4/windows/kDrive client/kDrive client/Pages/Settings/SettingsPage.xaml.cs
```diff
@@ -176,62 +212,31 @@ private void SyncRulesCard_Clicked(object sender, RoutedEventArgs e)
...
-                return;
             ConsentResult result = await MatomoContentDialog.ShowAsync(this.XamlRoot);
             if (result == ConsentResult.Cancelled)
                 return;
-            await ViewModel.Settings.ChangeMatomoEnabled(result == ConsentResult.Allowed);
+            if (!await ViewModel.Settings.ChangeMatomoEnabled(result == ConsentResult.Allowed))
+            {
+                Logger.Log(Logger.Level.Error, "Failed to change Matomo enabled setting");
+                Utility.ShowUnexpectedErrorTeachingTip();
+            }
```
ClementKunz, line 222: This comment comes very late in the process, I'm sorry about it, but wouldn't it be simpler and require less code to call `Utility.ShowUnexpectedErrorTeachingTip();` from inside the method returning the error?
This comment is not just about `ChangeMatomoEnabled` but to all methods that might trigger a call to `Utility.ShowUnexpectedErrorTeachingTip();`.

herve-er: `Utility.ShowUnexpectedErrorTeachingTip();` is tightly coupled to the GUI and should stay in the GUI layer, since it defines how an error is presented to the user. Calling UI-specific code from `AppViewModel` or `CommLayer` breaks separation of concerns, would be problematic if we later target Uno / Linux and add a litle overhead of marshalling to the UI thread whenever we are calling this method

ClementKunz (approved):

PR #1480, opened 2026-02-06
refactor(sync): Unify sync mode change logic

herve-er:
Unified radio button event handling for sync mode changes. Improved validation, user feedback, and UI consistency.

### DIFF
--- a/src/gui4/windows/kDrive client/kDrive client/Pages/Settings/DriveAdvancedSyncsPage.xaml.cs
+++ b/src/gui4/windows/kDrive client/kDrive client/Pages/Settings/DriveAdvancedSyncsPage.xaml.cs
@@ -7,6 +7,7 @@ using Microsoft.UI.Xaml.Navigation;
 using System;
 using System.Linq;
 using System.Threading;
+using System.Threading.Tasks;
 
 // To learn more about WinUI, the WinUI project structure,
 // and more about our project templates, see: http://aka.ms/winui-project-info.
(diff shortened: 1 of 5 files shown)


src/gui4/windows/kDrive client/kDrive client/Pages/Settings/DriveAdvancedSyncsPage.xaml.cs
```diff
@@ -145,94 +146,52 @@ private async void RemoveSyncButton_Click(object sender, Microsoft.UI.Xaml.Route
...
-        {
-            succeed = await sync.ChangeSyncType(Types.SyncType.Online);
-        }
-        else
-        {
-            Logger.Log(Logger.Level.Error, "Cannot change sync mode: radioButton.DataContext is null or not a sync");
-            succeed = false;
-        }
-
+        if (!radioButton.IsEnabled || radioButton.IsChecked != true)
```
ClementKunz, line 158:
```suggestion
        if (!radioButton.IsEnabled || !radioButton.IsChecked)
```

src/gui4/windows/kDrive client/kDrive client/Pages/Settings/DriveAdvancedSyncsPage.xaml.cs
```diff
@@ -145,94 +146,52 @@ private async void RemoveSyncButton_Click(object sender, Microsoft.UI.Xaml.Route
...
-
         if (canceledByUser)
         {
             Logger.Log(Logger.Level.Info, "User canceled the change to online Sync mode");
-            revertFunc();
+            // This is needed to revert the radio button state back to offline, as changing the sync type to online can fail and we want to reflect that in the UI.
+            if (radioButton.Name == "OnlineRadioButton")
+            {
+                sync.SyncType = Types.SyncType.Online;
+                sync.SyncType = Types.SyncType.Offline;
```
ClementKunz: I am not sure I understand this part, sorry. Is it really necessary to update this variable twice? Changing its value does trigger something?
If mandatory, can you add a comment to explain the process, please?

src/gui4/windows/kDrive client/kDrive client/Pages/Settings/DriveManagementPage.xaml.cs
```diff
@@ -90,113 +91,53 @@ private async void LocationHyperLinkButton_Clicked(object sender, Microsoft.UI.X
...
-            {
-                succeed = await ManagedDrive.MainSync.ChangeSyncType(Types.SyncType.Online);
-            }
-            else
-            {
-                Logger.Log(Logger.Level.Error, "Cannot change sync mode: ManagedDrive or MainSync is null");
-                succeed = false;
-            }
-
+            if (!radioButton.IsEnabled || radioButton.IsChecked != true)
```
ClementKunz, line 104:
```suggestion
            if (!radioButton.IsEnabled || !radioButton.IsChecked)
```

src/gui4/windows/kDrive client/kDrive client/Pages/Settings/DriveAdvancedSyncsPage.xaml.cs
```diff
@@ -145,94 +146,52 @@ private async void RemoveSyncButton_Click(object sender, Microsoft.UI.Xaml.Route
...
-            radioButton.IsEnabled = prevState;
-        };
-
-        if (canceledByUser)
+        if (radioButton is null)
         {
-            Logger.Log(Logger.Level.Info, "User canceled the change to online Sync mode");
-            revertFunc();
+            Logger.Log(Logger.Level.Error, "Online sync mode radio button is null when clicking on online sync mode radio button");
             return;
```
ClementKunz, line 155: We do not need to show an unexpected error message in the GUI in this case?

src/gui4/windows/kDrive client/kDrive client/Pages/Settings/DriveAdvancedSyncsPage.xaml.cs
herve-er: Little trick here: `IsChecked` is a nullable Boolean (`bool?`), so it can have three values (true — checked, false — unchecked, null — partially checked). Therefore, `!radioButton.IsChecked` is not possible (and is prevented by the compiler), as it would be ambiguous about what to return when `IsChecked` is null.

src/gui4/windows/kDrive client/kDrive client/Pages/Settings/DriveManagementPage.xaml.cs
herve-er: Same -> https://github.com/Infomaniak/desktop-kDrive/pull/1480#discussion_r2780992743

src/gui4/windows/kDrive client/kDrive client/Pages/Settings/DriveAdvancedSyncsPage.xaml.cs
herve-er: For the moment, I almost only show a user-facing message in the GUI for errors coming from the comm layer / GUI, as they can happen in normal cases, such as a network loss.

For logic errors (here it would mean that we set the SyncTypeRadioButton_Click event handler on an event of an object which is not a radio button), I only log a message (error level, which will trigger Sentry).

Let me know if you think this should also be done for logical errors. I can handle it in another PR, as it would involve quite a lot of changes.

src/gui4/windows/kDrive client/kDrive client/Pages/Settings/DriveAdvancedSyncsPage.xaml.cs
herve-er: Done here: https://github.com/Infomaniak/desktop-kDrive/pull/1480/commits/ed1bc4fb697ae61692e1836a3ea89362c4a0ad19

It is needed for the UI element to revert to the desired state.

src/gui4/windows/kDrive client/kDrive client/Pages/Settings/DriveAdvancedSyncsPage.xaml.cs
ClementKunz: Sounds good to me

ClementKunz (approved):

PR #1482, opened 2026-02-06
Block app based on version

ClementKunz:
If the application's version is lower than a specific value, the app will be blocked and the synchronizations stopped. The user must update to a newer version

PR #1484, opened 2026-02-06
Check min os version before update

ClementKunz:
Do not show available updates if the OS version is too old.

Depends on #1482

### DIFF
--- a/src/libcommon/utility/utility.cpp
+++ b/src/libcommon/utility/utility.cpp
@@ -1034,10 +1034,6 @@ bool CommonUtility::isVersionLower(const std::string &currentVersion, const std:
     std::vector<uint32_t> targetTabVersion;
     extractIntFromStrVersion(targetVersion, targetTabVersion);
 
-    if (currTabVersion.size() != targetTabVersion.size()) {
-        return false; // Should not happen
-    }
-
     return std::lexicographical_compare(currTabVersion.begin(), currTabVersion.end(), targetTabVersion.begin(),
                                         targetTabVersion.end());
 }
--- a/src/libcommon/utility/utility_linux.cpp
+++ b/src/libcommon/utility/utility_linux.cpp
@@ -68,4 +68,27 @@ bool CommonUtility::hasDarkSystray() {
     return true;
 }
 
+std::string CommonUtility::osVersion() {
+    std::string versionId;
+    // Try to get version from /etc/os-release (standard on modern Linux distributions)
+    std::ifstream osRelease("/etc/os-release");
+    if (osRelease.is_open()) {
+        std::string line;
+        while (std::getline(osRelease, line)) {
+            if (line.find("VERSION_ID=") == 0) {
+                versionId = line.substr(11);
+                // Remove quotes if present
+                if (!versionId.empty() && versionId.front() == '"') {
+                    versionId = versionId.substr(1);
+                }
+                if (!versionId.empty() && versionId.back() == '"') {
+                    versionId.pop_back();
+                }
+                break;
+            }
+        }
+    }
+    return versionId;
+}
+
 } // namespace KDC
--- a/src/libcommon/utility/utility_linux.cpp
+++ b/src/libcommon/utility/utility_linux.cpp
@@ -75,20 +75,28 @@ std::string CommonUtility::osVersion() {
     if (osRelease.is_open()) {
         std::string line;
         while (std::getline(osRelease, line)) {
-            if (line.find("VERSION_ID=") == 0) {
-                versionId = line.substr(11);
+            if (line.find(key) == 0) {
+                value = line.substr(key.length());
                 // Remove quotes if present
-                if (!versionId.empty() && versionId.front() == '"') {
-                    versionId = versionId.substr(1);
+                if (!value.empty() && value.front() == '"') {
+                    value = value.substr(1);
                 }
-                if (!versionId.empty() && versionId.back() == '"') {
-                    versionId.pop_back();
+                if (!value.empty() && value.back() == '"') {
+                    value.pop_back();
                 }
                 break;
             }
         }
     }
-    return versionId;
+    return value;
+}
+
+std::string CommonUtility::osVersion() {
+    return extractOSInfo("VERSION_ID");
+}
+
+std::string CommonUtility::distributionName() {
+    return extractOSInfo("NAME");
 }
 
 } // namespace KDC
--- a/src/libcommon/utility/utility_linux.cpp
+++ b/src/libcommon/utility/utility_linux.cpp
@@ -68,8 +68,8 @@ bool CommonUtility::hasDarkSystray() {
     return true;
 }
 
-std::string CommonUtility::osVersion() {
-    std::string versionId;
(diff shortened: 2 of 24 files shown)


src/libcommon/utility/utility.cpp
```diff
@@ -1034,10 +1034,6 @@ bool CommonUtility::isVersionLower(const std::string &currentVersion, const std:
     std::vector<uint32_t> targetTabVersion;
     extractIntFromStrVersion(targetVersion, targetTabVersion);
 
-    if (currTabVersion.size() != targetTabVersion.size()) {
```
luc-guyot-infomaniak, line 1037: Would it be worth to `LOG_IF_FAIL` or `assert`?

src/libcommon/utility/utility_linux.cpp
```diff
@@ -68,4 +68,35 @@ bool CommonUtility::hasDarkSystray() {
     return true;
 }
 
+std::string extractOSInfo(const std::string &key) {
+    std::string value;
+    // Try to get version from /etc/os-release (standard on modern Linux distributions)
+    std::ifstream osRelease("/etc/os-release");
```
luc-guyot-infomaniak: Doesn't SonarCloud advise you to move the declaration into the `if` clause?

src/libcommon/utility/utility_linux.cpp
```diff
@@ -68,4 +68,35 @@ bool CommonUtility::hasDarkSystray() {
...
+    if (osRelease.is_open()) {
+        std::string line;
+        while (std::getline(osRelease, line)) {
+            if (line.find(key) == 0) {
+                value = line.substr(key.length() + 1); // +1 because the file is formatted "key=value"
```
luc-guyot-infomaniak: It would be nice to have a concrete example to illustrate this extraction in the comments.

src/libcommon/utility/utility_linux.cpp
```diff
@@ -68,4 +68,35 @@ bool CommonUtility::hasDarkSystray() {
...
 }
 
+std::string extractOSInfo(const std::string &key) {
+    std::string value;
+    // Try to get version from /etc/os-release (standard on modern Linux distributions)
+    std::ifstream osRelease("/etc/os-release");
+    if (osRelease.is_open()) {
+        std::string line;
+        while (std::getline(osRelease, line)) {
+            if (line.find(key) == 0) {
```
luc-guyot-infomaniak:
```suggestion
            if (line.find(key) != 0) continue;
```

src/libcommon/utility/utility_mac.mm
```diff
@@ -90,4 +90,12 @@
     }
 }
 
+std::string CommonUtility::osVersion() {
```
luc-guyot-infomaniak, line 93: It seems constness may apply at several places.

src/libcommon/utility/utility_win.cpp
```diff
@@ -138,4 +138,25 @@ std::string CommonUtility::toUnsafeStr(const SyncName &name) {
     return unsafeName;
 }
 
+std::string CommonUtility::osVersion() {
+    // Use RtlGetVersion to get the real Windows version (works on Windows 10/11)
+    typedef LONG(WINAPI * RtlGetVersionPtr)(PRTL_OSVERSIONINFOW);
+
+    HMODULE hMod = GetModuleHandleW(L"ntdll.dll");
+    if (hMod) {
```
luc-guyot-infomaniak:
```suggestion
    if (!hMod) return {};
```

src/libcommon/utility/utility_win.cpp
```diff
@@ -138,4 +138,25 @@ std::string CommonUtility::toUnsafeStr(const SyncName &name) {
...
+        RtlGetVersionPtr RtlGetVersion = (RtlGetVersionPtr) GetProcAddress(hMod, "RtlGetVersion");
```
luc-guyot-infomaniak: This can be moved into the `if` clause below.

src/server/updater/linuxupdater.cpp
```diff
@@ -18,10 +18,52 @@
...
+
+    // Try format: major only
+    if (sscanf(versionStr.c_str(), "%d", &major) == 1) {
+        return true;
+    }
+
+    return false;
+}
+
+static bool compareVersions(int major1, int minor1, int patch1, int major2, int minor2, int patch2) {
```
luc-guyot-infomaniak: The logic and the return value could be modeled on https://en.cppreference.com/w/cpp/string/basic_string/compare.html instead.

Regarding the implementation, the operator <=> could be useful, see https://en.cppreference.com/w/cpp/language/default_comparisons.html.

luc-guyot-infomaniak: Is it used somewhere?

src/libcommon/utility/utility.cpp
ClementKunz: The goal here was to make this method more generic. For example, to be able to compare macOS versions `10.15` vs `26.3.1`

src/libcommon/utility/utility_linux.cpp
ClementKunz: Indeed it has. Sorry for that. Fixed in e3b709f2a4b5d8091dc3bf3b196974b0c4d0bec6

src/libcommon/utility/utility_linux.cpp
ClementKunz: Done in e3b709f2a4b5d8091dc3bf3b196974b0c4d0bec6

src/libcommon/utility/utility_linux.cpp
ClementKunz: Done in 9965f92892b54f6a378bb940dea8794f3e2d6a3d

src/libcommon/utility/utility_mac.mm
ClementKunz: Those are all static functions. Const is not allowed.

src/libcommon/utility/utility_win.cpp
ClementKunz: Done in e49c01876ea6c1509ff487df39835069f8aee9b1

src/libcommon/utility/utility_win.cpp
ClementKunz: Not possible anymore after fix above

src/server/updater/linuxupdater.cpp
ClementKunz: Indeed it's not. In removed this code in 80fac8762dbd83fd88e20a8ea4877de0bbd6f4ed

src/libcommon/utility/utility_mac.mm
luc-guyot-infomaniak: Sorry, it was not very clear. In the body of the function, it seems that some variables are initialized with a value that does not change.

ClementKunz: Done in ffee331e14f35b60f8030dbba4d36e56df7f066f

luc-guyot-infomaniak (approved):

PR #1489, opened 2026-02-09
fix(memory-leaks): opencode fixes for potential memory leaks

luc-guyot-infomaniak:
This PR implements changes suggested by _OpenCode_ to address memory leaks together with some good practices regarding global variables. The latter have been suggested by _SonarCloud_.

### DIFF
--- a/src/libcommon/theme/theme.cpp
+++ b/src/libcommon/theme/theme.cpp
@@ -99,11 +99,11 @@ QIcon Theme::themeIcon(const QString &name, bool sysTray, bool sysTrayMenuVisibl
 }
 
 void Theme::updateIconWithText(QIcon &icon, QString text) const {
-    QList<QSize> sizes = icon.availableSizes();
-    for (QSize size: sizes) {
+    const QList<QSize> &sizes = icon.availableSizes();
+    for (const QSize &size: sizes) {
         QPixmap px = icon.pixmap(size);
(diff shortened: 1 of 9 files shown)


src/libcommon/theme/theme.cpp
```diff
@@ -28,14 +28,8 @@
...
 
-Theme *Theme::_instance = 0;
-
 Theme *Theme::instance() {
-    if (!_instance) {
-        _instance = new Theme;
-        _instance->_mono = false;
-    }
-    return _instance;
+    return &_instance;
```
herve-er, line 32: To push the simplification even further, could this also be a solution?

```suggestion
    static Theme _instance;
    return &_instance;
```

luc-guyot-infomaniak: This is was the initial proposal, but _SonarCloud_ advised against it.

herve-er (commented):
Seems that this new implementation is leading to crash in tests

luc-guyot-infomaniak:
> Seems that this new implementation is leading to crash in tests

Indeed, the initialization of the `PerformanceMonitor` was causing a crash because logging happened in the constructor at a time the logger is not set yet. This is now fixed (as of now, the `PerformanceMonitor` is not used, so unnecessary include statements have been removes).

herve-er (approved):

PR #1493, opened 2026-02-10
Sync loop after a character case modification

ChristopheLarchier:
Issue on macOS with IoHelper::getFileStatus non case/encoding sensitive in LocalFileSystemObserverWorker::changesDetected.

### DIFF
--- a/src/libcommon/utility/utility_base.h
+++ b/src/libcommon/utility/utility_base.h
@@ -55,9 +55,10 @@ inline std::wstring getErrorMessage(DWORD errorMessageID) {
 inline std::wstring getLastErrorMessage() {
     return getErrorMessage(GetLastError());
 }
+
 inline bool isLikeFileNotFoundError(const DWORD dwError) noexcept {
     return (dwError == ERROR_FILE_NOT_FOUND) || (dwError == ERROR_PATH_NOT_FOUND) || (dwError == ERROR_INVALID_DRIVE) ||
-           (dwError == ERROR_BAD_NETPATH);
+           (dwError == ERROR_BAD_NETPATH) || (dwError == ERROR_INVALID_NAME) || (dwError == ERROR_DIRECTORY);
 }
 
 inline bool isLikeFileNotFoundError(const std::error_code &ec) noexcept {
--- a/src/libcommonserver/io/iohelper.h
+++ b/src/libcommonserver/io/iohelper.h
@@ -568,9 +568,12 @@ struct IoHelper {
         // Can be modified in tests.
         static std::function<bool(const SyncPath &path, SyncPath &targetPath, IoError &ioError)> _readAlias;
 #endif
-        static std::function<bool(const SyncPath &path, bool &exists, IoError &ioError)> _checkIfPathExistsSensitive;
+        static std::function<bool(const SyncPath &path, const std::filesystem::file_status &status, bool &exists,
+                                  IoError &ioError)>
+                _checkIfPathExistsSensitive;
         static std::function<bool(const SyncPath &path, FileStat *filestat, IoError &ioError)> _getFileStat;
-        static bool _checkIfPathExistsSensitiveFn(const SyncPath &path, bool &exists, IoError &ioError) noexcept;
+        static bool _checkIfPathExistsSensitiveFn(const SyncPath &path, const std::filesystem::file_status &status, bool &exists,
+                                                  IoError &ioError) noexcept;
         static bool _getFileStatFn(const SyncPath &path, FileStat *filestat, IoError &ioError) noexcept;
         static bool _unsuportedFSLogged;
         static void setCacheDirectoryPath(const SyncPath &newPath);
--- a/src/libcommonserver/io/iohelper.h
+++ b/src/libcommonserver/io/iohelper.h
@@ -79,6 +79,13 @@ struct IoHelper {
                 SyncPath _directoryPath;
                 std::filesystem::recursive_directory_iterator _dirIterator;
         };
+
+        enum class PathCheckOption {
+            Sensitive = 0,
+            Insensitive,
+            EnumEnd
+        };
+
         IoHelper() = default;
 
         inline static void setLogger(const log4cplus::Logger &logger) { _logger = logger; }
@@ -156,11 +163,11 @@ struct IoHelper {
          \param sensitive is a boolean set with true for a case & encoding sensitive check.
          \return true if no unexpected error occurred, false otherwise.
          */
-        static bool getFileStat(const SyncPath &path, FileStat *filestat, IoError &ioError, bool sensitive = false) noexcept;
+        static bool getFileStat(const SyncPath &path, FileStat *filestat, IoError &ioError, PathCheckOption option) noexcept;
 
         // The following prototype throws a std::runtime_error if some unexpected error is encountered when trying to retrieve the
         // file status. This is a convenience function to be used in tests only.
-        static void getFileStat(const SyncPath &path, FileStat *filestat, bool &exists, bool sensitive = false);
+        static void getFileStat(const SyncPath &path, FileStat *filestat, bool &exists, PathCheckOption option);
 
         //! Check if the item indicated by path has a size or a modification date different from the specified ones.
         /*!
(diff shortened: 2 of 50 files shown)


src/libcommonserver/io/iohelper.h
```diff
@@ -153,13 +153,14 @@ struct IoHelper {
          successfully retrieved, nullptr otherwise.
          !!! For a symlink, filestat.nodeType is set with the type of the target !!!
          \param ioError holds the error returned when an underlying OS API call fails.
+         \param sensitive is a boolean set with true for a case & encoding sensitive check.
          \return true if no unexpected error occurred, false otherwise.
          */
-        static bool getFileStat(const SyncPath &path, FileStat *filestat, IoError &ioError) noexcept;
+        static bool getFileStat(const SyncPath &path, FileStat *filestat, IoError &ioError, bool sensitive = false) noexcept;
```
ClementKunz: I would have preferred `sensitive` not have a default value so that we are forced to think about using the sensitive check or not. What do you think?

src/libcommon/utility/utility_base.h
```diff
@@ -55,9 +55,10 @@ inline std::wstring getErrorMessage(DWORD errorMessageID) {
 inline std::wstring getLastErrorMessage() {
     return getErrorMessage(GetLastError());
 }
+
 inline bool isLikeFileNotFoundError(const DWORD dwError) noexcept {
     return (dwError == ERROR_FILE_NOT_FOUND) || (dwError == ERROR_PATH_NOT_FOUND) || (dwError == ERROR_INVALID_DRIVE) ||
-           (dwError == ERROR_BAD_NETPATH);
+           (dwError == ERROR_BAD_NETPATH) || (dwError == ERROR_INVALID_NAME) || (dwError == ERROR_DIRECTORY);
```
luc-guyot-infomaniak, line 61: The downside of this change (`dwError == ERROR_INVALID_NAME`) is that we won't be able to detect immediately a logic error in our code in case the application attempts mistakenly to create a file with a an unreasonable name.

src/libcommonserver/io/iohelper.h
luc-guyot-infomaniak: In the same vein: defining a method `getFileStatSensitive` would help us identify where in the code the case sensitive checks matter, elucidating the meaning of an anonymous bool flag for the different usages..

src/libcommonserver/io/iohelper_linux.cpp
```diff
@@ -45,11 +45,21 @@ bool IoHelper::checkIfFileIsDehydrated(const SyncPath &itemPath, bool &isDehydra
     return true;
 }
 
+bool IoHelper::_checkIfPathExistsSensitiveFn(const SyncPath &path, const std::filesystem::file_status &status, bool &exists,
+                                             IoError &ioError) noexcept {
+    (void) path;
+    (void) status;
+    (void) exists;
```
luc-guyot-infomaniak:
```suggestion
    exists = false;
```

src/libcommonserver/io/iohelper_mac.mm
```diff
@@ -307,4 +308,34 @@ IoError nsError2ioError(NSError *nsError) noexcept {
...
 
+bool IoHelper::_checkIfPathExistsSensitiveFn(const SyncPath &path, const std::filesystem::file_status &status, bool &exists,
+                                             IoError &ioError) noexcept {
+    exists = false;
+    ioError = IoError::Success;
+
+    std::error_code ec;
+    auto filename = path.filename();
+    if (status.type() == std::filesystem::file_type::symlink) {
+        // Used only for symlinks because it can be long
```
luc-guyot-infomaniak: Don't we use it simply because the solution below, that is, `std::filesystem::canonical(path, ec).filename() == filename` cannot apply?

src/libcommonserver/io/iohelper_mac.mm
```diff
@@ -307,4 +308,34 @@ IoError nsError2ioError(NSError *nsError) noexcept {
...
+        for (auto const &dir_entry: std::filesystem::directory_iterator{path.parent_path(), ec}) {
+            if (dir_entry.path().filename() == filename) {
+                exists = true;
+                break;
+            }
+        }
+    } else {
+        auto p = std::filesystem::canonical(path, ec); // canonical does follow symlinks.
```
luc-guyot-infomaniak:
```suggestion
        const auto p = std::filesystem::canonical(path, ec); // canonical does follow symlinks.
```

src/libcommonserver/io/iohelper_mac.mm
```diff
@@ -307,4 +308,34 @@ IoError nsError2ioError(NSError *nsError) noexcept {
     return success;
 }
 
+bool IoHelper::_checkIfPathExistsSensitiveFn(const SyncPath &path, const std::filesystem::file_status &status, bool &exists,
+                                             IoError &ioError) noexcept {
+    exists = false;
+    ioError = IoError::Success;
+
+    std::error_code ec;
+    auto filename = path.filename();
```
luc-guyot-infomaniak:
```suggestion
    const auto filename = path.filename();
```

src/libcommonserver/io/iohelper_mac.mm
ChristopheLarchier: As indicated in the note below, canonical function follows links.

src/libcommonserver/io/iohelper_linux.cpp
ChristopheLarchier: We don't want to update "exists" (and "ioError") as the calling code is:
```
    if (exists && sensitive) {
        // Case & encoding check
        // NB:
        // - On Windows, the standard check is encoding insensitive
        // - On macOS, the standard check is case & encoding insensitive
        return _checkIfPathExistsSensitive(path, status, exists, ioError);
    } else {
        return ioError == IoError::Success || (ioError == IoError::FileNameTooLong) || isExpectedError(ioError);
    }
```

src/libcommonserver/io/iohelper.h
ChristopheLarchier: "bool sensitive" parameter replaced by "IoHelper::PathCheckOption option".
[aae69c3](https://github.com/Infomaniak/desktop-kDrive/pull/1493/commits/aae69c368808d8e6212487f3c3264658dc3f8c24)

src/libcommonserver/io/iohelper_mac.mm
ChristopheLarchier: [aae69c3](https://github.com/Infomaniak/desktop-kDrive/pull/1493/commits/aae69c368808d8e6212487f3c3264658dc3f8c24)

src/libcommonserver/io/iohelper_mac.mm
ChristopheLarchier: [aae69c3](https://github.com/Infomaniak/desktop-kDrive/pull/1493/commits/aae69c368808d8e6212487f3c3264658dc3f8c24)

src/libcommonserver/io/iohelper_linux.cpp
luc-guyot-infomaniak: This is not consistent with other methods where arguments passed by reference are initialized in the body of the function.

**Suggestion.** 
- Do not define `_checkIfPathExistsSensitive` on Linux.
- Define a Linux specific implementation of `IoHelper::checkIfPathExists` which is the plain "insensitive" method. This code can be reused by means of private helper function `IoHelper::checkIfPathExistsInsensitive`.

src/libcommonserver/io/iohelper_mac.mm
luc-guyot-infomaniak: So, it is not _"used only for symlinks **because** it can be long"_. It is used **because** `std::filesystem::canonical` follows symlinks and hence would not do the job.

ChristopheLarchier: This method could be used in all cases but it is not efficient (it involves browsing the contents of a folder).
I reversed the if statement and changed the comment to make it clearer.
[6995401](https://github.com/Infomaniak/desktop-kDrive/pull/1493/commits/69954017802c9835924de97b4651dd68ff2c859b)

src/libcommonserver/io/iohelper_linux.cpp
ChristopheLarchier: Function removed on Linux.
[6995401](https://github.com/Infomaniak/desktop-kDrive/pull/1493/commits/69954017802c9835924de97b4651dd68ff2c859b)

src/libcommon/utility/utility_base.h
ChristopheLarchier: The function is used in this way:
const bool exists = !utility_base::isLikeFileNotFoundError(ec);
So the new code corresponds to what we want (but the name of the function could be changed).

luc-guyot-infomaniak: As discussed offline, we have agreed on not handling `dwError == ERROR_INVALID_NAME` as a special case, even if it can be an hint towards a logical error in some cases. (Too many changes versus no compelling benefit).

luc-guyot-infomaniak (approved):

PR #1502, opened 2026-02-12
test: Observable available drive mechanism

adrien-coye:
I also added a fix on the behaviour of updating an `User` within the observable tree.
User update mechanism in well unit tested (300+L)

`IndexedAvailableDrives` was migrated to an ordered dictionary to preserve order of element given by the dedicated Job

### DIFF
--- a/.github/workflows/macos-redesign.yml
+++ b/.github/workflows/macos-redesign.yml
@@ -70,7 +70,7 @@ jobs:
         with:
           working_directory: src/front/macOS
       - name: SwiftFormat
-        run: swiftformat --lint . --reporter github-actions-log --reporter json
+        run: swiftformat --lint . --reporter github-actions-log
 
   periphery:
     name: Periphery
--- a/.github/workflows/macos-redesign.yml
+++ b/.github/workflows/macos-redesign.yml
@@ -70,7 +70,7 @@ jobs:
         with:
           working_directory: src/front/macOS
       - name: SwiftFormat
-        run: swiftformat --lint . --reporter github-actions-log
+        run: swiftformat --lint . --reporter github-actions-log --reporter json
 
   periphery:
     name: Periphery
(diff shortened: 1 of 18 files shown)


src/front/macOS/kDriveCore/Models/DTOs/User.swift
```diff
@@ -95,7 +95,7 @@ public extension User {
             name: other.name.isEmpty ? name : other.name,
             email: other.email.isEmpty ? email : other.email,
             accounts: other.accounts.isEmpty ? accounts : other.accounts,
-            availableDrives: other.availableDrives.isEmpty ? availableDrives : other.availableDrives,
+            availableDrives: other.availableDrives,
```
adrien-coye: The update policy of a `User` is something I need to check with @herve-er before I open this PR.

adrien-coye: On RequestNum.USER_AVAILABLEDRIVES I need to make sure we simply apply the fetched list to the matching user object

.github/workflows/macos-redesign.yml
```diff
@@ -70,7 +70,7 @@ jobs:
         with:
           working_directory: src/front/macOS
       - name: SwiftFormat
-        run: swiftformat --lint . --reporter github-actions-log
+        run: swiftformat --lint . --reporter github-actions-log --reporter json
```
valentinperignon: Do you want to keep it?

adrien-coye: Discussed, I keep it as it helps IMO to see the issues directly in the CI log, as it can be hard to see if the file is not part of the PR on GH. I'll create a ticket to do the same for other linter CI steps.

valentinperignon (dismissed):

.github/workflows/macos-redesign.yml
adrien-coye: This is done is a specific PR here https://github.com/Infomaniak/desktop-kDrive/pull/1527

valentinperignon (approved):

PR #1512, opened 2026-02-13
feat(trayicon): Add tray icon state & theme support

herve-er:
Refactored tray icon management for sync states and theme changes. 
Added TrayIconUpdater, new PNG icons (light/dark), and error handling.

https://github.com/user-attachments/assets/4107138e-95ae-40c4-af35-6efbb150209c

https://github.com/user-attachments/assets/875a12a3-12c1-4148-8030-28f6850f97d2


src/gui4/windows/kDrive client/kDrive client/App.xaml.cs
```diff
@@ -58,19 +57,20 @@ private set
         internal static IAppConstants Constants => new ProductionAppConstants();
         internal App()
         {
-            TrayIcoManager = new TrayIcon.TrayIconManager();
             _services.AddSingleton<AppModel>();
             _services.AddSingleton<IServerCommProtocol, SocketServerCommProtocol>();
             _services.AddSingleton<IServerCommService, ServerCommService>();
             _services.AddSingleton<UserDefaults>();
+            _services.AddSingleton<TrayIconManager>();
+            _services.AddSingleton<TrayIconUpdater>();
```
ClementKunz: Not very clear to me why we need both those classes. Moreover, `TrayIconUpdater` has a `TrayIconManager ` as attribute. Why a singleton on `TrayIconManager` is also needed here?

herve-er: The main goal was to clearly separate the logic for managing the trayIcon component from the app logic. However, I agree that there is almost no chance we’ll have two tray icons, so this abstraction might just be more confusing.

I merged both classes.

herve-er: here: https://github.com/Infomaniak/desktop-kDrive/pull/1512/commits/4c8722b5b7dae7bf6d83ca93fc326a0a59a444d2

ClementKunz (approved):

PR #1513, opened 2026-02-16
SyncGetPrivateLinkUrl job update

ChristopheLarchier:
Change fileId to nodeId in SYNC_GETPRIVATELINKURL job.

### DIFF
--- a/src/server/comm/guijobs/excltemplsetlistjob.cpp
+++ b/src/server/comm/guijobs/excltemplsetlistjob.cpp
@@ -47,7 +47,7 @@ ExitInfo ExclTemplSetUserListJob::process() {
     }
 
     const std::scoped_lock lock(_commManager->appServer().syncPalMapMutex);
-    for (const auto [_, syncPal]: _commManager->appServer().syncPalMap) {
+    for (const auto &[_, syncPal]: _commManager->appServer().syncPalMap) {
         if (!syncPal) continue;
 
         _commManager->appServer().unregisterSync(syncPal);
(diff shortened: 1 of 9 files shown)


src/server/comm/guijobs/excltemplsetlistjob.cpp
```diff
@@ -47,7 +47,7 @@ ExitInfo ExclTemplSetUserListJob::process() {
     }
 
     const std::scoped_lock lock(_commManager->appServer().syncPalMapMutex);
-    for (const auto [_, syncPal]: _commManager->appServer().syncPalMap) {
+    for (const auto &[_, syncPal]: _commManager->appServer().syncPalMap) {
```
herve-er, line 50: `using SyncPalMap = std::unordered_map<int, std::shared_ptr<SyncPal>>;`

`syncPal` would be a reference to a shared pointer (`std::shared_ptr<SyncPal>&`). Even if it should not be a problem thanks to the mutex, I think we should avoid this, as it breaks the benefit of `shared_ptr` and is not a real gain, since a `shared_ptr` is already quite small.

Let me know what your mind is about it

ChristopheLarchier: Ok, I've gone back.
[55ce084](https://github.com/Infomaniak/desktop-kDrive/pull/1513/commits/55ce08473080520175ab577e1e6f42803e64399a)

luc-guyot-infomaniak: I am bit late to the party, but passing a shared pointer as a `const &` parameter makes sense as long as there is no intent to share ownership : https://stackoverflow.com/questions/3310737/should-we-pass-a-shared-ptr-by-reference-or-by-value/60883927. I believe it is the case here and in several other places.

herve-er (approved):

PR #1525, opened 2026-02-18
feat: Swap to new mac gui

adrien-coye:


adrien-coye:
Old PR for tests, droping now

PR #1529, opened 2026-02-19
feat: Finish "Activity" parsing and updating in cache

adrien-coye:
depends on https://github.com/Infomaniak/desktop-kDrive/pull/1528

UI optimisation / basic `.throttle` will be required in the activity view


valentinperignon (approved):
Nice!

PR #1530, opened 2026-02-19
WinUI3 ErrorPage Step1

herve-er:


src/gui4/windows/kDrive client/kDrive client/Pages/Errors/ErrorPageVM.cs
```diff
@@ -0,0 +1,80 @@
...
+using DynamicData.Binding;
+using Infomaniak.kDrive.Types;
+using Infomaniak.kDrive.ViewModels;
+using System;
+using System.Collections.Generic;
+using System.Collections.ObjectModel;
+
+namespace Infomaniak.kDrive.Pages.Errors
+{
+    partial class ErrorPageVM : UISafeObservableObject, IDisposable
```
ClementKunz, line 11: What does VM stand for?

src/gui4/windows/kDrive client/kDrive client/Pages/Errors/ErrorPageVM.cs
```diff
@@ -0,0 +1,80 @@
...
+        }
+
+        private static bool IsInFileErrorsList(Error error)
+        {
+            return error.ErrorLevel == ErrorLevel.Node;
+        }
+
+        private static bool IsInSyncDirErrorList(Error error)
+        {
+            return false; //error.ErrorLevel == ErrorLevel.SyncPal;
```
ClementKunz, line 64: This normal? Will be implemented later?

src/gui4/windows/kDrive client/kDrive client/Pages/Errors/ErrorPageVM.cs
herve-er: ViewModel: I did not use it much until now, as most of the pages did not contain a lot of page-specific logic and only consumed data as exposed by the AppModel.

However, the Error page includes some logic. For example, the error list is refelcted into three lists (file errors, sync errors, and others). The MVVM pattern recommends placing this kind of logic inside a ViewModel dedicated to the page.

src/gui4/windows/kDrive client/kDrive client/Pages/Errors/ErrorPageVM.cs
herve-er: Yes, I will implement it when working on sync directory errors. For now, they just appear as default errors in the other section

ClementKunz (approved):

PR #1534, opened 2026-02-19
fix(app-image-launch): Use `APPDIR` to build the working directory path on Linux

luc-guyot-infomaniak:


luc-guyot-infomaniak:
Abandoned as the actual fixes is now merged, see https://github.com/Infomaniak/desktop-kDrive/pull/1548.

PR #1549, opened 2026-02-25
feat(io): add isPathOnMountedDisk method and tests

herve-er:
This PR improves detection and handling of cases where the synchronization folder’s disk is missing or not mounted on Linux, macOS, and Windows.

### Key changes

* Added a new exit cause `SyncDirDiskMissing` (C++ and C#) to distinguish missing disks from general access errors.
* Introduced `Utility::exitCauseFromInaccessibleSyncDirectory` to determine whether an inaccessible sync folder is due to a missing disk or a permission/access issue.
* Implemented `IoHelper::isPathOnMountedDisk` on all platforms using OS-specific APIs.
* Updated sync workflow logic to properly pause/ignore sync when the disk is missing.
* Improved user-facing error messages to clearly instruct users to reconnect the drive.
* Minor cleanup: safer list insertion (`Math.Clamp`) and more robust enum handling.

Overall, this makes disk-related failures more accurate, robust, and user-friendly—especially for external or removable drives.

### DIFF
--- a/src/libcommon/log/sentry/handler.cpp
+++ b/src/libcommon/log/sentry/handler.cpp
@@ -486,7 +486,6 @@ SyncPath Handler::getEventFilePath(const AppType appType, const bool crash) {
             fileName = crash ? testCrashEventFileName : testSendEventFileName;
             break;
         default:
-            // assert(false && "Invalid enum value in switch statement.");
             fileName = crash ? clientCrashEventFileName : clientSendEventFileName;
     }
 
--- a/src/libcommon/log/sentry/handler.cpp
+++ b/src/libcommon/log/sentry/handler.cpp
@@ -486,7 +486,7 @@ SyncPath Handler::getEventFilePath(const AppType appType, const bool crash) {
             fileName = crash ? testCrashEventFileName : testSendEventFileName;
             break;
         default:
-            assert(false && "Invalid enum value in switch statement.");
+            // assert(false && "Invalid enum value in switch statement.");
             fileName = crash ? clientCrashEventFileName : clientSendEventFileName;
     }
 
--- a/src/libcommonserver/io/iohelper_linux.cpp
+++ b/src/libcommonserver/io/iohelper_linux.cpp
@@ -26,6 +26,7 @@
 #include <stdio.h>
 #include <string.h>
 #include <fcntl.h>
+#include <filesystem>
 #include <mntent.h>
 #include <sys/types.h>
 #include <sys/stat.h>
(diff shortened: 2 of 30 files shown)


src/libcommon/log/sentry/handler.cpp
```diff
@@ -486,7 +486,7 @@ SyncPath Handler::getEventFilePath(const AppType appType, const bool crash) {
             fileName = crash ? testCrashEventFileName : testSendEventFileName;
             break;
         default:
-            assert(false && "Invalid enum value in switch statement.");
+            // assert(false && "Invalid enum value in switch statement.");
```
ClementKunz: Remove it completely if not wanted

src/libcommonserver/io/iohelper_linux.cpp
```diff
@@ -166,4 +168,53 @@ bool IoHelper::moveItemToTrash(const SyncPath &itemPath) {
     return true;
 }
 
+bool IoHelper::isPathOnMountedDisk(const SyncPath &path, bool &isMounted, IoError &ioError) noexcept {
```
ClementKunz, line 172: I know this is a never-ending discussion, but why not returning directly an `IoError` instead of a boolean value? That would make 1 less parameter to check.

src/libsyncengine/update_detection/file_system_observer/localfilesystemobserverworker.cpp
```diff
@@ -596,13 +596,10 @@ ExitInfo LocalFileSystemObserverWorker::exploreDir(const SyncPath &absoluteParen
...
-            switch (ioError) {
-                case IoError::NoSuchFileOrDirectory:
-                case IoError::AccessDenied:
-                    return {ExitCode::SystemError, ExitCause::SyncDirAccessError};
-                default:
-                    return ExitCode::SystemError;
+            if (ioError == IoError::AccessDenied) {
+                return {ExitCode::SystemError, ExitCause::SyncDirAccessError};
             }
+            return {ExitCode::SystemError, Utility::exitCauseFromInaccessibleSyncDirectory(absoluteParentDirPath)};
```
ClementKunz, line 602: We've lost the `IoError::NoSuchFileOrDirectory` here? Is it wanted?
What if the user had an advanced sync on a mounted disk and the root folder of the advanced sync is deleted?

herve-er: Before this change, both `IoError::NoSuchFileOrDirectory` and `IoError::AccessDenied` resulted in:

```
{ ExitCode::SystemError, ExitCause::SyncDirAccessError }
```

After the change:

* `IoError::AccessDenied` still results in:

  ```
  { ExitCode::SystemError, ExitCause::SyncDirAccessError }
  ```
* `IoError::NoSuchFileOrDirectory` now results in:

  ```
  { ExitCode::SystemError, Utility::exitCauseFromInaccessibleSyncDirectory(absoluteParentDirPath) }
  ```

  This resolves to either:

  * `{ ExitCode::SystemError, ExitCause::SyncDirAccessError }`, or
  * `{ ExitCode::SystemError, ExitCause::SyncDirDiskMissing }`
    depending on whether the disk is mounted.

> What if the user has an advanced sync on a mounted disk and the root folder of the advanced sync is deleted?

If you mean the remote folder: this case is handled elsewhere and results in:

```
{ ExitCode::InvalidSync, ExitCause::SyncDirAccessError }
```

src/libcommonserver/io/iohelper_linux.cpp
herve-er: I would love to, and I had already advocated for this approach. However, while waiting for it to be implemented across all `IoHelper`, I’m keeping the signature consistent with the other class methods.

Following our previous meeting, it will probably evolve into an `ExitInfo` 🙏🏽

src/libcommon/log/sentry/handler.cpp
herve-er: Done here: https://github.com/Infomaniak/desktop-kDrive/pull/1549/commits/5f34ebf433656d18c177540800c34150b7a2f06c

src/libsyncengine/update_detection/file_system_observer/localfilesystemobserverworker.cpp
ClementKunz: I have mixed up `ExitCause` and `IoError`, sorry.

> If you mean the remote folder

No I meant the local one. But in that case also it should end up with `ExitCause::SyncDirAccessError` which is fine.

ClementKunz (approved):

PR #1552, opened 2026-02-26
perf(macOS UI): Throttle node updates

valentinperignon:


src/gui4/macOS/kDrive/Utils/Observers/SynchroState/UISynchroStateObserver.swift
```diff
@@ -0,0 +1,94 @@
...
+}
+
+public protocol UISynchroStateObserving: Sendable {
+    var synchroState: UISynchroState { get }
+    var synchroStatePublisher: AnyPublisher<UISynchroState, Never> { get }
+
+    func observeSynchro(_ synchroDbId: UISynchro.ID)
+}
+
+public final class UISynchroStateObserver: UISynchroStateObserving {
```
adrien-coye: If we do the observation setup during the init, we can switch this to a non mutable struct.

src/gui4/macOS/kDrive/UI/Controllers/MainWindow/MainView/MainViewModel.swift
```diff
@@ -28,57 +28,69 @@ import OrderedCollections
...
+    @LazyInjectService private var synchroStateObserver: UISynchroStateObserving
+    @LazyInjectService private var synchroNodesObserver: UISynchroNodesObserving
+
+    @Published private(set) var currentSynchroContext: UISynchroContext? {
+        didSet {
+            guard let currentSynchroContext else {
+                return
+            }
+
+            synchroStateObserver.observeSynchro(currentSynchroContext.synchro.id)
```
adrien-coye, line 43: We might be able to get away with standard swift UI context set by the root view

adrien-coye: Discussed, this can cause performance drawbacks. I added a ticket to check this on release. We probably can do it with the later @observable (mac OS 14?)

adrien-coye (approved):
I did check with a large synchro, it works well on my machine.

PR #1555, opened 2026-02-26
feat: Adapt waiting time in case of error 429

ClementKunz:
Time to wait in case of rate limit should be read from reply's header.

### DIFF
--- a/src/libsyncengine/jobs/network/abstractnetworkjob.cpp
+++ b/src/libsyncengine/jobs/network/abstractnetworkjob.cpp
@@ -141,8 +141,11 @@ void AbstractNetworkJob::logReplyInfo() {
     }
 }
 
-int AbstractNetworkJob::extractWaitingTime() {
-    if (httpResponse().has(RateLimitHeaderReset)) {
+int64_t AbstractNetworkJob::extractWaitingTime() {
+    int64_t waitingTime = -1;
+    if (httpResponse().has(RateLimitHeaderDelay)) {
+        waitingTime = std::stoi(httpResponse().get(RateLimitHeaderDelay));
+    } else if (httpResponse().has(RateLimitHeaderReset)) {
         int timestamp = 0;
         try {
             timestamp = std::stoi(httpResponse().get(RateLimitHeaderReset));
@@ -150,12 +153,12 @@ int AbstractNetworkJob::extractWaitingTime() {
             LOG_WARN(_logger, "Failed to extract int value from header " << RateLimitHeaderReset << " : " << e.what());
         }
         const auto now = std::chrono::duration_cast<std::chrono::seconds>(std::chrono::system_clock::now().time_since_epoch());
-        return static_cast<int>(timestamp - now.count()) * 1000;
-    }
-    if (httpResponse().has(RateLimitHeaderDelay)) {
-        return std::stoi(httpResponse().get(RateLimitHeaderDelay)) * 1000;
+        waitingTime = static_cast<int64_t>(timestamp - now.count());
     }
 
+    if (waitingTime > 0) {
+        return (waitingTime + CommonUtility::generateRandomNumber(1, 15)) * 1000; // Add a random delay between 1 and 15sec
+    }
     return -1;
 }
 
@@ -518,8 +521,7 @@ ExitInfo AbstractNetworkJob::receiveResponse(const Poco::URI &uri) {
             LOG_WARN(_logger, "Received HTTP_TOO_MANY_REQUESTS, rate limited");
 
             // Update time to wait
-            const auto newWaitTime = extractWaitingTime();
-            if (newWaitTime > 0) {
+            if (const auto newWaitTime = extractWaitingTime(); newWaitTime > 0) {
                 _sleepTime = newWaitTime;
                 LOG_INFO(_logger, "New waiting time: " << _sleepTime);
             }
--- a/src/libsyncengine/jobs/network/abstractnetworkjob.cpp
+++ b/src/libsyncengine/jobs/network/abstractnetworkjob.cpp
@@ -535,6 +535,9 @@ ExitInfo AbstractNetworkJob::receiveResponse(const Poco::URI &uri) {
             if (const auto newWaitTime = extractWaitingTime(); newWaitTime > 0) {
                 _sleepDuration = newWaitTime;
                 LOG_INFO(_logger, "New waiting time: " << _sleepDuration);
+            } else {
+                // If no specific header is provided, wait for the longest acceptable duration
+                _sleepDuration = sleepDurationThreshold - 1;
             }
             return ExitCode::RateLimited;
         }
--- a/src/libsyncengine/jobs/network/abstractnetworkjob.cpp
+++ b/src/libsyncengine/jobs/network/abstractnetworkjob.cpp
@@ -46,8 +46,9 @@
 
 namespace KDC {
 
-const std::string RateLimitHeaderReset = "X-RateLimit-Reset"; // Timestamp (in seconds since epoch)
-const std::string RateLimitHeaderDelay = "Retry-After"; // Delay (in seconds)
+const std::string rateLimitHeaderReset = "X-RateLimit-Reset"; // Timestamp (in seconds since epoch)
+const std::string rateLimitHeaderDelay = "Retry-After"; // Delay (in seconds)
+const int64_t sleepDurationThreshold = 60000; // 60'000 ms -> 1 min
 
 const std::string AbstractNetworkJob::_userAgent = KDC::CommonUtility::userAgentString();
 Poco::Net::Context::Ptr AbstractNetworkJob::_context = nullptr;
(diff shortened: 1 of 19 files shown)


src/libsyncengine/jobs/network/abstractnetworkjob.cpp
```diff
@@ -138,6 +141,24 @@ void AbstractNetworkJob::logReplyInfo() {
     }
 }
 
+int AbstractNetworkJob::extractWaitingTime() {
```
herve-er: In some cases, the delay can be very high.

For example, `LogUploadSession` is limited to 10 per day, so we could end up with a retry after 24 hours, which we don’t want.

We could set a maximum delay of 30 seconds?

src/libsyncengine/jobs/network/abstractnetworkjob.cpp
```diff
@@ -223,6 +244,10 @@ ExitInfo AbstractNetworkJob::runJob() noexcept {
                 break;
             }
 
+            if (outputExitInfo.code() == ExitCode::RateLimited) {
```
herve-er, line 256: To avoid an infinite loop, if for any reason we still get a RateLimit after `MAX_TRIALS` calls, it could be interesting to keep a `_trials` mechanism.

```suggestion
            if (trials < _trials && outputExitInfo.code() == ExitCode::RateLimited) {
```

herve-er (commented):
Remarks:

We might still face issues when starting many jobs in parallel, as they would all hit the rate limit with the same waiting delay. When the delay expires, they would all retry simultaneously, some would get rate-limited again, and this could repeat until all jobs eventually pass.

It is still much better than the current behavior, but it continues to put pressure on the backend.

The only scenario where this could become a real problem is during fulllisting ( when many folders are created at once). However, this should not be a major issue once we ensure listings are triggered only if the folder is effectively moved out of an excluded path.

Edit:

Just saw the other PR https://github.com/Infomaniak/desktop-kDrive/pull/1558 so it should be all good

src/libsyncengine/jobs/network/abstractnetworkjob.cpp
FabianDevel: 30 seconds seems not much for a 429, wouldn't it defeat the goal of the header ?

src/libsyncengine/jobs/network/abstractnetworkjob.cpp
ClementKunz: This condition leads to a `continue` that makes the following loop run again :
```
for (int trials = 1; trials <= std::min(_trials, MAX_TRIALS); trials++) {
```

Therefore, we already have a MAX_TRIALS check

herve-er (commented):
It's a must, but by adding:

```cpp
            if (exitInfo.code() == ExitCode::RateLimited) {
                setPauseDuration(networkJob->sleepDuration());
            }
```

here:
[https://github.com/Infomaniak/desktop-kDrive/blob/9f83896078cc87e8df0a6392bc94d76ae0cd53af/src/libsyncengine/propagation/executor/executorworker.cpp?plain=1#L1373C1-L1373C36](https://github.com/Infomaniak/desktop-kDrive/blob/9f83896078cc87e8df0a6392bc94d76ae0cd53af/src/libsyncengine/propagation/executor/executorworker.cpp?plain=1#L1373C1-L1373C36)

the sync would gracefully handle any rate limit > 60s happening inside the executor.

It could slightly improve the UX: currently, if we get a rate limit of 6s on a download, we pause for 1 minute anyway. With this change, it’s easy to wait only 6s.

src/libsyncengine/propagation/executor/executorworker.cpp
```diff
@@ -1358,6 +1358,10 @@ ExitInfo ExecutorWorker::handleFinishedJob(std::shared_ptr<SyncJob> job, SyncOpP
     }
 
     if (job->exitInfo().code() != ExitCode::Ok) {
+        if (job->exitInfo().code() == ExitCode::RateLimited) {
```
herve-er:
```suggestion
        if (networkJob && job->exitInfo().code() == ExitCode::RateLimited) {
```

Just in case we get a rate limit from a non-network job 🦺

ClementKunz: Of course, thank you for noticing. Fixed in cd4ae77dda047fbf9a7687117f7e8e4ce76c5525

src/libsyncengine/jobs/network/abstractnetworkjob.cpp
```diff
@@ -138,6 +142,27 @@ void AbstractNetworkJob::logReplyInfo() {
     }
 }
 
+int64_t AbstractNetworkJob::extractWaitingTime() {
+    int64_t waitingTime = -1;
+    if (httpResponse().has(rateLimitHeaderDelay)) {
+        waitingTime = std::stoll(httpResponse().get(rateLimitHeaderDelay));
```
herve-er: As for `rateLimitHeaderReset`, this call to `std::stoll` should be wrapped in a try-catch block to avoid a crash if, for any reason, we receive an invalid number

src/libsyncengine/jobs/network/abstractnetworkjob.cpp
```diff
@@ -138,6 +142,27 @@ void AbstractNetworkJob::logReplyInfo() {
...
+    } else if (httpResponse().has(rateLimitHeaderReset)) {
+        int64_t timestamp = 0;
+        try {
+            timestamp = std::stoll(httpResponse().get(rateLimitHeaderReset));
+        } catch (std::exception const &e) {
+            LOG_WARN(_logger, "Failed to extract int value from header " << rateLimitHeaderReset << " : " << e.what());
+        }
+        const auto now = std::chrono::duration_cast<std::chrono::seconds>(std::chrono::system_clock::now().time_since_epoch());
+        waitingTime = static_cast<int64_t>(timestamp - now.count());
```
herve-er: Suggestion:

Move these two lines inside the `try-catch` block, as the calculated waiting time would not make sense if the timestamp is `0`.

That said, the performance impact is negligible, and it should not introduce bugs since negative `waitingTime` values are already handled on line 160.

src/libsyncengine/jobs/network/abstractnetworkjob.cpp
ClementKunz: Done in 46e431f2d78c2ba2a2cc8d53b99d236ccc756dcf

src/libsyncengine/jobs/network/abstractnetworkjob.cpp
ClementKunz: Done in 46e431f2d78c2ba2a2cc8d53b99d236ccc756dcf

src/libsyncengine/jobs/network/abstractnetworkjob.cpp
```diff
@@ -138,6 +142,27 @@ void AbstractNetworkJob::logReplyInfo() {
     }
 }
 
+int64_t AbstractNetworkJob::extractWaitingTime() {
+    int64_t waitingTime = -1;
+    if (httpResponse().has(rateLimitHeaderDelay)) {
+        waitingTime = std::stoll(httpResponse().get(rateLimitHeaderDelay));
+    } else if (httpResponse().has(rateLimitHeaderReset)) {
```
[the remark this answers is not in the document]
ClementKunz: Done in 46e431f2d78c2ba2a2cc8d53b99d236ccc756dcf

src/libsyncengine/jobs/network/abstractnetworkjob.cpp
```diff
@@ -491,6 +525,12 @@ ExitInfo AbstractNetworkJob::receiveResponse(const Poco::URI &uri) {
...
             // Rate limitation
             LOG_WARN(_logger, "Received HTTP_TOO_MANY_REQUESTS, rate limited");
+
+            // Update time to wait
+            if (const auto newWaitTime = extractWaitingTime(); newWaitTime > 0) {
+                _sleepDuration = newWaitTime;
+                LOG_INFO(_logger, "New waiting time: " << _sleepDuration);
+            }
             return ExitCode::RateLimited;
         }
```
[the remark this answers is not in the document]
ClementKunz: Done in b44a8cee2886e9e2933c02594da0ff54977d6ad5

src/libsyncengine/jobs/network/abstractnetworkjob.cpp
```diff
@@ -138,6 +142,32 @@ void AbstractNetworkJob::logReplyInfo() {
...
+        } catch (std::exception const &e) {
+            LOG_WARN(_logger, "Failed to extract int value from header " << rateLimitHeaderReset << " : " << e.what());
+        }
+    }
+
+    if (waitingTime > 0) {
+        return (waitingTime + CommonUtility::generateRandomNumber(1, 15)) * 1000; // Add a random delay between 1 and 15sec
+    }
+    return -1;
+}
```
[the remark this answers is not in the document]
ClementKunz: It is caped un `runJob`:
```
if (outputExitInfo.code() == ExitCode::RateLimited) {
    if (_sleepDuration < sleepDurationThreshold) {
        // The waiting time is short enough, wait and retry to send the request
        continue;
    }
    // Waiting time is too long, pause the sync
    break;
}
```

herve-er (approved):

PR #1557, opened 2026-02-27
feat: Mac GUI - new "parameters" job support

adrien-coye:


src/gui4/macOS/kDriveCore/ServerBridge/Jobs/ParametersJobs.swift
```diff
@@ -0,0 +1,51 @@
...
+
+import Foundation
+import InfomaniakDI
+
+public struct ParametersJobs: Sendable {
+    @LazyInjectService private var queryFetcher: XPCQueryFetcherProtocol
+
+    public init() {}
+
+    @discardableResult
```
valentinperignon: As the only responsibility of this function is to return `ParametersInfo`  (no caching), does it make sense to annotate ` @discardableResult`?

valentinperignon (approved):

PR #1558, opened 2026-02-27
fix: avoid listing request on simple move operations

ClementKunz:
Efficiently handles move operations on already-synchronized directories to prevent redundant API requests.

## Problem
When a move operation occurred on a directory that was already synchronized, the sync engine would still trigger a full listing request (`listing/full` route) to fetch all children. This was redundant since the folder contents were already known and tracked in the snapshot.

## Solution
- Introduced an `alreadySynced` check that considers an item as synced if it either:
  - Exists in the live snapshot, OR
  - Has been moved (tracked in the `movedItems` set)
- Modified the `exploreDir` condition to only request a full listing when the directory is **not** already synced
- Added the item to `movedItems` set during `ActionCodeMoveOut` handling to properly track moved items

## Changes
- `src/libsyncengine/update_detection/file_system_observer/remotefilesystemobserverworker.cpp`
  - Refactored directory exploration logic to check sync status before triggering listing requests
  - Added moved item tracking in move-out action handler

## Impact
Reduces unnecessary API calls to the `listing/full` endpoint, improving sync performance and reducing server load for move operations on existing synchronized folders.

### DIFF
--- a/src/libsyncengine/update_detection/file_system_observer/remotefilesystemobserverworker.cpp
+++ b/src/libsyncengine/update_detection/file_system_observer/remotefilesystemobserverworker.cpp
@@ -624,9 +624,11 @@ ExitInfo RemoteFileSystemObserverWorker::processAction(ActionInfo &actionInfo, s
         case ActionCode::ActionCodeRestore:
         case ActionCode::ActionCodeCreate:
         case ActionCode::ActionCodeRename: {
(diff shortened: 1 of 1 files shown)


luc-guyot-infomaniak (commented):
Looks good to me. Testing soon.

luc-guyot-infomaniak (approved):

PR #1559, opened 2026-02-27
feat: Build a workaround getxattr

adrien-coye:
## Summary of the Fix

I've implemented a **dehydrated files cache** to avoid the `getxattr()` calls in the Endpoint Security callback, which was causing the APFS recursive lock kernel panic.

### Key Changes:

**Extension side (`extensions/MacOSX/kDriveLiteSync/Extension/`):**
1. **`xpcService.h/m`** - Added `_dehydratedFilesCache` NSMutableSet with thread-safe methods
2. **`main.m`** - Replaced `getxattr()` calls with cache lookups using `isFileDehydrated:`
3. **Protocol files** - Added XPC methods for app-to-extension cache management

**Client side (`src/server/comm/litesynccommclient_mac.mm`):**
4. Added `addDehydratedFile:`, `removeDehydratedFile:`, `clearDehydratedFilesCache:` methods
5. Cache is populated when files are created as placeholders or dehydrated in `vfsCreatePlaceHolder()` and `vfsDehydratePlaceHolder()`

### How it works:
- When a file becomes dehydrated (status "O"), the main app notifies the extension via XPC to add it to the cache
- The ES callback checks this cache instead of calling `getxattr()` 
- When a file is hydrated, the extension automatically removes it from cache in `updateFetchStatus:`

This avoids the recursive APFS locking that was causing your kernel panics.

### DIFF
--- a/extensions/MacOSX/kDriveLiteSync/Extension/main.m
+++ b/extensions/MacOSX/kDriveLiteSync/Extension/main.m
@@ -17,14 +17,12 @@
  */
 
 #include "xpcService.h"
-#include "fileAttributes.h"
 
 #import <Foundation/Foundation.h>
 #import <EndpointSecurity/EndpointSecurity.h>
 #import <bsm/libbsm.h>
 
 #include <dispatch/queue.h>
-#include <sys/xattr.h>
 
 es_client_t *g_client = NULL;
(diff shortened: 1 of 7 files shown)


adrien-coye:
Discussed in MD, closing now

PR #1561, opened 2026-02-27
refactor(core): modernize collections and code style

herve-er:
Refactor codebase to use C# 12 collection initializers (`[]`), add `readonly` where appropriate, and clean up unused usings and obsolete comments. Improves code clarity, immutability, and consistency across the project. Also standardizes object initialization and error handling.

### DIFF
--- a/src/gui4/windows/kDrive client/kDrive client/CustomControls/SvgIcon.cs
+++ b/src/gui4/windows/kDrive client/kDrive client/CustomControls/SvgIcon.cs
@@ -59,7 +59,6 @@ namespace Infomaniak.kDrive.CustomControls
 
         private static void OnUriChanged(DependencyObject d, DependencyPropertyChangedEventArgs e)
         {
-            var type = d.GetType();
             if (d is SvgIcon icon && icon._isLoaded)
                 icon.ScheduleRefresh();
         }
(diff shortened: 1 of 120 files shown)


src/gui4/windows/kDrive client/kDrive client/CustomControls/SvgIcon.cs
```diff
@@ -209,8 +208,10 @@ private void ApplyForeground(SvgDocument svgDoc)
             if (double.IsNaN(ratio) || double.IsInfinity(ratio))
                 ratio = 1.0F;
 
-            double targetWidth = 0;
-            double targetHeight = 0;
+
+            double targetWidth;
+
+            double targetHeight;
```
ClementKunz, line 214: Basic types do not need to be initialized in C#?

herve-er: I learned that with this PR too 😺

[https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/default-values](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/default-values)

And it is considered (by the IDE) to be good practice not to explicitly set an initial value when it is equal to the default.

ClementKunz (approved):

PR #1566, opened 2026-03-03
feat(macOS): Connect General Settings to XPC

valentinperignon:


adrien-coye (approved):
Code looks good. I did play with the settings a bit, cool 👍

PR #1570, opened 2026-03-03
fix: macOS panic with Acronis

ChristopheLarchier:
Don't stop anymore `com.apple.quicklook.ThumbnailsAgent`.


adrien-coye (commented):
If we no longer kill the `ThumbnailsAgent` what happens with the thumbnails while we browse the finder ?

FabianDevel:
Fixed on Acronis side

PR #1572, opened 2026-03-04
feat(conflict): fetch additional information on conflicts

herve-er:
depends on #1567

## Summary
Enhances metadata handling for conflicts by introducing new API v3 jobs and structures for retrieving **file information** and **drive user details**, and by adding a **NodeConflictInfo** model.

### User & Drive Information
- Added `GetDriveUserInfoJob` to fetch **user info in a drive** (name, email, avatar URL) via API v3.
- Extended `UserInfo` with **avatar URL** support (serialization, deserialization, equality).

### File Information (API v3)
- Added `GetFileInfoJobV3` to retrieve **file metadata**:
  - creation/modification dates  
  - `author` -> require V3
  - file size

### Conflict Metadata
- Introduced `NodeConflictInfo` to represent conflict metadata:
  - author  
  - file size  
  - last modification date
- Added `NODE_CONFLICT_INFO` to `RequestNum`.

### Caching
- Added `DriveUserInfoCache` for caching drive user information.

### Cleanup
- Removed unused constants from `GetInfoUserJob`.

PR #1574, opened 2026-03-04
feat(winui): bind the conflict dialog to real data

herve-er:
depend on #1572


src/gui4/windows/kDrive client/kDrive client/CustomControls/Errors/Templates/Node/ConflictError.xaml.cs
```diff
@@ -54,7 +53,7 @@ private async void ErrorCard_ActionClick(object sender, RoutedEventArgs e)
                 }
                 else
                 {
-                    await Utility.OpenFolderSecurely(absolutPath);
+             //       await Utility.OpenFolderSecurely(absolutPath);
```
ClementKunz, line 56: So we don't want to open the folder anymore?

herve-er: It will be replaced by the actual server call that starts the conflict resolution propagation (PR in progress).

ClementKunz (approved):

PR #1576, opened 2026-03-04
fix(utility): replace system calls with execvp for trash operations

R-Gld:


ChristopheLarchier (dismissed):

luc-guyot-infomaniak (commented):
It's a bit longer more than just calling `QFile::moveToTrash`.

R-Gld:
We chose to use the Qt implementation for moving files to the trash (see #1577 using `QFile::moveToTrash()`) for version 3.7.10, allowing us to reduce potential risks.

This method, or a similar one, will soon be reimplemented in future Linux versions of the application, allowing us to avoid including additional Qt code in the server.

PR #1579, opened 2026-03-05
Fix pgm7461 1781 on v3.8.2

ChristopheLarchier:


R-Gld (approved):
Same fix as the one did here: #1577  👍🏼

PR #1580, opened 2026-03-05
feat(gui): add conflict resolution jobs

herve-er:
depends on #1574


src/libsyncengine/syncpal/conflictingfilescorrector.cpp
```diff
@@ -31,14 +31,21 @@
...
     _syncPal(syncPal),
-    _keepLocalVersion(keepLocalVersion),
-    _errors(std::move(errors)) {}
+    _keepLocalErrors(keepLocalErrorList),
+    _keepRemoteErrors(keepRemoteErrorList) {}
 
 ExitInfo ConflictingFilesCorrector::runJob() {
-    for (auto &error: _errors) {
+    resolveConflicts(_keepLocalErrors, true);
+    resolveConflicts(_keepRemoteErrors, false);
```
ClementKunz: We are now trying to avoid as much as possible boolean flags like this, for readability's sake. Maybe create an enum value instead?

src/libsyncengine/syncpal/syncpal.cpp
```diff
@@ -1065,7 +1065,19 @@ ExitCode SyncPal::excludeListUpdated() {
     return ExitCode::Ok;
 }
 
-ExitCode SyncPal::fixConflictingFiles(bool keepLocalVersion, std::vector<Error> &errorList) {
+ExitCode SyncPal::fixConflictingFilesAsync(const std::vector<Error> &keepLocalErrorList,
+                                           const std::vector<Error> &keepRemoteErrorList) {
+    std::vector<int64_t> removedErrorsDbIds; // Not used in (legacy) async mode
+    return fixConflictingFiles(keepLocalErrorList, keepRemoteErrorList, removedErrorsDbIds, true);
```
ClementKunz: Same remark about the boolean flag.
Perhaps split those methods as follows: one method that generates the job to be called asynchronously in `fixConflictingFilesAsync` and synchronously in `fixConflictingFiles`.

src/server/comm/guijobs/abstracterrorresolveconflictsjob.cpp
```diff
@@ -0,0 +1,108 @@
...
+                 "AbstractErrorResolveConflictsJob::getSyncDbIdFromErrors: Expected exactly one syncDbId, found "
+                         << syncDbIdSet.size());
+        return ExitCode::LogicError;
+    }
+
+    syncDbId = *syncDbIdSet.begin();
+    return ExitCode::Ok;
+}
+
+ExitInfo AbstractErrorResolveConflictsJob::getSyncPal(const int64_t syncDbId, std::shared_ptr<SyncPal> &syncPal) {
```
ClementKunz: Such methods exist in at least 3 classes: `SyncStatusJob`, `ExtensionJob` and `AbstractErrorResolveConflictsJob`. Maybe we can move this logic of retrieving the associated `SyncPal` in a parent or utility class. 
Especially since you are correctly using the mutex here, but it is not done everywhere.

But maybe this should be the topic of a dedicated PR

src/server/comm/guijobs/abstracterrorresolveconflictsjob.cpp
```diff
@@ -0,0 +1,108 @@
...
+        LOG_INFO(_logger,
+                 "The number of removed error dbIds returned by SyncPal::fixConflictingFiles "
+                 "differs from the number of provided errors. This is normal if some conflicts have been manually resolved "
+                 "(removedErrorsDbIds="
+                         << removedErrorsDbIds.size() << ", expected=" << totalExpected << ")");
+    }
+
+    for (const auto &dbId: removedErrorsDbIds) {
+        // SyncPal doesn't have access to the CommManager, so we notify the GUI about removed errors here.
+        _commManager->sendGuiSignal(std::make_shared<SignalErrorRemovedJob>(dbId));
```
ClementKunz, line 86: I am getting off topic here, but I feel like creating a shared pointer here is unnecessary since this class does not use it. Shouldn't we have more specific methods like `sendErrorRemovedSignal` in `CommManager` and let the comm manager manage its internal data, jobs, pointer, ...
Moreover, this would make the inclusion of `signalerrorremovedjob.h` unnecessary.

But again, if we want to implement this, it should be the topic of a dedicated PR and be done everywhere.

src/server/comm/guijobs/errorresolveconflictsquickjob.cpp
```diff
@@ -0,0 +1,148 @@
...
+                    LOG_WARN(_logger, "Error getting file stat for original path, defaulting to keepLocal for errorDbId="
+                                              << error.dbId());
+                    keepLocalErrors.push_back(error);
+                    break;
+                }
+
+                if (!IoHelper::getFileStat(conflictAbsPath, &conflictStat, ioError) || ioError != IoError::Success) {
+                    LOG_WARN(_logger,
+                             "Error getting file stat for conflict path, defaulting to keepLocal for errorDbId=" << error.dbId());
+                    keepLocalErrors.push_back(error);
```
ClementKunz, line 127: I am not sure to I understand why we are pushing the error into `keepLocalErrors` since we are breaking anyway right after. Shouldn't we return an ExitInfo immediately and abort the process of resolving the conflicts?

src/server/comm/guijobs/guijobfactory.cpp
```diff
@@ -118,6 +120,8 @@ GuiJobFactory::GuiJobFactory() {
                 {RequestNum::NODE_CREATEMISSINGFOLDERS, makeShared<NodeCreateMissingFoldersJob>},
                 {RequestNum::NODE_CONFLICT_INFO, makeShared<NodeConflictInfoJob>},
                 {RequestNum::ERROR_INFOLIST, makeShared<ErrorInfolistJob>},
+                {RequestNum::ERROR_RESOLVE_CONFLICTS, makeShared<ErrorResolveConflictsJob>},
+                {RequestNum::ERROR_RESOLVE_CONFLICTS_QUICK, makeShared<ErrorResolveConflictsQuickJob>},
```
ClementKunz, line 124: What is the difference between those 2 jobs?
Maybe explicit this somewhere

ClementKunz:
There are also a few SonarCloud issues that can be handled, such as those about constness or using return values.

src/libsyncengine/syncpal/conflictingfilescorrector.cpp
herve-er: Done here: https://github.com/Infomaniak/desktop-kDrive/pull/1580/commits/0c12241b745ae65c8bf18ed4197b857e270c08a2

src/libsyncengine/syncpal/syncpal.cpp
herve-er: Done here: https://github.com/Infomaniak/desktop-kDrive/pull/1580/commits/0c12241b745ae65c8bf18ed4197b857e270c08a2

src/server/comm/guijobs/abstracterrorresolveconflictsjob.cpp
herve-er: The first base class between **GUI jobs** and **extension jobs** doesn’t have access to **AppServer**. However, by moving `getSyncPal` in `AbstractGuiJob`, we can already solve the issue for all GUI jobs, done here: https://github.com/Infomaniak/desktop-kDrive/pull/1580/commits/0c12241b745ae65c8bf18ed4197b857e270c08a2

src/server/comm/guijobs/abstracterrorresolveconflictsjob.cpp
herve-er: CommManager doesn't have to know about the content of what it is transporting, it just takes an AbstractGuiJob as input and processes it.

Otherwise, if we want to stay consistent, we would need to add a dedicated function and signal for each job in CommManager, which would become hard to maintain

src/server/comm/guijobs/errorresolveconflictsquickjob.cpp
herve-er: We are just breaking the case and will continue with the next conflict. Aborting the conflict resolution for that would mean that if the user has several conflicts, and one of the remote or local files involved in any of them has been moved, then none of the conflicts would be resolved.

We have two options:

* We can decide to keep the local version. If it doesn't exist anymore, the conflict will resolve itself later. Otherwise, even if it is not the most recent version, we ensure that no data is lost, as the remote version is available in the file versions. If we choose the remote version, we would permanently delete the local version.

* Or we can simply ignore the conflict, but that would create friction for the user, who would then need to resolve all the conflicts manually, one by one.

src/server/comm/guijobs/guijobfactory.cpp
herve-er: Done here: https://github.com/Infomaniak/desktop-kDrive/pull/1580/commits/eebb7fb0caaeb8413a8c6030a539d2e9e3a0b714

src/server/comm/guijobs/errorresolveconflictsquickjob.cpp
ClementKunz: Sorry, indeed it is not the for loop that is broken, my bad. That is perfect.

> We can decide to keep the local version. If it doesn't exist anymore, the conflict will resolve itself later.

Ok, I did not understand that at first. Maybe a comment to indicate that would be helpful

src/server/comm/guijobs/abstracterrorresolveconflictsjob.cpp
ClementKunz: Thanks!
I am still a bit worried of the `extension job` not using the mutex.
Do you think it could cause concurrency issues?

herve-er: All calls to `ExtensionJob::retrieveSyncPalMapIt` are preceded by acquiring the lock:

```cpp
const std::scoped_lock lock(_commManager->appServer().vfsMapMutex);
const auto syncPalMapIt = retrieveSyncPalMapIt(fileData.syncDbId);
```

I don’t know why it was implemented this way instead of acquiring the lock directly inside `retrieveSyncPalMapIt`, but it works and changing that here would clearly be out of scope 😅

However, I did find a missing lock in `ExtensionJob::addDownloadJob`, which I added here:
[https://github.com/Infomaniak/desktop-kDrive/pull/1580/commits/7dddac0a2bea534fb0deedff296c2e127b3ae2b2](https://github.com/Infomaniak/desktop-kDrive/pull/1580/commits/7dddac0a2bea534fb0deedff296c2e127b3ae2b2)

src/server/comm/guijobs/errorresolveconflictsquickjob.cpp
herve-er: For sure, done here: [https://github.com/Infomaniak/desktop-kDrive/pull/1580/commits/b3b528a8ec25d467a769ce2307210624a22e720d](https://github.com/Infomaniak/desktop-kDrive/pull/1580/commits/b3b528a8ec25d467a769ce2307210624a22e720d)

ClementKunz (approved):

PR #1586, opened 2026-03-06
feat(conflicts): propagate user selection to the server (WinUI)

herve-er:
depends on #1580

### DIFF
--- a/src/libsyncengine/syncpal/syncpal.cpp
+++ b/src/libsyncengine/syncpal/syncpal.cpp
@@ -323,6 +323,10 @@ ExitCode SyncPal::clearNodes() {
 
 void SyncPal::syncPalStartCallback([[maybe_unused]] UniqueId jobId) {
     auto jobPtr = SyncJobManagerSingleton::instance()->getJob(jobId);
+    if (!jobPtr && _conflictingFilesCorrector) {
+        jobPtr = _conflictingFilesCorrector;
+    }
+
     if (jobPtr) {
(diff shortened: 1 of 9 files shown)


src/libsyncengine/syncpal/syncpal.cpp
```diff
@@ -323,6 +323,10 @@ ExitCode SyncPal::clearNodes() {
 
 void SyncPal::syncPalStartCallback([[maybe_unused]] UniqueId jobId) {
     auto jobPtr = SyncJobManagerSingleton::instance()->getJob(jobId);
+    if (!jobPtr && _conflictingFilesCorrector) {
+        jobPtr = _conflictingFilesCorrector;
+    }
```
ClementKunz, line 332: So after resolving conflict, the SyncPal start callback is called without a valid job ID? Why is it different than other `Propagator`? 
And if it has to be different, can you add a comment expliciting it please?

herve-er: Done here: https://github.com/Infomaniak/desktop-kDrive/pull/1586/commits/aa39ad739814c44cd7300f0c808db510cf3a4800

```cpp
// _conflictingFilesCorrector may run synchronously (i.e. without SyncJobManager).
// If other jobs also move to synchronous execution, SyncPal should maintain
// an internal map to associate jobId with the actual job instance.
```

ClementKunz (approved):

PR #1587, opened 2026-03-06
fix(unit-tests): Fixes the `isInTrash` Linux implementation

luc-guyot-infomaniak:
This pull-request re-implements on Linux the helper function `IoHelper::isInTrash` that checks whether an item is located in the user trash in a reliable way. This new implementation follows the [Freedesktop specifications](https://specifications.freedesktop.org/trash/latest/).
This helper is used in unit tests to check whether some clean up tasks are executed successfully. With these changes, the related tests will not fail anymore for unclear and seemingly random reasons.

### DIFF
--- a/test/test_utility/testhelpers_linux.cpp
+++ b/test/test_utility/testhelpers_linux.cpp
@@ -32,7 +32,7 @@
 #include <utime.h>
 
 namespace KDC::testhelpers {
-
+namespace {
 SyncPath removeNumericSuffix(const SyncPath &relativePath) {
     if (relativePath.empty()) return {};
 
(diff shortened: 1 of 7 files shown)


test/test_utility/testhelpers_linux.cpp
```diff
@@ -61,6 +61,59 @@ SyncPath removeNumericSuffix(const SyncPath &relativePath) {
...
+
+
+// Parse the mandatory `Path` entry from a .trashinfo file
+std::string getOriginalPath(const SyncPath &infoFile) {
+    std::ifstream file(infoFile);
+    if (!file.is_open()) return "";
+
+    std::string line;
+    while (std::getline(file, line)) {
+        if (line.rfind("Path=", 0) != 0) continue;
```
ChristopheLarchier: You could use a constant string for "Path=" and then the length of the string instead of "5".

luc-guyot-infomaniak: Great idea, thanks! Done in 0f6628626d2901e926b4f5b2253c44aa50303b15.

ChristopheLarchier (approved):

PR #1589, opened 2026-03-06
feat(conflicts): add quick bulk resolve flow for file conflicts (WinUI)

herve-er:
depends on #1586 

## Quick Resolve Flow for File Conflicts

Introduces a dedicated `ConflictQuickResolvePage` enabling bulk resolution of file conflicts via three strategies: keep most recent, keep local, or keep remote. Integrated into the error page with a "Manage" button for high-volume conflicts.

### Changes
- **New Feature**: Bulk conflict resolution UI with strategy selection
- **Localization**: Added `GetString1i` method for dynamic integer arguments
- **UI Polish**: Refined padding, margins, and typography in error lists and dialogs
- **Stability**: Null-safety in navigation, proper ViewModel disposal

### Testing
Verify bulk resolution strategies apply correctly across multiple conflicts and that the error page displays the new management card when conflicts exceed the threshold.

### DIFF
--- a/src/gui4/windows/kDrive client/kDrive client/Pages/Errors/ErrorPageVM.cs
+++ b/src/gui4/windows/kDrive client/kDrive client/Pages/Errors/ErrorPageVM.cs
@@ -3,15 +3,18 @@ using DynamicData.Binding;
 using Infomaniak.kDrive.CustomControls.Errors;
 using Infomaniak.kDrive.Types;
 using Infomaniak.kDrive.ViewModels;
+using Microsoft.Extensions.DependencyInjection;
 using System;
 using System.Collections.Generic;
 using System.Collections.ObjectModel;
 using System.Linq;
+using static Infomaniak.kDrive.ViewModels.AppModel;
 
 namespace Infomaniak.kDrive.Pages.Errors
 {
     partial class ErrorPageVM : UISafeObservableObject, IDisposable
     {
+        private AppModel _viewModel = App.ServiceProvider.GetRequiredService<AppModel>();
 
         private Sync? _sync;
         private const int _maxConflictsForIndividualDisplay = 5;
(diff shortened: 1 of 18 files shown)


src/gui4/windows/kDrive client/kDrive client/Pages/Errors/ErrorPageVM.cs
```diff
@@ -85,9 +138,20 @@ private void InitErrorLists()
...
+            if (HasManyConflicts && IsConflictUserResolvable(error))
+                return false;
+
+            return true;
+        }
+
+        private static bool IsConflictUserResolvable(Error error)
         {
-            return error.ErrorLevel == ErrorLevel.Node;
+            return error.ErrorLevel == ErrorLevel.Node && (error.ConflictType == ConflictType.CreateCreate || error.ConflictType == ConflictType.EditEdit);
```
ClementKunz, line 154: There is also Move-Create conflict with move on remote side that can lead to a file renamed with the conflict suffix

Image

herve-er: In the current GUI, we only allow the user to resolve conflicts caused by edit/edit or create/create operations. I just tried the move (remote) / create (local) case, and we simply display the error + rename with conflict.

image

I agree that it’s not clear why it behaves like this. I will run some tests to see if we can safely and easily handle this case. If not, I will create a card to add this feature (or fix this bug, depending on the point of view 😄)

herve-er: It is a bit tricky because the handling is not the same depending on the direction (move local vs. move remote) and the old gui logic will also need to be updated to support this.

I have created a card dedicated to this topic.

ClementKunz (approved):

PR #1609, opened 2026-03-10
fix: invalid abstract token network job cache

herve-er:
## Summary

When a user is deleted via `ServerRequests::deleteUser`, all related `accounts / drives / syncs` are removed **directly by database cascade** (foreign keys).

Because of this, `ServerRequests::deleteDrive` is **not called**, which means the `AbstractTokenNetworkJob` cache is not cleaned.

As a result, when the user later creates a new drive, it may reuse a previous `driveDbId`, and `AbstractTokenNetworkJob` can incorrectly retrieve the **cached ID of the old drive**.

This issue did not occur previously because the only way to remove a user was by deleting all their drives first. However, the **new GUI introduces a direct “disconnect account” action**, which exposes this problem.


luc-guyot-infomaniak (commented):
The changes look good to me. Proceeding with the tests now.

luc-guyot-infomaniak (approved):
Tested successfully.

PR #1628, opened 2026-03-16
Fix/mititgate 3.8.2.5 and .7 deletion issue

herve-er:


src/libsyncengine/db/syncdb.cpp
```diff
@@ -544,11 +545,239 @@ bool SyncDb::upgrade(const std::string &fromVersion, const std::string &toVersio
...
+                                                 "Error preparing delete node request");
+            return false;
+        }
+
+        std::function freeRequests = [this]() {
+            queryFree(SELECT_NODE_BY_PARENTNODEID_ROOT_REQUEST_ID);
+            queryFree(SELECT_NODE_BY_PARENTNODEID_REQUEST_ID);
+            queryFree(SELECT_NODE_BY_NODEID_FULL_ID);
+            queryFree(DELETE_NODE_REQUEST_ID);
+        };
```
ClementKunz: We free the requests only in case of success?
Therefore, what's the point of defining this method here since it is called only once on line 593?

src/libsyncengine/db/syncdb.cpp
```diff
@@ -544,11 +545,239 @@ bool SyncDb::upgrade(const std::string &fromVersion, const std::string &toVersio
...
+
+        if (!revertAllLocalDeletes()) {
+            LOG_ERROR(_logger, "Error reverting all local deletes");
+            KDC::sentry::Handler::captureMessage(KDC::sentry::Level::Error, "SyncDb::upgrade::revertAllLocalDeletes",
+                                                 "Error reverting all local deletes");
+            return true;
```
ClementKunz: We return true even though the call to `revertAllLocalDeletes` has errors?

src/libsyncengine/db/syncdb.cpp
```diff
@@ -544,11 +545,239 @@ bool SyncDb::upgrade(const std::string &fromVersion, const std::string &toVersio
...
+
     LOG_DEBUG(_logger, "Upgrade of Sync DB successfully completed.");
 
     return true;
 }
 
+bool SyncDb::revertAllLocalDeletes() {
+    // Revert all local deletes by removing from the DB all nodes that are not present on the local file system anymore.
+
+    IoError ioError;
```
ClementKunz: Variable initialization

src/libsyncengine/db/syncdb.cpp
```diff
@@ -544,11 +545,239 @@ bool SyncDb::upgrade(const std::string &fromVersion, const std::string &toVersio
...
+                                                 "Duplicate local node id " + dbNode.nodeIdLocal().value() + " found for dbId " +
+                                                         std::to_string(dbId) + ", skipping it for local delete revert");
+        }
+    }
+
+    return true;
+}
+
+bool SyncDb::fsFileLocalNodeIds(const SyncPath &localSyncPath, std::unordered_set<NodeId> &localFSNodeIds) {
+    IoError ioError;
```
ClementKunz: Variable initialization

src/libsyncengine/db/syncdb.cpp
herve-er: In case of error, we don't want to **block** the user indefinitely.

However, you're right that we **shouldn't return here at all**, as the migration should continue

src/libsyncengine/db/syncdb.cpp
herve-er: Done here: https://github.com/Infomaniak/desktop-kDrive/pull/1628/commits/df3869594bb50135f2ba750ff2486bb4fe8a950d

src/libsyncengine/db/syncdb.cpp
herve-er: Done here: https://github.com/Infomaniak/desktop-kDrive/pull/1628/commits/df3869594bb50135f2ba750ff2486bb4fe8a950d

src/libsyncengine/db/syncdb.cpp
herve-er: Done here: https://github.com/Infomaniak/desktop-kDrive/pull/1628/commits/df3869594bb50135f2ba750ff2486bb4fe8a950d

src/libsyncengine/db/syncdb.cpp
herve-er: Done here: https://github.com/Infomaniak/desktop-kDrive/pull/1628/commits/df3869594bb50135f2ba750ff2486bb4fe8a950d

src/libsyncengine/db/syncdb.cpp
```diff
@@ -544,11 +545,239 @@ bool SyncDb::upgrade(const std::string &fromVersion, const std::string &toVersio
...
+            LOG_ERROR(_logger, "Error preparing select node by nodeId full request");
+            KDC::sentry::Handler::captureMessage(KDC::sentry::Level::Error, "SyncDb::upgrade::revertAllLocalDeletes",
+                                                 "Error preparing select node by nodeId full request");
+            return false;
+        }
+        if (!createAndPrepareRequest(DELETE_NODE_REQUEST_ID, DELETE_NODE_REQUEST)) {
+            LOG_ERROR(_logger, "Error preparing delete node request");
+            KDC::sentry::Handler::captureMessage(KDC::sentry::Level::Error, "SyncDb::upgrade::revertAllLocalDeletes",
+                                                 "Error preparing delete node request");
+            return false;
```
ClementKunz, line 576: So we do not need to free previous requests in case of error here?

ClementKunz (approved):

PR #1632, opened 2026-03-16
feat(macOS): Synchro Settings modal

valentinperignon:
This PR introduces the sync configuration flow, which will be used during both onboarding and in the settings window.

For now, the flow is only wired up from the settings. File selection for synchronization is not yet implemented, as it will require AppKit.

**1. Synchro picker**
CleanShot 2026-03-20 at 11 40 33@2x
**2. Synchro configuration** 
CleanShot 2026-03-20 at 11 39 47@2x
**3. Selection of files to synchronize**
CleanShot 2026-03-20 at 11 40 06@2x


adrien-coye (approved):
Code looks Ok to me.
An issue still exists server side, but this PR is all good to me.

PR #1650, opened 2026-03-19
fix: crashes in KDC::CustomRollingFileAppender::append

ChristopheLarchier:
There may be a missing call to `log4cplus::threadCleanup()` when closing a thread.


luc-guyot-infomaniak (commented):
Looks good to me. Testing tomorrow.

luc-guyot-infomaniak (approved):

PR #1659, opened 2026-03-20
feat(winui): Update errors templates

herve-er:


src/gui4/windows/kDrive client/kDrive client/CustomControls/Errors/Templates/Node/NameLengthError.xaml.cs
```diff
@@ -17,5 +17,19 @@ public NameLengthError(Error error)
...
+                Utility.ShowUnexpectedErrorTeachingTip();
+            else if (!string.IsNullOrEmpty(Error.Path) && !await Utility.OpenFolderSecurely(System.IO.Path.Combine(Error.Sync.LocalPath, Error.Path)))
+                Utility.ShowUnexpectedErrorTeachingTip();
+            else if (string.IsNullOrEmpty(Error.RemoteNodeId) && string.IsNullOrEmpty(Error.Path))
+            {
+                Logger.Log(Logger.Level.Warning, $"Error {Error.DbId} has no RemoteNodeId or Path, cannot open in web or file explorer.");
+                Utility.ShowUnexpectedErrorTeachingTip();
+            }
+
+        }
```
ClementKunz, line 33: Why is this resolution more complex than, for example, the one in `ForbiddenCharOnlySpacesError`?
Moreover, could we imagine a parent class to all classes that only redirect to the webapp to open the file? That way, `ErrorCard_ActionClick` would be defined only once for this action.
(If feasible, this could be in another PR)

herve-er: This one is checked on both the local and remote file systems, so in some rare cases the user should rename the file locally and not remotely.

Even if I don't see how it could happen on Windows, it doesn't cost much to do this check.

For the base class, it is possible. I already put almost all the logic in `Error::OpenItemInWebViewAsync` to avoid a large part of the code duplication.

The only issue I see with having a base class is that it leads to a slightly more complex architecture / more friction if the UX evolves differently between cards, so it's a question of whether the benefit is worth it in this case

ClementKunz: Ok, let's keep it that way 👍

ClementKunz (approved):

PR #1662, opened 2026-03-23
refactor(cstypes): move GuiJobType enum from AbstractGuiJob to cstypes.h

R-Gld:
## Summary
The `GuiJobType` enum was nested inside `AbstractGuiJob` as an inner type, making it awkward to reference from files that should not depend on `AbstractGuiJob`. This PR promotes it to `KDC::GuiJobType` in `cstypes.h`, aligning it with other shared enums and making it accessible across the communication layer without pulling in the full job class hierarchy.

## Changes
- Add `GuiJobType` enum (Unknown, Query, Signal) to `src/libcommon/utility/cstypes.h` in the `KDC` namespace
- Remove the nested `GuiJobType` enum from `AbstractGuiJob` in `abstractguijob.h`
- Add a doc comment on `MsgType` in `comm.h` noting its relationship to the new `GuiJobType`
- Update all call sites in `commmanager.cpp` and `guicommserver_mac.mm` to use `GuiJobType::` instead of `AbstractGuiJob::GuiJobType::`
- Update all test files under `test/server/comm/guicommchannel/` to use the top-level `GuiJobType`

## Technical Details
`GuiJobType` is used to distinguish between Query and Signal jobs when dispatching messages through the GUI communication channel. Keeping it as a nested type of `AbstractGuiJob` forced unrelated files (such as `guicommserver_mac.mm` and test helpers) to include the full `abstractguijob.h` header just to reference this simple enum. Moving it to `cstypes.h` breaks this unnecessary coupling.

The integer values of the enum variants are unchanged, so the serialized wire format is not affected.

### DIFF
--- a/src/libcommon/utility/cstypes.h
+++ b/src/libcommon/utility/cstypes.h
@@ -20,6 +20,12 @@
 
 namespace KDC {
 
+enum class GuiJobType {
+    Unknown,
+    Query,
+    Signal
+};
+
 enum class UpdateState {
     UpToDate,
     Checking,
--- a/.sonarcloud.properties
+++ b/.sonarcloud.properties
@@ -1,0 +1,1 @@
+sonar.cpd.exclusions=src/gui4/macOS/kDriveCore/ServerBridge/XPC/**/*
--- a/src/libcommon/comm.h
+++ b/src/libcommon/comm.h
@@ -39,8 +39,8 @@
 #define EXECUTE_ERROR_MSG "C/S function call timeout or error!"
 
 /**
- * This enum is used in for old communication layer.
- * The new communication layer is using the one defined here: ./src/libcommon/utility/cstypes.h#GuiJobType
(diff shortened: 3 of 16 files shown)


src/libcommon/utility/cstypes.h
```diff
@@ -20,6 +20,12 @@
 
 namespace KDC {
 
+enum class GuiJobType {
```
adrien-coye, line 23: Since `gui4/macos` uses XPC we are able to get away without using this one. No refactor needed on the mac side.

herve-er (approved):

adrien-coye (commented):
LGTM 👍

src/libcommon/utility/cstypes.h
R-Gld: For the `gui4/linux`, this enum currently needs to be in libcommon rather than in the server to avoid duplicate declarations.

PR #1668, opened 2026-03-25
chore(logging): set up proper logging infrastructure for linux-v4 client

R-Gld:
## Summary
Wires the existing `libcommongui` Logger into the new Linux v4 client so that all
log output goes through the standard kDrive log pipeline instead of raw `qDebug`/`qFatal`
calls. Every call site in `IpcClient` and `main.cpp` is updated to use named Qt logging
categories, and `appclient.cpp` receives the same category-macro fix.

## Changes
- `main.cpp`: add `setupLogging()` that instantiates `KDC::Logger`, enables debug output,
  and opens the first log file; call it immediately after `QGuiApplication` is created
- `CMakeLists.txt`: link `libcommongui` to the v4 client executable (required by Logger)
- `IpcClient.cpp`: introduce `Q_LOGGING_CATEGORY(lcIpcClient, "gui.v4.ipc")` and replace
  all `qFatal()` calls with `qCCritical(lcIpcClient)` + `exit(EXIT_FAILURE)`; add
  `qCWarning` / `qCDebug` messages for `.comm` file read and socket connection steps
- `IpcClient.cpp`: fix `GuiJobType` extraction - use `convert<int>()` instead of
  `extract<int>()` to handle Poco type coercion correctly
- `IpcClient.cpp`: capture `this` in the `errorOccurred` lambda so `_socket->errorString()`
  is accessible in the error message
- `main.cpp` (debug block): switch `qDebug` to `qCDebug(lcMain)`, drop the redundant
  `[IpcClient]` prefix (now expressed by the category), and use the typed
  `KDC::GuiJobType` enum in the `messageReceived` lambda
- `appclient.cpp`: remove erroneous parentheses in `qCCritical(lcAppClient())` calls -
  logging category objects must not be called as functions


src/gui4/linux/CMakeLists.txt
```diff
@@ -12,6 +12,8 @@ qt_standard_project_setup(REQUIRES 6.8)
 
 set(GUI_SOURCE_FILES # C++ files for the GUI application
         src/main.cpp
+        src/AppClientLinux.cpp
```
ChristopheLarchier: Remove src folder, we are already in src/gui4/linux...
No upper case in file names.

ChristopheLarchier (approved):

PR #1669, opened 2026-03-25
feat(parameters): enable Sentry and Matomo by default

herve-er:


src/libcommon/info/parametersinfo.cpp
```diff
@@ -141,7 +141,7 @@ QDataStream &operator>>(QDataStream &in, ParametersInfo &parametersInfo) {
             parametersInfo._notificationsDisabled >> parametersInfo._useLog >> parametersInfo._logLevel >>
             parametersInfo._extendedLog >> parametersInfo._purgeOldLogs >> parametersInfo._darkTheme >>
             parametersInfo._dialogGeometry >> parametersInfo._maxAllowedCpu >> parametersInfo._proxyConfigInfo >>
-            parametersInfo._distributionChannel;
+            parametersInfo._distributionChannel >> parametersInfo._sentryEnabled >> parametersInfo._matomoEnabled;
```
ChristopheLarchier, line 144: Usefull? This is for the old comm layer.

herve-er: Without this change, each time the old client updates the parameters, it resets the Matomo and Sentry enabled fields to their default values. This happens because in aAppServer we do something like:

```cpp
case RequestNum::PARAMETERS_UPDATE:{
ParametersInfo parms; // Initialized with default values
stream >> parms;      // Previously did not overwrite Matomo and Sentry, leaving them at defaults
ServerRequest::updateParameters(parms); // Persists default states
...
}
```

ChristopheLarchier: OK

ChristopheLarchier (approved):

PR #1673, opened 2026-03-25
feat(macOS): Handle showNotification signal

valentinperignon:
CleanShot 2026-03-25 at 17 03 18@2x


adrien-coye (approved):
Just how I would have wanted to do it with a test. Good one 👍

PR #1678, opened 2026-03-26
feat(backend-api-v3): Migrates the remaining listing requests to the v3 backend API

luc-guyot-infomaniak:
- **(feat)** The version 3 of the full listing and continue listing requests is implemented.
- **(feat)** The `RemoteFileSystemObserver` now consumes the newly introduced `v3` requests.
- **(refactoring)** The unused `UserId` argument has been removed from the constructor of `AbstractNetworkJob`
-  Unit tests for the new columns introduced in `ParmsDb.cpp` (i.e., the special folders cursors) are implemented.

**Note.** The `ServerRequests` and the `ExecutorWorker` will be adapted in the next and final pull-request.


src/libparms/db/parmsdb.cpp
```diff
@@ -270,25 +270,34 @@
     "notificationsDisabled INTEGER,"                                                         \
     "hasFullyCompleted INTEGER,"                                                             \
     "navigationPaneClsid TEXT,"                                                              \
-    "listingCursor TEXT,"                                                                    \
-    "listingCursorTimestamp INTEGER,"                                                        \
+    "userPrivateFolderCursor TEXT,"                                                          \
+    "userPrivateFolderTimestamp INTEGER,"                                                    \
```
ClementKunz: The timestamp is the cursor's timestamp, not the folder's timestamp. Do you think that we can have a column name that reflects better this?

src/libparms/db/parmsdb.cpp
```diff
@@ -1152,34 +1176,38 @@ bool ParmsDb::upgradeTables() {
...
+        return false;
     }
 
     // Sync table
     tableName = "sync";
-    if (!addTextColumnIfMissing(tableName, "localNodeId")) {
-        return false;
-    }
+    if (!addTextColumnIfMissing(tableName, "localNodeId")) return false;
+    if (!addTextColumnIfMissing(tableName, "userPrivateFolderCursor")) return false;
```
ClementKunz, line 1197: So what happened to the columns `listingCursor` and `listingCursorTimestamp`? They just remain in the table but are not used anymore? Should we remove them?

src/libparms/db/sync.h
```diff
@@ -62,13 +61,31 @@ class PARMS_EXPORT Sync {
...
+            _cursorStore.commonDocumentsFolderCursor = {listingCursor, timestamp};
+        }
+        inline void setSharedFolderCursor(const Cursor &listingCursor, const Timestamp timestamp) {
+            _cursorStore.sharedFolderCursor = {listingCursor, timestamp};
         }
-        inline void listingCursor(std::string &listingCursor, int64_t &timestamp) const {
-            listingCursor = _listingCursor;
-            timestamp = _listingCursorTimestamp;
+
+        inline void userPrivateFolderCursor(Cursor &listingCursor, Timestamp &timestamp) const {
```
ClementKunz: Why not directly return a `CursorData` here? IMO it is more concise and more readable.

src/libsyncengine/jobs/network/kDrive_API/listing/abstractlistingjob.cpp
```diff
@@ -22,35 +22,41 @@
...
+    _apiVersion = 3;
+}
 
-AbstractListingJob::AbstractListingJob(const ApiType apiType, const DriveDbId driveDbId, const NodeSet &blacklist /*= {}*/) :
-    AbstractTokenNetworkJob(apiType, 0, 0, driveDbId, 0),
+AbstractListingJob::AbstractListingJob(const ApiType apiType, const DriveDbId driveDbId,
+                                       const RemoteNodeIdSet &blacklist /*= {}*/) :
+    AbstractTokenNetworkJob(apiType, 0, driveDbId, 0),
     _blacklist(blacklist) {
+    _apiVersion = 3;
```
ClementKunz, line 35: Shouldn't we set the value to 3 by default now in `abstractnetworkjob.h`

src/libsyncengine/jobs/network/kDrive_API/listing/continuefilelistwithcursorjob.cpp
```diff
@@ -20,14 +20,18 @@
...
 namespace KDC {
 
-ContinueFileListWithCursorJob::ContinueFileListWithCursorJob(const int driveDbId, const std::string &cursor,
-                                                             NodeSet blacklist /*= {}*/) :
+ContinueFileListWithCursorJob::ContinueFileListWithCursorJob(const int driveDbId, NodeId remoteDirId, std::string cursor,
+                                                             const NodeSet &blacklist /*= {}*/) :
     AbstractListingJob(driveDbId, blacklist),
-    _cursor(cursor) {}
+    _remoteDirId(std::move(remoteDirId)),
+    _cursor(std::move(cursor)) {}
```
ClementKunz, line 27: Why is this preferable to simply passing the reference? The copy is not avoided right?

ClementKunz: If I understand correctly, there is a small gain from passing by value then moving: https://stackoverflow.com/questions/51705967/advantages-of-pass-by-value-and-stdmove-over-pass-by-reference
But the downside is that the moved-from object remains in an unspecified state:
```
Although note that -in the standard library- moving implies that the moved-from object is left in a valid but unspecified state. Which means that, after such an operation, the value of the moved-from object should only be destroyed or assigned a new value; accessing it otherwise yields an unspecified value.
```

I would prioritize safety over speed here since this is not a method that is called very frequently (once every 50 seconds if there are changes).

src/libsyncengine/jobs/network/kDrive_API/listing/csvfullfilelistwithcursorjob.cpp
```diff
@@ -26,20 +29,25 @@ namespace KDC {
...
+
+CsvFullFileListWithCursorJob::CsvFullFileListWithCursorJob(const DriveDbId driveDbId, RemoteNodeId remoteDirId,
+                                                           const RemoteNodeIdSet &blacklist /*= {}*/,
+                                                           const Zip zip /*= Zip:On*/) :
     AbstractListingJob(driveDbId, blacklist),
-    _dirId(dirId),
+    _remoteDirId(std::move(remoteDirId)),
     _zip(zip),
-    _snapshotItemHandler(_logger) {
+    _snapshotItemHandler(userDbId(), driveId(), _logger) {
```
ClementKunz, line 39: I know it is a bit late for that, but all this translator thing seems quite complex for something temporary. It is modifying data structure and making the code more complex.
Are we sure we do not want to directly push towards a definitive solution that mimics the webapp layout?

src/libsyncengine/jobs/network/kDrive_API/listing/initfilelistwithcursorjob.cpp
```diff
@@ -22,20 +22,22 @@
 
 namespace KDC {
 
-InitFileListWithCursorJob::InitFileListWithCursorJob(const DriveDbId driveDbId, const NodeId &dirId) :
+InitFileListWithCursorJob::InitFileListWithCursorJob(const DriveDbId driveDbId, RemoteNodeId dirId) :
     AbstractListingJob(ApiType::Drive, driveDbId),
-    _dirId(dirId) {
+    _remoteDirId(std::move(dirId)) {
     _httpMethod = Poco::Net::HTTPRequest::HTTP_GET;
+    _apiVersion = 2;
```
ClementKunz, line 29: Is it normal that we are still in API v2 here?
Is it because this route is not used? In this case, should we keep this job?

src/libsyncengine/jobs/network/kDrive_API/apitranslator.h
```diff
@@ -27,34 +27,29 @@ class ApiTranslator {
...
 
         static RemoteNodeId v2RootFolderRemoteId();
 
-        using SpecialFolderNames = std::unordered_map<SpecialFolder, std::string>;
         static const SpecialFolderNames v3SpecialFolderNames;
 
     private:
         static bool getDriveDbIds(DriveDbIdMap &driveIdMap);
-        static ExitInfo updateCache(DriveDbId driveDbId);
+        static ExitInfo updateCache(UserDbId userDbId, DriveDbId driveDbId);
```
ClementKunz: We use `DriveId` everywhere else

src/libsyncengine/jobs/network/kDrive_API/getfilesindirectoryjob.cpp
```diff
@@ -27,9 +27,9 @@
 
 namespace KDC {
 
-void GetFilesInDirectoryJob::translateDriveDbIdFromV2ToV3(const TranslationMode translationMode) {
+void GetFilesInDirectoryJob::translateRemoteIdFromV2ToV3(const TranslationMode translationMode) {
```
ClementKunz: What's the point of giving `TranslationMode` as input since there is only 1 mode accepted (`V2ToV3`) and that the function name explicitly states it?

src/libsyncengine/jobs/network/kDrive_API/getfilesindirectoryjob.cpp
```diff
@@ -38,22 +38,22 @@ void GetFilesInDirectoryJob::translateDriveDbIdFromV2ToV3(const TranslationMode
...
                                                std::string cursorInput,
                                                const TranslationMode translationMode /* = TranslationMode::V2ToV3 */) :
-    AbstractTokenNetworkJob(ApiType::Drive, static_cast<int>(userDbId), 0, 0, static_cast<int>(driveId)),
+    AbstractTokenNetworkJob(ApiType::Drive, userDbId, 0, driveId),
     _fileId(std::move(fileId)),
     _cursorInput(std::move(cursorInput)) {
     _apiVersion = 3;
     _httpMethod = Poco::Net::HTTPRequest::HTTP_GET;
-    translateDriveDbIdFromV2ToV3(translationMode);
+    translateRemoteIdFromV2ToV3(translationMode);
```
ClementKunz: Another thing about this method: it is not clear what it is doing. Translate remote ID of what? Perhaps prefer a generic method that takes as input the ID to translate.

src/libsyncengine/syncpal/syncpal.cpp
```diff
@@ -905,19 +911,57 @@ ExitInfo SyncPal::setListingCursor(const std::string &value, int64_t timestamp)
     return ExitCode::Ok;
 }
 
-ExitInfo SyncPal::listingCursor(std::string &value, int64_t &timestamp) {
+ExitInfo SyncPal::setUserPrivateFolderCursor(const Cursor &listingCursor, const Timestamp timestamp) {
```
ClementKunz: Can we use `CursorData` here (and the other ones below)?

src/libsyncengine/update_detection/file_system_observer/remotefilesystemobserverworker.cpp
```diff
@@ -73,16 +76,29 @@ void RemoteFileSystemObserverWorker::execute() {
...
-        if (exitInfo = processEvents(); !exitInfo) {
-            LOG_SYNCPAL_DEBUG(_logger, "Error in processEvents: " << exitInfo);
+        std::vector<std::string> mainDirectoriesRemoteIds;
+        if (const auto mainDirectoriesExitInfo = getMainDirectoriesRemoteIds(mainDirectoriesRemoteIds);
+            !mainDirectoriesExitInfo) {
+            LOG_SYNCPAL_DEBUG(_logger, "Error in getMainDirectoriesRemoteIds: " << mainDirectoriesExitInfo);
             break;
         }
+        for (const auto &remoteDirId: mainDirectoriesRemoteIds) {
+            exitInfo = processEvents(remoteDirId);
```
ClementKunz, line 93: The `processEvents` method is blocking (at least with the actual implementation of the longpoll). So we will not monitor the modification on all main directories in parallel. Is this something that is planned to be handled in a subsequent PR?

src/libsyncengine/update_detection/file_system_observer/remotefilesystemobserverworker.cpp
```diff
@@ -171,31 +186,34 @@ ExitInfo RemoteFileSystemObserverWorker::processEvents() {
         }
 
         std::shared_ptr<ContinueFileListWithCursorJob> job = nullptr;
-        if (_cursor.empty()) {
-            LOG_SYNCPAL_WARN(_logger, "Cursor is empty for driveDbId=" << _driveDbId << ", invalidating snapshot");
+        if (_listingCursorMap.at(remoteDirId).empty()) {
+            LOG_SYNCPAL_WARN(_logger, "Cursor is empty for driveDbId=" << _driveDbId << ", invalidating remote snapshot.");
```
ClementKunz: The log should also mention the `remoteDirId`

src/libsyncengine/update_detection/file_system_observer/remotefilesystemobserverworker.cpp
```diff
@@ -215,293 +233,366 @@ ExitInfo RemoteFileSystemObserverWorker::processEvents() {
...
-                    break;
-                }
-            }
-
-            if (!JsonParserUtility::extractValue(dataObj, hasMoreKey, hasMore)) {
-                exitInfo = ExitCode::BackError;
-            }
+        std::string cursor;
+        if (!JsonParserUtility::extractValue(resObj, cursorKey, cursor)) {
+            exitInfo = ExitCode::BackError;
```
ClementKunz:
```suggestion
            exitInfo = {ExitCode::BackError, ExitCause::MissingReplyData};
```

src/libsyncengine/update_detection/file_system_observer/remotefilesystemobserverworker.cpp
```diff
@@ -215,293 +233,366 @@ ExitInfo RemoteFileSystemObserverWorker::processEvents() {
...
+            break;
+        }
 
-            // Look for new actions
-            exitInfo = processActions(dataObj->getArray(actionsKey));
+        if (cursor != _listingCursorMap[remoteDirId]) {
+            _listingCursorMap[remoteDirId] = cursor;
+            LOG_SYNCPAL_DEBUG(_logger, "Sync cursor updated: " << _listingCursorMap[remoteDirId]);
```
ClementKunz: We should specify for which directory

src/libsyncengine/update_detection/file_system_observer/remotefilesystemobserverworker.cpp
```diff
@@ -215,293 +233,366 @@ ExitInfo RemoteFileSystemObserverWorker::processEvents() {
...
             if (!exitInfo) {
-                LOG_SYNCPAL_WARN(_logger, "Error in RemoteFileSystemObserverWorker::processActions: " << exitInfo);
-                tryToInvalidateSnapshot();
+                LOG_SYNCPAL_WARN(_logger, "Error in RemoteFileSystemObserverWorker::saveListingCursor: " << exitInfo);
                 break;
             }
         }
+
+        if (!JsonParserUtility::extractValue(resObj, hasMoreKey, hasMore)) {
+            exitInfo = ExitCode::BackError;
```
ClementKunz:
```suggestion
            exitInfo = {ExitCode::BackError, ExitCause::MissingReplyData};
```

src/libsyncengine/update_detection/file_system_observer/remotefilesystemobserverworker.cpp
```diff
@@ -215,293 +233,366 @@ ExitInfo RemoteFileSystemObserverWorker::processEvents() {
...
+    if (stopAsked()) return ExitCode::Ok;
+
+    std::vector<std::string> mainDirectoriesRemoteIds;
+    if (const auto mainDirectoriesExitInfo = getMainDirectoriesRemoteIds(mainDirectoriesRemoteIds); !mainDirectoriesExitInfo) {
+        LOG_SYNCPAL_DEBUG(_logger, "Error in getMainDirectoriesRemoteIds: " << mainDirectoriesExitInfo);
+        return mainDirectoriesExitInfo;
+    }
+
+    ExitInfo exitInfo;
+    for (const auto &mainFolderRemoteId: mainDirectoriesRemoteIds) {
```
ClementKunz, line 331: This should be done in parallel. Maybe in a subsequent PR

src/libsyncengine/update_detection/file_system_observer/remotefilesystemobserverworker.cpp
```diff
@@ -215,293 +233,366 @@ ExitInfo RemoteFileSystemObserverWorker::processEvents() {
...
+    if (stopAsked()) return ExitCode::Ok;
+
+    return getItemsInDir(nodeId, CursorPersistence::None);
+}
+
+ExitInfo RemoteFileSystemObserverWorker::handleSnapshotItem(
+        const SnapshotItem &item, SyncNameSet &existingFiles, ParsingIterationState &iterationState,
+        sentry::pTraces::counterScoped::RFSOExploreItem &itemHandlingMonitor) {
+    if (iterationState.ignore) return ExitCode::Ok; // continue
+    if (iterationState.eof) return ExitCode::Ok; // break
```
ClementKunz: If the behavior changes, those comments might not be up-to-date anymore and lead to misunderstanding. Should be explicit on caller side

src/libsyncengine/jobs/network/kDrive_API/listing/csvfullfilelistwithcursorjob.cpp
ClementKunz: Discussed offline. The proposed implementation is the preferred one.

src/libsyncengine/update_detection/file_system_observer/remotefilesystemobserverworker.cpp
```diff
@@ -515,15 +606,80 @@ ExitInfo RemoteFileSystemObserverWorker::processActions(Poco::JSON::Array::Ptr a
...
+}
+} // namespace
+
+ExitInfo RemoteFileSystemObserverWorker::processActions(const Poco::JSON::Array::Ptr actionArray,
+                                                        const Poco::JSON::Array::Ptr actionsFilesArray) {
+    if (!actionArray) return ExitCode::Ok;
+
+    ActionInfoList actionInfoList;
+    if (const auto exitInfo = createActionInfoList(actionArray, actionInfoList); !exitInfo) return exitInfo;
+    if (const auto exitInfo = fillActionsFilesInfo(actionsFilesArray, actionInfoList); !exitInfo) return exitInfo;
```
ClementKunz, line 651: Maybe add a short explanation saying that we have to parse both objects `actions` and `actions_files` to extract all the required information.

src/libsyncengine/update_detection/file_system_observer/remotefilesystemobserverworker.cpp
```diff
@@ -847,4 +1022,134 @@ void RemoteFileSystemObserverWorker::countListingRequests() {
     _listingFullCounter++;
 }
 
+void RemoteFileSystemObserverWorker::ActionInfo::setPath(const KDC::SyncName &remotePath) {
+    SyncPath remotePath_{remotePath};
+    ApiTranslator::translateV3ToV2(remotePath_);
```
ClementKunz, line 1030: It is personal, but I feel like this notation is weird. Don't you think it is clearer to have an implementation like `translateV3ToV2(const SyncPath &v3path, SyncPath &v2path)`.
Is there a reason why you choose to modify the parameter directly?

src/libparms/db/parmsdb.cpp
luc-guyot-infomaniak: Addressed with f0555b39388e3c44b30cc31bc7f21a47ae4c7663.

src/libparms/db/parmsdb.cpp
luc-guyot-infomaniak: We have a task in our backlog about the deletion of obsolete columns in the application databases.
I am willing to pick it up once this PR is merged. Would it be Ok for you?

src/libsyncengine/jobs/network/kDrive_API/listing/continuefilelistwithcursorjob.cpp
luc-guyot-infomaniak: As discussed off-line, there is no risk with the proposed change as the parameter that is subsequently moved is first passed by value. There is no gain in performance neither. This syntax is encouraged by SonarCloud.

src/libparms/db/sync.h
luc-guyot-infomaniak: Done in f78f15d61e11cb408d5796213bc50978b1983a26.

src/libsyncengine/jobs/network/kDrive_API/listing/abstractlistingjob.cpp
luc-guyot-infomaniak: Yes, we should, indeed. This will be done in the final PR where all the requests are beeing migrated: this PR is only about the listing requests migration.

src/libsyncengine/jobs/network/kDrive_API/listing/initfilelistwithcursorjob.cpp
luc-guyot-infomaniak: Yes, it is normal: this request not depreciated and has not `v3` counterpart.

src/libsyncengine/jobs/network/kDrive_API/apitranslator.h
luc-guyot-infomaniak: Good catch, thanks! Fixed with bd1438a13095ea8fcdcfd18d40b31e6ba5112057.

src/libsyncengine/jobs/network/kDrive_API/getfilesindirectoryjob.cpp
luc-guyot-infomaniak: This is a subtle point that was left without explanation so far: the API translations rely on this specific listing request. Thus translation must be disabled when the request is needed solely for translation purpose so as to avoid an infinite series of nested listing requests. This is where we use the `TranslationMode::None`.

I have added a comment to clarify this point. I have also renamed the function to better reflect what it does. 
Done in 6df3fab8ef6c7103441180535ba9e87ca9ca9696.

src/libsyncengine/syncpal/syncpal.cpp
luc-guyot-infomaniak: Yes, done in f78f15d61e11cb408d5796213bc50978b1983a26.

src/libsyncengine/jobs/network/kDrive_API/getfilesindirectoryjob.cpp
luc-guyot-infomaniak: Renaming done in 6df3fab8ef6c7103441180535ba9e87ca9ca9696.

src/libsyncengine/update_detection/file_system_observer/remotefilesystemobserverworker.cpp
luc-guyot-infomaniak: I have created a follow-up card in our backlog so that we can discuss its priority.
Might be implemented in the next PR if it is deemed as important.

src/libsyncengine/update_detection/file_system_observer/remotefilesystemobserverworker.cpp
luc-guyot-infomaniak: Done here: 06782357d8e7da2952f443fcf75d415d667d5e94.

src/libsyncengine/update_detection/file_system_observer/remotefilesystemobserverworker.cpp
luc-guyot-infomaniak: Done here 06782357d8e7da2952f443fcf75d415d667d5e94.

src/libsyncengine/update_detection/file_system_observer/remotefilesystemobserverworker.cpp
luc-guyot-infomaniak: Done here bea88163975c1ed60d90007ab4e1cf1db34d5b78.

src/libsyncengine/update_detection/file_system_observer/remotefilesystemobserverworker.cpp
luc-guyot-infomaniak: Done here: bea88163975c1ed60d90007ab4e1cf1db34d5b78.

src/libsyncengine/jobs/network/kDrive_API/apitranslator.cpp
```diff
@@ -66,14 +65,14 @@ ExitInfo ApiTranslator::getDriveDbId(const DriveId driveId, DriveDbId &driveDbId
...
-ExitInfo ApiTranslator::updateCache(const DriveDbId driveDbId) {
+ExitInfo ApiTranslator::updateCache(const UserDbId userDbId, const DriveId driveId) {
     constexpr auto maxNumberOfItems = 1000;
 
     const RemoteNodeId remoteFolderId = v2RootFolderRemoteId();
     std::shared_ptr<GetAllFilesInDirectoryJob> fileListJob;
 
     try {
-        fileListJob = std::make_shared<GetAllFilesInDirectoryJob>(driveDbId, remoteFolderId, TranslationMode::None);
+        fileListJob = std::make_shared<GetAllFilesInDirectoryJob>(userDbId, driveId, remoteFolderId, TranslationMode::None);
```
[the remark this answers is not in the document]
luc-guyot-infomaniak: The translation mode is actually correct. The reasons why the remote ID is specified this way and why translation is deactivated in this case are now explained in a comment, see commit 0ab86e281dc74b619d5d291877519a98edb4c85a.

test/libparms/db/testparmsdb.cpp
```diff
@@ -267,6 +267,20 @@ SyncSetupData createSyncs() {
...
+
+    sync1.setCursorStore(cursorStore1);
+
+    CursorStore cursorStore2;
+    cursorStore1.userPrivateFolderCursor = CursorData{"aaa", 10};
+    cursorStore1.commonDocumentsFolderCursor = CursorData{"bbb", 20};
+    cursorStore1.sharedFolderCursor = CursorData{"ccc", 30};
+
+    sync1.setCursorStore(cursorStore1);
+
```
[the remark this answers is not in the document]
luc-guyot-infomaniak: Yes this is an actual mistake. It is now addressed by f78f15d61e11cb408d5796213bc50978b1983a26.

test/libsyncengine/test.cpp
```diff
@@ -59,8 +59,11 @@
 #endif
 
 namespace KDC {
+
+CPPUNIT_TEST_SUITE_REGISTRATION(TestApiTranslator);
 CPPUNIT_TEST_SUITE_REGISTRATION(TestOperationProcessor);
 CPPUNIT_TEST_SUITE_REGISTRATION(TestExclusionTemplateCache);
+
 CPPUNIT_TEST_SUITE_REGISTRATION(TestNetworkJobs);
 CPPUNIT_TEST_SUITE_REGISTRATION(TestApiTranslator);
```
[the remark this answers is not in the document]
luc-guyot-infomaniak: This is an actual mistake. It is now corrected by 845f8893a9d53a2ccb0bbd9c820a43e147204cca.

src/libsyncengine/update_detection/file_system_observer/remotefilesystemobserverworker.cpp
```diff
@@ -215,293 +233,366 @@ ExitInfo RemoteFileSystemObserverWorker::processEvents() {
...
+        LOG_SYNCPAL_WARN(_logger, "Error in SyncDb::node");
+        return ExitCode::DbError;
+    }
+
+    RemoteSnapshotItem remoteSnapshotItem(remoteNodeId);
+    const SyncTime now = std::chrono::steady_clock::now().time_since_epoch().count();
+    remoteSnapshotItem.setType(NodeType::Directory);
+    remoteSnapshotItem.setName(folderName);
+    remoteSnapshotItem.setCreatedAt(now);
+    remoteSnapshotItem.setLastModified(now);
```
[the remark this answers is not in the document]
luc-guyot-infomaniak: The default value `now` is computed as suggested in the fix 625fdee70ed9e8f3c3051ea7ff23e761f66d3693.

src/libsyncengine/update_detection/file_system_observer/remotefilesystemobserverworker.cpp
```diff
@@ -215,293 +233,366 @@ ExitInfo RemoteFileSystemObserverWorker::processEvents() {
...
-            if (stopAsked()) {
-                LOG_DEBUG(_logger, "Request " << notifyJob->jobId() << ": aborting LongPoll job");
-                notifyJob->abort();
+    return ExitCode::Ok;
 
-                return ExitCode::Ok;
-            }
+    // Actual implementation
+    changes = false;
+    if (!_liveSnapshot.isValid()) return ExitCode::Ok;
```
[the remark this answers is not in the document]
luc-guyot-infomaniak: @copilot This branch will be merged in a feature branch where the actual implementation of the long poll request will done subsequently. The proposed code as it is now, will not be merged into the `develop` branch. It will not be used by any running application.

src/libsyncengine/update_detection/file_system_observer/remotefilesystemobserverworker.cpp
luc-guyot-infomaniak: Good point, indeed! I have created a card in our backlog for this. I am willing to do achieve this parallelism once we have decided on its priority level.

src/libsyncengine/update_detection/file_system_observer/remotefilesystemobserverworker.cpp
luc-guyot-infomaniak: Thanks for noticing it. These were comments that I used to check that my refactoring was consistent with the previous implementation. They are now obsolete. Removed in 0cc901211ce8f3f7a6863a8fb84789aeea6b5462.

src/libsyncengine/update_detection/file_system_observer/remotefilesystemobserverworker.cpp
luc-guyot-infomaniak: Done here: 2c7103d526e5507ebc8a8f23febbc1ee1dfd905e

src/libsyncengine/update_detection/file_system_observer/remotefilesystemobserverworker.cpp
luc-guyot-infomaniak: I agree that it looks awkward in this context. The reason why this method takes only one argument is that it allows us to be consistent with the other translation methods of the same class.

src/libparms/db/parmsdb.cpp
ClementKunz: Yes perfect :)

src/libsyncengine/update_detection/file_system_observer/remotefilesystemobserverworker.cpp
ClementKunz: Can we apply the same logic in `extractActionInfo` and `extractActionFileInfo`

luc-guyot-infomaniak: Done in 9620423451982389a43bbf418c62f7701d0ea5c0.

src/libsyncengine/update_detection/file_system_observer/snapshot/snapshotitem.cpp
```diff
@@ -171,4 +164,29 @@ void SnapshotItem::removeAllChildren() {
...
+
+ExitInfo RemoteSnapshotItem::setId(const UserDbId userDbId, const DriveId driveId, const NodeId &id) {
+    _id = id;
+    if (const auto exitInfo = ApiTranslator::translateV3ToV2(userDbId, driveId, _id); !exitInfo) return exitInfo;
+    _lastChangeRevision = _snapshotRevisionHandler ? _snapshotRevisionHandler->nextVersion() : 0;
+
+    return ExitCode::Ok;
+}
+
+ExitInfo RemoteSnapshotItem::setParentId(const UserDbId userDbId, const DriveDbId driveId, const NodeId &newParentId) {
```
R-Gld: We use `DriveDbId` as the type for `driveId` in the parameters, but it is declared as `DriveId` in the [header file](https://github.com/Infomaniak/desktop-kDrive/blob/9620423451982389a43bbf418c62f7701d0ea5c0/src/libsyncengine/update_detection/file_system_observer/snapshot/snapshotitem.h#L115).

This compiles because both types are `using` aliases of the same underlying type (`int64_t`), but it is semantically incorrect and may lead to maintenance issues.


```suggestion
ExitInfo RemoteSnapshotItem::setParentId(const UserDbId userDbId, const DriveId driveId, const NodeId &newParentId) {
```

src/libsyncengine/jobs/network/kDrive_API/apitranslator.h
```diff
@@ -27,34 +27,29 @@ class ApiTranslator {
...
 
-        using SpecialFolderNames = std::unordered_map<SpecialFolder, std::string>;
         static const SpecialFolderNames v3SpecialFolderNames;
 
     private:
         static bool getDriveDbIds(DriveDbIdMap &driveIdMap);
-        static ExitInfo updateCache(DriveDbId driveDbId);
+        static ExitInfo updateCache(UserDbId userDbId, DriveId driveId);
 
         using RemoteNodeIdCacheMap = std::unordered_map<DriveDbId, RemoteNodeId>;
```
R-Gld:
```suggestion
        using RemoteNodeIdCacheMap = std::unordered_map<DriveId, RemoteNodeId>;
```

src/libsyncengine/jobs/network/kDrive_API/getallfilesindirectoryjob.cpp
```diff
@@ -23,13 +23,13 @@
 
 namespace KDC {
 
-GetAllFilesInDirectoryJob::GetAllFilesInDirectoryJob(const UserDbId userDbId, const DriveId driveId, NodeId fileId,
+GetAllFilesInDirectoryJob::GetAllFilesInDirectoryJob(const UserDbId userDbId, const DriveId driveId, RemoteNodeId remoteDirId,
```
R-Gld, line 26: The `.cpp` correctly uses `DriveId` here, but [the corresponding header (`getallfilesindirectoryjob.h`)](https://github.com/Infomaniak/desktop-kDrive/blob/9620423451982389a43bbf418c62f7701d0ea5c0/src/libsyncengine/jobs/network/kDrive_API/getallfilesindirectoryjob.h#L30) still declares this parameter as `DriveDbId`, it wasn't updated alongside this change.

As well as the others, it compile well but can generate confusion

src/libsyncengine/jobs/network/kDrive_API/getfilesindirectoryjob.cpp
ClementKunz: Discussed again offline. This is clearly not a blocking point in this PR and may be rediscussed later. If we want to implement this change, it should be the topic of a dedicated PR.

src/libsyncengine/jobs/network/kDrive_API/apitranslator.h
luc-guyot-infomaniak: Thanks! Fixed with e6c287c484dc210f8bc64e7a9ccb42884d04defa.

src/libsyncengine/jobs/network/kDrive_API/getallfilesindirectoryjob.cpp
luc-guyot-infomaniak: Thanks! Fixed with e6c287c484dc210f8bc64e7a9ccb42884d04defa.

src/libsyncengine/update_detection/file_system_observer/snapshot/snapshotitem.cpp
luc-guyot-infomaniak: Thanks! Fixed with 25a47fd5ebbc2c3898c301011f930945bd087d7a.

ClementKunz (approved):

PR #1685, opened 2026-03-27
fix(macOS UI): App crashes when opening settings several times

valentinperignon:
The app crashes if you open, close then re-open the settings window


adrien-coye (approved):
Code LGTM, I did not try to reproduce.

PR #1695, opened 2026-03-27
feat(macOS): Connect Sentry

valentinperignon:


adrien-coye (approved):
Dsicussed, good as is.

PR #1702, opened 2026-03-30
fix: Dehydrated file blacklisted then deleted after EDIT on the drive (step 2)

ChristopheLarchier:


src/libsyncengine/update_detection/file_system_observer/localfilesystemobserverworker.cpp
```diff
@@ -521,7 +521,15 @@ ExitCode LocalFileSystemObserverWorker::isEditValid(const NodeId &nodeId, const
...
 
     if (vfsStatus.isPlaceholder && !vfsStatus.isHydrated) {
-        // Check if it is a metadata update
+        // Check if the local Edit operation is caused by the current sync (i.e. check in syncOps)
+        const auto relativePath = CommonUtility::relativePath(_syncPal->localPath(), path);
+        if (_syncPal->_syncOps->isLocalEditCausedBySync(nodeId, _syncPal->localPath(), relativePath, lastModifiedLocal,
+                                                        sizeLocal)) {
+            valid = false;
+            return ExitCode::Ok;
+        }
```
ClementKunz, line 530: I am not a fan of this change. The file system observer should be completely independent of other sync steps or state. Therefore, it should not use `SyncOps` that are generated for the propagation step.

src/libsyncengine/update_detection/file_system_observer/computefsoperationworker.cpp
```diff
@@ -290,13 +287,14 @@ ExitCode ComputeFSOperationWorker::inferChangeFromDbNode(const ReplicaSide side,
     const SyncTime snapshotModificationTime = snapshot->lastModified(nodeId);
     // On FAT filesystems, the time resolution for modification time is 2 seconds. Therefor, we ignore EDIT operations if the
     // difference between modification time in DB and the one on the filesystem is less or equal to 1sec.
-    const bool modifiedTimeDiffIsEnough = abs(snapshotModificationTime - dbModificationTime) > _timeDifferenceThresholdForEdit;
+    const bool sameModifiedTime =
+            CommonUtility::modificationTimesAreEqual(_syncPal->localPath(), snapshotModificationTime, dbModificationTime);
```
ClementKunz, line 291: Is this related to the issue being fixed?

src/libsyncengine/reconciliation/syncoperation.cpp
```diff
@@ -58,61 +59,121 @@ void SyncOperationList::setOpList(const std::list<SyncOpPtr> &opList) {
...
-    for (const auto opId: _node2op[nodeId]) {
+    for (const auto opId: _nodeIdSource2ops[nodeId]) {
         const auto op = _allOps[opId];
         // Keep the op only if its source side is the same as `side`.
         if (op && otherSide(op->targetSide()) == side) opList.push_back(opId);
     }
     return opList;
 }
 
+std::vector<SyncOpPtr> SyncOperationList::getOpsFromTargetNodeId(const NodeId &nodeId, ReplicaSide side, OperationType type,
```
ClementKunz, line 88: Why return a vector here? There should be only 1 operation for a given node ID, side and operation type.

ClementKunz:
Maybe we can handle Sonar's issue about constness now

PR #1703, opened 2026-03-30
feat(macOS UI): Show status bar icon

valentinperignon:
CleanShot 2026-03-31 at 09 54 30@2x


adrien-coye (approved):
I did review the code, tested it, and discussed in DM the future need to hide the app from the Dock and how to kill the app.

PR #1709, opened 2026-03-30
fix(sync): fix sync item removal and add progress callback

herve-er:
Update sync item removal logic to also retain items with Syncing status. Add per-job progress callback in executorworker to update SyncPal and log warnings on failure.


ClementKunz:
Maybe the SonarCloud issue could addressed

ClementKunz (approved):

PR #1712, opened 2026-04-01
feat(macOS): Add status bar right click

valentinperignon:
CleanShot 2026-04-01 at 10 05 33@2x


adrien-coye (approved):
New menu is looking good.
I did update the quit action to also quit the app.
I made sure to factorise that when the app exits it first asks the server to quit as well.

PR #1713, opened 2026-04-01
fix: handle unavailable content length for download progress

herve-er:


ClementKunz (approved):

adrien-coye (commented):
This is a non-breaking change for the Mac GUI4, and our support will come later. I'm fine with this PR 👍

ChristopheLarchier (approved):

PR #1715, opened 2026-04-01
feat(macOS): Present correct onboarding animation according to OS version

valentinperignon:


adrien-coye (approved):
LGTM as well. I kept a note to double-check the onboarding on an old macOS before the public beta.

PR #1718, opened 2026-04-01
feat(linux): add ServerCommService to handle server-initiated signals

R-Gld:
## Summary
Introduces `ServerCommService`, a new Qt service class that registers
handlers on `SignalDispatcher` for all server-initiated IPC signals and
re-emits them as typed Qt signals for consumption by app models and QML.
As a prerequisite, all signal job files in `src/server/comm/guijobs/`
are migrated to use shared `MSG_PARAM_*` string constants defined in
`comm.h`, replacing the previous file-local `static const auto` keys.

## Changes
- Add `ServerCommService` (`app/services/servercommservice.cpp/.h`):
  registers handlers for user, account, drive, sync, error, updater, and
  utility signals, and exposes them as typed Qt signals
- Declare 24 `MSG_PARAM_*` inline constexpr string constants in
  `src/libcommon/comm.h` to serve as the single source of truth for IPC
  payload parameter keys
- Replace file-local parameter key variables with `MSG_PARAM_*` constants
  across all 20+ signal job `.cpp` files in `src/server/comm/guijobs/`
- Instantiate `ServerCommService` as a member of `AppClientLinux` and
  wire its `quit` signal to `QCoreApplication::quit`
- Register `servercommservice.cpp/.h` in `CMakeLists.txt`

## Technical Details
The migration of server-side job files to `MSG_PARAM_*` constants is a
prerequisite: the GUI handler and the server serializer must agree on the
same key strings, and having them defined in one place (`comm.h`) removes
the previous implicit contract maintained by duplicated string literals.

### DIFF
--- a/src/gui4/linux/app/services/servercommservice.cpp
+++ b/src/gui4/linux/app/services/servercommservice.cpp
@@ -1,0 +1,98 @@
+/*
+ * Infomaniak kDrive - Desktop
+ * Copyright (C) 2023-2026 Infomaniak Network SA
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "servercommservice.h"
+
+#include "libcommon/utility/utility.h"
+
+#include <QLoggingCategory>
+
+Q_LOGGING_CATEGORY(lcServerCommService, "gui.v4.servercommservice", QtInfoMsg)
+
+namespace KDC {
+
+// Param keys ─ mirrors the server-side signal job static constants.
+namespace {
+// User
+constexpr auto kUserInfo = "userInfo";
+constexpr auto kUserDbId = "userDbId";
+// Account
+constexpr auto kAccountInfo = "accountInfo";
+constexpr auto kAccountDbId = "accountDbId";
+// Drive
+constexpr auto kDriveInfo = "driveInfo";
+constexpr auto kDriveDbId = "driveDbId";
+// Sync
+constexpr auto kSyncInfo = "syncInfo";
+constexpr auto kSyncDbId = "syncDbId";
+constexpr auto kSyncStatus = "syncStatus";
+constexpr auto kSyncStep = "syncStep";
+constexpr auto kSyncProgress = "SyncProgress";
+constexpr auto kCurrentFile = "currentFile";
+constexpr auto kTotalFiles = "totalFiles";
+constexpr auto kCompletedSize = "completedSize";
+constexpr auto kTotalSize = "totalSize";
+constexpr auto kEstimatedRemainingTime = "estimatedRemainingTime";
+constexpr auto kItemInfo = "itemInfo";
+// Error
+constexpr auto kErrorInfo = "errorInfo";
+constexpr auto kErrorDbId = "errorDbId";
+// Updater
+constexpr auto kUpdateState = "updateState";
+constexpr auto kVersionInfo = "versionInfo";
+// Utility
+constexpr auto kTitle = "title";
+constexpr auto kMessage = "message";
+constexpr auto kLogUploadState = "state";
+constexpr auto kPercentage = "percentage";
+} // namespace
+
+ServerCommService::ServerCommService(SignalDispatcher &dispatcher, QObject *parent) :
+    QObject(parent) {
+    registerUserHandlers(dispatcher);
+    registerAccountHandlers(dispatcher);
+    registerDriveHandlers(dispatcher);
+    registerSyncHandlers(dispatcher);
+    registerErrorHandlers(dispatcher);
+    registerUpdaterHandlers(dispatcher);
+    registerUtilityHandlers(dispatcher);
+}
+
+// -- User ---------------------------------------------------------------------
+
+void ServerCommService::registerUserHandlers(SignalDispatcher &dispatcher) {
+    dispatcher.registerHandler(SignalNum::USER_ADDED, [this](const Poco::DynamicStruct &params) {
+        UserInfo info;
+        info.fromDynamicStruct(params[kUserInfo].extract<Poco::DynamicStruct>());
+        emit userAdded(info);
+    });
+
+    dispatcher.registerHandler(SignalNum::USER_UPDATED, [this](const Poco::DynamicStruct &params) {
+        UserInfo info;
+        info.fromDynamicStruct(params[kUserInfo].extract<Poco::DynamicStruct>());
+        emit userUpdated(info);
+    });
+
+    dispatcher.registerHandler(SignalNum::USER_REMOVED, [this](const Poco::DynamicStruct &params) {
+        UserDbId userDbId = 0;
+        CommonUtility::readValueFromStruct(params, kUserDbId, userDbId);
+        emit userRemoved(userDbId);
+    });
+}
+
+} // namespace KDC
--- a/src/gui4/linux/app/services/servercommservice.cpp
+++ b/src/gui4/linux/app/services/servercommservice.cpp
@@ -22,11 +22,24 @@
 #include "libcommon/utility/utility.h"
 
 #include <QLoggingCategory>
+#include <QPointer>
 
 Q_LOGGING_CATEGORY(lcServerCommService, "gui.v4.servercommservice", QtInfoMsg)
 
-namespace KDC {
+namespace {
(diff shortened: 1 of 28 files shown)


src/gui4/linux/app/services/servercommservice.cpp
```diff
@@ -0,0 +1,284 @@
...
+// -- User ---------------------------------------------------------------------
+
+void ServerCommService::registerUserHandlers(SignalDispatcher &dispatcher) {
+    dispatcher.registerHandler(SignalNum::USER_ADDED,
+                               [self = QPointer<ServerCommService>(this)](const Poco::DynamicStruct &params) {
+                                   if (checkSelf(self, SignalNum::USER_ADDED)) return;
+                                   UserInfo info;
+                                   info.fromDynamicStruct(params[MSG_PARAM_USER_INFO].extract<Poco::DynamicStruct>());
+                                   emit self->userAdded(info);
+                               });
```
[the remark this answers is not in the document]
R-Gld: every dispatch is wrapped between a try an d a catch.

src/gui4/linux/app/services/servercommservice.cpp
```diff
@@ -0,0 +1,284 @@
...
+                                   LogUploadState state = LogUploadState::None;
+                                   int32_t percentage = 0;
+                                   CommonUtility::readValueFromStruct(params, MSG_PARAM_LOG_UPLOAD_STATE, state);
+                                   CommonUtility::readValueFromStruct(params, MSG_PARAM_PERCENTAGE, percentage);
+                                   emit self->logUploadStatusUpdated(state, percentage);
+                               });
+
+    dispatcher.registerHandler(SignalNum::UTILITY_QUIT, [self = QPointer<ServerCommService>(this)](const Poco::DynamicStruct &) {
+        if (checkSelf(self, SignalNum::UTILITY_QUIT)) return;
+        emit self->quit();
```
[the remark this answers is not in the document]
R-Gld: we are on the main thread

src/gui4/linux/app/services/servercommservice.cpp
```diff
@@ -0,0 +1,284 @@
...
+}
+
+// -- User ---------------------------------------------------------------------
+
+void ServerCommService::registerUserHandlers(SignalDispatcher &dispatcher) {
+    dispatcher.registerHandler(SignalNum::USER_ADDED,
+                               [self = QPointer<ServerCommService>(this)](const Poco::DynamicStruct &params) {
+                                   if (checkSelf(self, SignalNum::USER_ADDED)) return;
+                                   UserInfo info;
+                                   info.fromDynamicStruct(params[MSG_PARAM_USER_INFO].extract<Poco::DynamicStruct>());
```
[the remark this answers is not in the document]
R-Gld: No. Here, params is a const DynamicStruct & so it throw a NotFoundException.

src/gui4/linux/app/services/servercommservice.cpp
```diff
@@ -0,0 +1,284 @@
...
+                                   emit self->userAdded(info);
```
ChristopheLarchier: The self pointer could be destroyed since the call to checkSelf.
I suggest to use a QWeakPointer (and QSharedPointer instead of QPointer).
i.e. ServerCommService should inherit of QEnableSharedFromThis and use sharedFromThis() instead of this.

src/gui4/linux/app/services/servercommservice.cpp
```diff
@@ -0,0 +1,284 @@
...
+#include "libcommon/utility/utility.h"
+
+#include <QLoggingCategory>
+#include <QPointer>
+
+Q_LOGGING_CATEGORY(lcServerCommService, "gui.v4.servercommservice", QtInfoMsg)
+
+namespace {
+
+bool checkSelf(const QPointer<KDC::ServerCommService> &self, SignalNum num) {
```
ChristopheLarchier: Rename to something like isSelfInvalid.

src/gui4/linux/app/services/servercommservice.cpp
R-Gld: Since everything runs on the main Qt event loop, is there really a window between `checkSelf` and the `emit` where the object could be destroyed ?

src/gui4/linux/appclientlinux.cpp
```diff
@@ -43,6 +43,7 @@ AppClientLinux::AppClientLinux(int &argc, char **argv) :
     (void) connect(&_ipcClient, &IpcClient::connected, this, &AppClientLinux::ipcConnected);
     (void) connect(&_ipcClient, &IpcClient::disconnected, this, &AppClientLinux::ipcDisconnected);
     (void) connect(&_ipcClient, &IpcClient::serverSignalReceived, &_signalDispatcher, &SignalDispatcher::dispatch);
+    (void) connect(&_serverCommService, &ServerCommService::quit, this, &QCoreApplication::quit);
```
ChristopheLarchier: Don't you want to do the opposite?

src/gui4/linux/app/services/servercommservice.cpp
R-Gld: Done, renamed to `isSelfInvalid` https://github.com/Infomaniak/desktop-kDrive/pull/1718/commits/08366665e6ebc43073fb1e25bd4d342e4af724dc

Thanks for the suggestion.

src/libcommon/comm.h
```diff
@@ -39,6 +39,40 @@ inline constexpr char MSG_SIGNAL_ID[] = "id";
 inline constexpr char MSG_SIGNAL_NUM[] = "num";
 inline constexpr char MSG_SIGNAL_PARAMS[] = "params";
 
+// Signal / request–response payload parameter keys.
+// Shared between server-side jobs (src/server/comm/guijobs/) and GUI clients.
+inline constexpr char MSG_PARAM_USER_INFO[] = "userInfo";
```
ChristopheLarchier, line 44: We usually use the camelcase syntax for constants.

src/gui4/linux/app/services/servercommservice.cpp
ChristopheLarchier: Probably not.
In any case, _serverCommService is destroyed by the destructor of AppClientLinux so the event queue will be destroyed also, so no slot will be called, so isSelfInvalid will never be true (I think).
In practice, you've seen some problems?

src/gui4/linux/appclientlinux.cpp
R-Gld: You are right, the signal UTILITY_QUIT is not sent by the server when he quit.
```suggestion
```

src/libcommon/comm.h
R-Gld: Yes, the names are inherited from the old `#define` syntax. I'll rename them in another PR to avoid flooding this one.

src/gui4/linux/app/services/servercommservice.cpp
```diff
@@ -0,0 +1,284 @@
...
+            SignalNum::SYNC_PROGRESSINFO, [self = QPointer<ServerCommService>(this)](const Poco::DynamicStruct &params) {
+                if (isSelfInvalid(self, SignalNum::SYNC_PROGRESSINFO)) return;
+                SyncDbId syncDbId = 0;
+                auto status = SyncStatus::Undefined;
+                auto step = SyncStep::Idle;
+                CommonUtility::readValueFromStruct(params, MSG_PARAM_SYNC_DB_ID, syncDbId);
+                CommonUtility::readValueFromStruct(params, MSG_PARAM_SYNC_STATUS, status);
+                CommonUtility::readValueFromStruct(params, MSG_PARAM_SYNC_STEP, step);
+
+                const Poco::DynamicStruct progressStruct = params[MSG_PARAM_SYNC_PROGRESS].extract<Poco::DynamicStruct>();
```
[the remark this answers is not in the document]
R-Gld: every callback is executed in a try catch block

src/gui4/linux/app/services/servercommservice.cpp
```diff
@@ -0,0 +1,284 @@
...
+
+// -- User ---------------------------------------------------------------------
+
+void ServerCommService::registerUserHandlers(SignalDispatcher &dispatcher) {
+    dispatcher.registerHandler(SignalNum::USER_ADDED,
+                               [self = QPointer<ServerCommService>(this)](const Poco::DynamicStruct &params) {
+                                   if (isSelfInvalid(self, SignalNum::USER_ADDED)) return;
+                                   UserInfo info;
+                                   info.fromDynamicStruct(params[MSG_PARAM_USER_INFO].extract<Poco::DynamicStruct>());
+                                   emit self->userAdded(info);
```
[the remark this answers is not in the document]
R-Gld: everything is in the main qt thread.

src/gui4/linux/app/services/servercommservice.cpp
R-Gld: Yeah I think you are right, I haven't seen any problem in practice. I'll remove the `QPointer` and `isSelfInvalid` and use `this` directly as I did before.

R-Gld: Done here: https://github.com/Infomaniak/desktop-kDrive/pull/1718/commits/7404b917dfa8138a3b5168471df4dfc8616fba65

ChristopheLarchier (approved):

PR #1728, opened 2026-04-07
feat(macOS UI): Show permissions when necessary

valentinperignon:


adrien-coye (approved):
I did test this to confirm it works as expected.

PR #1729, opened 2026-04-07
fix(sync): restore missing sync add and drive search response fields

R-Gld:
## Summary
This PR fixes two server-side response contract issues in GUI jobs.

First, `AbstractSyncAddJob` now assigns the created `SyncInfo` to the member
that is actually serialized back to the client. Without that assignment, the
JSON response for sync creation could return an empty or stale `syncInfo`
payload even when the sync creation itself succeeded.

Second, `DriveSearchJob` now propagates the `hasMore` flag returned by
`SearchJob`. Until now, the JSON job always serialized `false`, which made the
search pagination state incorrect for Linux v4 and any other client consuming
the JSON contract.

## Changes
- assign `_syncInfo = syncInfo` at the end of
  `src/server/comm/guijobs/abstractsyncaddjob.cpp`
- assign `_hasMore = searchJob.hasMore()` in
  `src/server/comm/guijobs/drivesearchjob.cpp`
- update the copyright year in the two touched files

## Why
- `SYNC_ADD` / `SYNC_ADD2` responses are expected to return a valid `syncInfo`
  object to the caller
- `DRIVE_SEARCH` responses are expected to expose whether additional pages are
  available
- both bugs were in the server job layer: the underlying work completed, but
  part of the response state was never copied into the serialized output

### DIFF
--- a/src/server/comm/guijobs/abstractsyncaddjob.cpp
+++ b/src/server/comm/guijobs/abstractsyncaddjob.cpp
@@ -110,6 +110,7 @@ ExitInfo AbstractSyncAddJob::process(SyncInfo &syncInfo) {
     Utility::restartFinderExtension();
 #endif
 
+    _syncInfo = syncInfo;
     return ExitCode::Ok;
 }
 
--- a/src/server/comm/guijobs/abstractsyncaddjob.cpp
+++ b/src/server/comm/guijobs/abstractsyncaddjob.cpp
@@ -1,6 +1,6 @@
 /*
  * Infomaniak kDrive - Desktop
- * Copyright (C) 2023-2025 Infomaniak Network SA
+ * Copyright (C) 2023-2026 Infomaniak Network SA
  *
  * This program is free software: you can redistribute it and/or modify
  * it under the terms of the GNU General Public License as published by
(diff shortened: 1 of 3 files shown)


src/server/comm/guijobs/abstractsyncaddjob.cpp
ClementKunz: `syncInfo` is not modified. Can we make it const?

R-Gld: Fixed here: https://github.com/Infomaniak/desktop-kDrive/pull/1729/commits/142632a516ace68eaccb4bb8239a277601ca7304

src/server/comm/guijobs/drivesearchjob.cpp
```diff
@@ -85,6 +85,7 @@ ExitInfo DriveSearchJob::process() {
     // Send search request (synchronously for now)
     SearchJob searchJob(sync.driveDbId(), _syncDbId, CommonUtility::commString2Str(_searchString));
     (void) searchJob.runSynchronously();
```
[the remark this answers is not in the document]
R-Gld: The goal of this PR is only to fix the response fields.

ClementKunz (approved):

PR #1751, opened 2026-04-13
feat(winui): handle app version lock

herve-er:


src/gui4/windows/kDrive client/kDrive client/ServerCommunication/Services/ServerCommService.cs
```diff
@@ -990,23 +991,28 @@ public async Task<bool> StartUpdate(CancellationToken cancellationToken)
...
             // First check the update state
-            CommData data = await _commClient.SendRequestAsync(RequestNum.UPDATER_STATE, [], cancellationToken);
-            if (!CheckJobResultAndLogIfError(data))
-                return false;
-
-            if (!HasRequiredParam(data, JsonKeys.UpdateState))
-                return false;
-
-            UpdateState? updateState = (UpdateState?)data.Params[JsonKeys.UpdateState]?.GetValue<int>();
             if (updateState is null)
```
ClementKunz, line 997: Don't we want to send the request anyway? Maybe the state has changed since last check

src/gui4/windows/kDrive client/kDrive client/kDrive client.csproj
```diff
@@ -1,6 +1,6 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <AssemblyVersion>4.0.0.2</AssemblyVersion>
+    <AssemblyVersion>3.8.0.1</AssemblyVersion>
```
ClementKunz: Is this normal? The version for the new GUI should be `4.0.0` no?

version.json
```diff
@@ -2,7 +2,7 @@
   "Version": {
     "major": 3,
     "minor": 8,
-    "patch": 4,
+    "patch": 0,
```
ClementKunz: Why?

src/server/comm/guijobs/utilityactivateloadinfojob.cpp
```diff
@@ -38,9 +39,46 @@ ExitInfo UtilityActivateLoadInfoJob::serializeOutputParms() {
...
+    if (ExitInfo exitInfo = checkUpdateIfNeeded(); !exitInfo) {
+        return exitInfo;
+    }
+
     _commManager->appServer().triggerSyncProgressUpdate();
     _commManager->appServer().loadUsersInfo();
     return ExitCode::Ok;
 }
 
+ExitInfo UtilityActivateLoadInfoJob::checkUpdateIfNeeded() {
```
ClementKunz, line 51: Is this job triggered often?
What happens if there are hundreds of errors? Could this lead to lantency?

src/gui4/windows/kDrive client/kDrive client/ServerCommunication/Services/ServerCommService.cs
herve-er: RefreshUpdaterVersionInfo can be called when we receive the `UPDATER_STATE_CHANGED` signal, which already includes the `UpdateState`, making it unnecessary to query the server again for it. However, this state is not preserved across calls.

src/gui4/windows/kDrive client/kDrive client/kDrive client.csproj
herve-er: This field is updated automatically from `version.json` at each build. I will revert it to avoid ping-pong changes between builds -> https://github.com/Infomaniak/desktop-kDrive/pull/1751/commits/2aeeaf336057a8b9db68068ca61fe0a846b85697

version.json
herve-er: I ran some tests. I shouldn’t have committed this. It’s fixed here. https://github.com/Infomaniak/desktop-kDrive/pull/1751/commits/2aeeaf336057a8b9db68068ca61fe0a846b85697

src/server/comm/guijobs/utilityactivateloadinfojob.cpp
herve-er: This request is called each time the user opens the app window.

It runs fully in the background without blocking anything.

This job also refreshes all users/drives info, which involves several network calls, making the database query quite negligible, even with thousands of errors. (Currently, the job already takes between a few seconds and up to 90 seconds, as we often encounter timeouts from the login API.)

ClementKunz (approved):

PR #1752, opened 2026-04-13
fix: sleep(200) in quick loop to reduce CPU utilization

RaphaelOstier:
# changes :
- sleep(200) in quick loop to reduce CPU utilization

### DIFF
--- a/src/libsyncengine/propagation/executor/executorworker.cpp
+++ b/src/libsyncengine/propagation/executor/executorworker.cpp
@@ -1307,6 +1307,7 @@ ExitInfo ExecutorWorker::waitForAllJobsToFinish() {
             cancelAllOngoingJobs();
             return exitInfo;
         }
+        Utility::msleep(LOOP_EXEC_SLEEP_PERIOD);
     }
     return ExitCode::Ok;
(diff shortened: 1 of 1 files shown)


src/libsyncengine/propagation/executor/executorworker.cpp
```diff
@@ -1307,6 +1307,7 @@ ExitInfo ExecutorWorker::waitForAllJobsToFinish() {
             cancelAllOngoingJobs();
             return exitInfo;
         }
+        std::this_thread::sleep_for(std::chrono::milliseconds(500));
```
herve-er: 500 ms is a bit overkill; 100 ms should be enough

```suggestion
        std::this_thread::sleep_for(std::chrono::milliseconds(100));
```

luc-guyot-infomaniak: **Suggestion.** Use the suggestive and less verbose `Utility::msleep(LOOP_EXEC_SLEEP_PERIOD)` with possibly a shorter duration.

herve-er (approved):

PR #1760, opened 2026-04-15
feat(checksum): checksum verifications of newly downloaded installer

RaphaelOstier:
depends on #1749
# checksum verification
- implemented a checksum of the installer to verify the integrity of the file
- moved the signature verification of the installer into correct place
- sentry logs if installer is corrupted

### DIFF
--- a/src/server/updater/windowsupdater.cpp
+++ b/src/server/updater/windowsupdater.cpp
@@ -187,7 +187,9 @@ bool WindowsUpdater::verifyFileChecksum(const SyncPath &filepath) {
         if (ioError == IoError::Success) {
             LOGW_INFO(Log::instance()->getLogger(), L"corrupted file at " << Utility::formatSyncPath(filepath) << L" deleted");
         } else {
-            LOGW_WARN(Log::instance()->getLogger(), L"couldn't reach corrupted file at " << Utility::formatSyncPath(filepath));
+            LOGW_WARN(Log::instance()->getLogger(), L"couldn't reach corrupted file at " << Utility::formatSyncPath(filepath)
+                                                                                         << L" : IOError state "
+                                                                                         << static_cast<int>(ioError));
         }
 
         // Send to Sentry
--- a/src/server/updater/windowsupdater.cpp
+++ b/src/server/updater/windowsupdater.cpp
@@ -191,8 +191,7 @@ bool WindowsUpdater::verifyFileChecksum(const SyncPath &filepath) {
         }
 
         // Send to Sentry
-        KDC::sentry::Handler::captureMessage(KDC::sentry::Level::Error,
-                                             "Updater::verifyChecksum::" + reason,
+        KDC::sentry::Handler::captureMessage(KDC::sentry::Level::Error, "Updater::verifyChecksum::" + reason,
                                              "Checksum verification failed: " + reason);
         return false;
     };
@@ -239,10 +238,10 @@ std::string WindowsUpdater::computeFileChecksum(const SyncPath &filepath) {
 
     std::array<char, 8192> buffer{};
     while (file.read(buffer.data(), buffer.size())) {
-        SHA256_Update(&sha256, buffer.data(), static_cast<std::size_t>(file.gcount()));
+        (void) SHA256_Update(&sha256, buffer.data(), static_cast<std::size_t>(file.gcount()));
     }
     // Process remaining bytes
-    SHA256_Update(&sha256, buffer.data(), static_cast<std::size_t>(file.gcount()));
+    (void) SHA256_Update(&sha256, buffer.data(), static_cast<std::size_t>(file.gcount()));
 
     std::array<std::uint8_t, SHA256_DIGEST_LENGTH> hash{};
     if (SHA256_Final(hash.data(), &sha256) != 1) {
--- a/src/server/updater/windowsupdater.cpp
+++ b/src/server/updater/windowsupdater.cpp
@@ -67,17 +67,11 @@ void WindowsUpdater::startInstaller() {
     }
     if (std::error_code ec; !std::filesystem::exists(filepath, ec)) {
         LOGW_WARN(Log::instance()->getLogger(), L"Installer file not found. " << Utility::formatStdError(filepath, ec));
-        if (_autoUpdate) {
-            LOG_ERROR(Log::instance()->getLogger(), "Already tried to re-download the installer before.");
-            setState(UpdateState::UpdateError);
-            return;
-        }
-        _autoUpdate = true;
-        downloadUpdate();
+        retryDownload(filepath);
         return;
     }
     if (!verifyFileChecksum(filepath)) {
-        setState(UpdateState::UpdateError); // Or create UpdateState::ChecksumError
+        retryDownload(filepath);
         return;
     }
 
@@ -271,6 +265,16 @@ bool WindowsUpdater::verifyDigitalSignature(const SyncPath &filepath) {
         return false;
     }
     return true;
+}
+
+void WindowsUpdater::retryDownload(const SyncPath &filepath) {
+    if (_autoUpdate) {
+        LOG_ERROR(Log::instance()->getLogger(), "Already tried to re-download the installer before.");
+        setState(UpdateState::UpdateError);
+        return;
+    }
+    _autoUpdate = true;
+    downloadUpdate();
 }
 
 } // namespace KDC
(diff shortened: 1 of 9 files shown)


src/server/updater/windowsupdater.cpp
```diff
@@ -76,6 +76,16 @@ void WindowsUpdater::startInstaller() {
         downloadUpdate();
         return;
     }
+    if (!verifyFileChecksum(filepath)) {
+        setState(UpdateState::UpdateError); // Or create UpdateState::ChecksumError
```
herve-er:
```suggestion
        setState(UpdateState::UpdateError);
```

herve-er: You can do the same as when the installer file is not found (see line 69): log + try to redownload once.
Same goes for `verifyDigitalSignature`.

src/server/updater/windowsupdater.cpp
```diff
@@ -138,11 +148,6 @@ void WindowsUpdater::downloadFinished(const UniqueId jobId) {
         return;
     }
 
-    if (!verifyDigitalSignature(filepath)) {
```
herve-er, line 141: This check can be kept here; it will avoid showing the update window to the user if we already know that it is not usable. You can also check the checksum here (keep it here and in StartInstaller, as the file can be altered between the download and the actual update).

(No need to restart the download when failing here; it will be done automatically every hour.)

test/server/updater/testwindowsupdater.cpp
```diff
@@ -146,4 +151,47 @@ void TestWindowsUpdater::testIsSignatureValid() {
...
+void TestWindowsUpdater::testIsChecksumValid() {
+    struct TestCase {
+            std::string checksumValue;
+            std::string fileName;
+            bool expectedValid;
+    };
+
+    static const std::string noChecksumValue("");
+    static const std::string fakeChecksumValue("083a301369cd711e9803f7d90d342a3778f9cb864ab22992b49fccddc3b9256c");
+    static const std::string trueChecksumValue("3d735840895bcb958f359009b06cbe9b840ae9e2df22651f431bfec4ac7b696f");
```
herve-er:
```suggestion
    static const std::string invalidChecksumValue("083a301369cd711e9803f7d90d342a3778f9cb864ab22992b49fccddc3b9256c");
    static const std::string validChecksumValue("3d735840895bcb958f359009b06cbe9b840ae9e2df22651f431bfec4ac7b696f");
```

test/server/updater/testwindowsupdater.cpp
```diff
@@ -146,4 +151,47 @@ void TestWindowsUpdater::testIsSignatureValid() {
...
+    };
+
+    for (const auto &testCase: testCases) {
+        LocalTemporaryDirectory tmpDir;
+        IoError ioError = IoError::Success;
+
+        // Only copy file if it's expected to exist
+        if (testCase.fileName == "picture-1.jpg") {
+            (void) IoHelper::copyFileOrDirectory(testhelpers::localTestDirPath() / "test_pictures/picture-1.jpg", tmpDir.path(),
+                                                 ioError);
```
herve-er, line 181: You can run the method in a `CPPUNIT_ASSERT` and check that `ioError` is success (`CPPUNIT_ASSERT_EQUAL(IoError::Success, ioError)`).

This helps pinpoint the source of the failure if we break this test.

src/server/updater/windowsupdater.cpp
```diff
@@ -173,6 +178,86 @@ std::streamsize WindowsUpdater::getExpectedInstallerSize(const std::string &down
...
+bool WindowsUpdater::verifyFileChecksum(const SyncPath &filepath) {
+    const std::string &expectedChecksum = versionInfo(_currentChannel).checksum;
+
+    auto cleanupAndFail = [&](const std::string &reason) {
+        auto ioError = IoError::Success;
+        (void) IoHelper::deleteItem(filepath, ioError);
+        if (ioError == IoError::Success) {
+            LOGW_INFO(Log::instance()->getLogger(), L"corrupted file at " << Utility::formatSyncPath(filepath) << L" deleted");
+        } else {
+            LOGW_WARN(Log::instance()->getLogger(), L"couldn't reach corrupted file at " << Utility::formatSyncPath(filepath));
```
herve-er: It could be nice to show the reason in logs.

herve-er (commented):
I also wait for the API to provide a checksum to be able to test this feature before approving it.

src/server/updater/windowsupdater.cpp
RaphaelOstier: It had been removed from there since the file could have already been installed when this happened. Which would skip this verification.
I guess it wouldn't hurt to check there too the signature and the checksum just to be sure the downloaded file isn't corrupted intentionally or accidentally.

src/server/updater/windowsupdater.cpp
RaphaelOstier: You mean the reason the delete failed, right?
because the reason the files were deleted is given in the sentry right after (computeFailed / mismatch)

src/libsyncengine/jobs/network/infomaniak_API/getappversionjob.cpp
```diff
@@ -155,8 +156,10 @@ ExitInfo GetAppVersionJob::handleResponse(std::istream &is) {
         (void) JsonParserUtility::extractValue(obj, buildVersionKey, _versionsInfo[channel].buildVersion);
         (void) JsonParserUtility::extractValue(obj, buildMinOsVersionKey, _versionsInfo[channel].buildMinOsVersion);
         (void) JsonParserUtility::extractValue(obj, downloadUrlKey, _versionsInfo[channel].downloadUrl);
+        (void) JsonParserUtility::extractValue(obj, checksumKey, _versionsInfo[channel].checksum);
 
-        if (!_versionsInfo[channel].isValid()) {
+
```
herve-er:
```suggestion
```

src/server/updater/windowsupdater.cpp
```diff
@@ -186,4 +277,14 @@ bool WindowsUpdater::verifyDigitalSignature(const SyncPath &filepath) {
     return true;
 }
 
+void WindowsUpdater::retryDownload(const SyncPath &filepath) {
+    if (_autoUpdate) {
+        LOG_ERROR(Log::instance()->getLogger(), "Already tried to re-download the installer before.");
+        setState(UpdateState::UpdateError);
```
herve-er, line 282:
```suggestion
        setState(UpdateState::UpdateError);
        _autoUpdate = false;
```

This will allow retries on the next user retry (manual or every hour).

src/server/updater/windowsupdater.cpp
```diff
@@ -173,6 +182,88 @@ std::streamsize WindowsUpdater::getExpectedInstallerSize(const std::string &down
...
+                                                                                         << static_cast<int>(ioError));
+        }
+
+        // Send to Sentry
+        KDC::sentry::Handler::captureMessage(KDC::sentry::Level::Error, "Updater::verifyChecksum::" + reason,
+                                             "Checksum verification failed: " + reason);
+        return false;
+    };
+
+    // Skip if API doesn't provide checksum (development)
```
herve-er:
```suggestion
    // Skip if API doesn't provide checksum
```

src/server/updater/windowsupdater.cpp
```diff
@@ -173,6 +198,68 @@ std::streamsize WindowsUpdater::getExpectedInstallerSize(const std::string &down
...
+
+    // Compute actual checksum
+    const std::string actualChecksum = CommonUtility::toLower(computeFileChecksum(filepath));
+    if (actualChecksum.empty()) {
+        LOGW_ERROR(Log::instance()->getLogger(), L"Failed to compute file checksum.");
+        return cleanupAndFail("computeFailed");
+    }
+
+    // verify checksum
+    if (actualChecksum != expectedChecksum) {
```
herve-er, line 237:
```suggestion
    if (actualChecksum != CommonUtility::toLower(expectedChecksum)) {
```

src/server/updater/windowsupdater.cpp
herve-er: Yes, but Sentry won’t log it. If we need to investigate an issue later, it’s better to have the information in both places.

src/server/updater/windowsupdater.cpp
RaphaelOstier: done https://github.com/Infomaniak/desktop-kDrive/pull/1760/commits/1dab13e22092a4bdfd2a3462180bec3beae8c26e

herve-er (approved):

PR #1762, opened 2026-04-15
feat: error auto resolve

herve-er:


src/libparms/db/parmsdb.h
```diff
@@ -128,6 +128,10 @@ class PARMS_EXPORT ParmsDb : public Db {
         bool selectAllErrors(ErrorLevel level, SyncDbId syncDbId, int limit, std::vector<Error> &errs);
         bool selectAllErrors(int limit, std::vector<Error> &errs);
         bool selectError(ErrorDbId dbId, Error &error, bool &found);
+        bool selectErrorByNodeInfo(SyncDbId syncDbId, const std::optional<NodeId> &localNodeId,
+                                   const std::optional<NodeId> &remoteNodeId, const std::optional<SyncPath> &path,
+                                   const std::optional<SyncPath> &destinationPath, std::vector<Error> &errs, bool &found);
```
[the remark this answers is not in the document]
herve-er: @copilot apply changes based on [this feedback](https://github.com/Infomaniak/desktop-kDrive/pull/1762#discussion_r3086188087)

src/libsyncengine/syncpal/syncpal.cpp
```diff
@@ -1428,16 +1428,19 @@ ExitInfo SyncPal::handleAccessDeniedItem(const SyncPath &relativeLocalPath, bool
...
 
     LOG_IF_FAIL(_localFSObserverWorker)
     LOG_IF_FAIL(_remoteFSObserverWorker)
     if (!_localFSObserverWorker || !_remoteFSObserverWorker) return ExitCode::LogicError;
 
     NodeId localNodeId = liveSnapshot(ReplicaSide::Local).itemId(relativeLocalPath);
+    NodeId remoteNodeId = liveSnapshot(ReplicaSide::Remote).itemId(relativeLocalPath);
+
+    Error error(syncDbId(), localNodeId, remoteNodeId,
+                relativeLocalPath.extension() == SyncPath() ? NodeType::Directory : NodeType::File, relativeLocalPath,
```
ChristopheLarchier, line 1440: So we assume that a file name without extension corresponds to a directory?
This code was already existing but seems strange to me.

herve-er: It’s a bit old, but from memory, since we’re handling items with access issues, I had trouble reliably determining the item type (IOHelper would return “access denied”). Since it’s only for display, I made a best guess, assuming it wasn’t a big deal if the UI sometimes showed a folder icon instead of a file icon in those cases.

But in hindsight, it might be better to always show a file icon?

ChristopheLarchier: Yes, it seems to me that when in doubt it's better to put the file icon there all the time or no icon.

herve-er: Done here: https://github.com/Infomaniak/desktop-kDrive/pull/1762/commits/ac0ef500d90529edad23430249dd6831f76dbe58

ChristopheLarchier (approved):

PR #1782, opened 2026-04-17
feat(macos): add debug deployment script

ClementKunz:
## Summary
- add a fast `deploy-debug.sh` helper for local macOS debug bundle deployment
- auto-detect Qt and Conan dependency locations from the build tree instead of relying on hardcoded Conan cache paths
- support a direct bundle path, a directory containing `kDrive.app`, and a `--help` option

## Testing
- `bash -n infomaniak-build-tools/macos/deploy-debug.sh`
- `bash infomaniak-build-tools/macos/deploy-debug.sh --help`

### DIFF
--- a/infomaniak-build-tools/macos/deploy-debug.sh
+++ b/infomaniak-build-tools/macos/deploy-debug.sh
@@ -62,7 +62,9 @@ fi
 # ---------------------------------------------------------------------------
 # Target bundle
 # ---------------------------------------------------------------------------
-DEFAULT_APP_BUNDLE="/Users/$(whoami)/Projects/desktop-kDrive/build-macos/client/build/Debug/install/kDrive.app"
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+REPO_ROOT="$(cd "$SCRIPT_DIR/../.." && pwd)"
+DEFAULT_APP_BUNDLE="$REPO_ROOT/build-macos/client/build/Debug/install/kDrive.app"
 INPUT_PATH="${1:-}"
 
 if [[ -z "$INPUT_PATH" ]]; then
--- a/infomaniak-build-tools/macos/deploy-debug.sh
+++ b/infomaniak-build-tools/macos/deploy-debug.sh
@@ -97,7 +97,7 @@ GENERATORS_DIR="$BUILD_DIR/generators"
 conan_pkg_folder() {
     local pattern="$1" varname="$2"
     local file
-    file=$(ls "$GENERATORS_DIR"/$pattern 2>/dev/null | head -1)
+    file=$(find "$GENERATORS_DIR" -maxdepth 1 -name "$pattern" -print -quit 2>/dev/null || true)
     if [[ -z "$file" ]]; then
         echo ""
         return
--- a/infomaniak-build-tools/macos/deploy-debug.sh
+++ b/infomaniak-build-tools/macos/deploy-debug.sh
@@ -193,26 +193,70 @@ fi
 # Manifest
 # ---------------------------------------------------------------------------
 
+# Resolve one dylib basename from a glob pattern in a directory.
+# If several files match, pick the first one in lexical order for deterministic output.
+resolve_required_dylib() {
+    local lib_dir="$1" pattern="$2" label="$3"
+    local matches=()
+    local unique=()
+    local path
+
+    shopt -s nullglob
+    for path in "$lib_dir"/$pattern; do
+        [[ -e "$path" ]] || continue
+        matches+=("$(basename "$path")")
+    done
+    shopt -u nullglob
+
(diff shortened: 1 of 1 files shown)


infomaniak-build-tools/macos/deploy-debug.sh
```diff
@@ -0,0 +1,491 @@
...
+if [[ "${1:-}" == -* ]]; then
+    echo "ERROR: Unknown option: ${1:-}" >&2
+    print_usage >&2
+    exit 2
+fi
+
+# ---------------------------------------------------------------------------
+# Target bundle
+# ---------------------------------------------------------------------------
+DEFAULT_APP_BUNDLE="/Users/$(whoami)/Projects/desktop-kDrive/build-macos/client/build/Debug/install/kDrive.app"
```
R-Gld: Maybe we should resolve the `.app` path dynamically instead of relying on a semi-hardcoded path. 
```suggestion
SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
REPO_ROOT="$(cd "$SCRIPT_DIR/../.." && pwd)"
DEFAULT_APP_BUNDLE="$REPO_ROOT/build-macos/client/build/Debug/install/kDrive.app"
```

infomaniak-build-tools/macos/deploy-debug.sh
```diff
@@ -0,0 +1,491 @@
...
+BUILD_DIR="$(dirname "$APP_BUNDLE")"   # .../install
+BUILD_DIR="$(dirname "$BUILD_DIR")"    # .../Debug  (or whatever build type)
+GENERATORS_DIR="$BUILD_DIR/generators"
+
+# Helper: extract the value of a CMake set() variable from a data file.
+# Usage: conan_pkg_folder <glob-pattern-for-data-file> <variable-name>
+conan_pkg_folder() {
+    local pattern="$1" varname="$2"
+    local file
+    file=$(ls "$GENERATORS_DIR"/$pattern 2>/dev/null | head -1)
```
R-Gld: Because we are using `set -eo pipefail`, if ls fails to find a file, or if head closes the pipe early, the whole script might unexpectedly exit. It would be safer to use `find "$GENERATORS_DIR" -maxdepth 1 -name "$pattern" -print -quit` and add `|| true` to prevent unhandled pipeline failures.
```suggestion
    file=$(find "$GENERATORS_DIR" -maxdepth 1 -name "$pattern" -print -quit 2>/dev/null || true)
```

infomaniak-build-tools/macos/deploy-debug.sh
```diff
@@ -0,0 +1,491 @@
...
+    libPocoXML.103.dylib
+    libPocoJSON.103.dylib
+)
+
+# Other Conan dylibs
+OPENSSL_DYLIBS=(libssl.3.dylib libcrypto.3.dylib)
+ZLIB_DYLIBS=(libz.1.dylib)
+LOG4CPLUS_DYLIBS=(liblog4cplus.9.dylib)
+XXHASH_DYLIBS=(libxxhash.0.dylib)
+SENTRY_DYLIBS=(libsentry.dylib)
```
R-Gld: Maybe we could use a wildcard here to automatically adapt to the version pulled by Conan in case we change dependency versions. Idk

infomaniak-build-tools/macos/deploy-debug.sh
```diff
@@ -0,0 +1,491 @@
...
+# ---------------------------------------------------------------------------
+# Pre-flight checks
+# ---------------------------------------------------------------------------
+step "Pre-flight checks"
+
+if [[ ! -f "$EXECUTABLE" ]]; then
+    echo "ERROR: kDrive executable not found at $EXECUTABLE"
+    echo "       Build the project first (cmake --build ... && cmake --install ...)"
+    exit 1
+fi
```
R-Gld, line 365: We only consider the kDrive executable here, so we don’t copy some client-only dependencies such as QtSvgWidgets.framework.

infomaniak-build-tools/macos/deploy-debug.sh
ClementKunz: Done in 4997b05d55a923eecb49deec7ee5e9ad499a3353

infomaniak-build-tools/macos/deploy-debug.sh
ClementKunz: The support of `QtSvg` is hardcoded in this script (just like all other dependencies). So `QtSvg` is deployed here, even if the `kDrive` executable does not need it.
However, you are right that we only check the existence of the `kDrive` executable, not the `kDrive_client`. Do you think it should be the case?

R-Gld: No its 👌🏼

R-Gld (approved):

PR #1800, opened 2026-04-20
fix(socket): use 127.0.0.1 instead of localhost for TCP and at auto retry

herve-er:
Replaced "localhost" with "127.0.0.1" in both C# and C++ socket connection logic to avoid DNS resolution issues and ensure consistent use of the IPv4 loopback address. This affects both client and server code paths.

### DIFF
--- a/src/gui4/windows/kDrive client/kDrive client/ServerCommunication/Services/SocketServerCommProtocol.cs
+++ b/src/gui4/windows/kDrive client/kDrive client/ServerCommunication/Services/SocketServerCommProtocol.cs
@@ -43,6 +43,7 @@ namespace Infomaniak.kDrive.ServerCommunication.Services
         private int _inBufferJsonBalance = 0;
         private readonly ConcurrentDictionary<long, TaskCompletionSource<CommData>> _pendingRequests = [];
         private bool _stopRequested = false;
+        private const string _host = "127.0.0.1";
         private long NextId
         {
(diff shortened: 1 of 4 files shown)


src/gui4/windows/kDrive client/kDrive client/ServerCommunication/Services/SocketServerCommProtocol.cs
```diff
@@ -95,7 +95,7 @@ private async Task ReconnectLoop()
                         Logger.Log(Logger.Level.Info, $"Attempting to connect to localhost:{port}");
                         _client?.Dispose();
                         _client = new TcpClient();
-                        await _client.ConnectAsync("localhost", port).ConfigureAwait(false);
+                        await _client.ConnectAsync("127.0.0.1", port).ConfigureAwait(false);
```
ChristopheLarchier: If possible, use a constant.

herve-er: Done here:
https://github.com/Infomaniak/desktop-kDrive/pull/1800/commits/27d2b45a0c0dd565e467e63bc7e09194ce6cfb1f

ChristopheLarchier (dismissed):

ChristopheLarchier (approved):

PR #1807, opened 2026-04-21
fix(macOS UI): Re-order actions so main window is always key and front

valentinperignon:


src/gui4/macOS/kDrive/AppDelegate.swift
```diff
@@ -56,14 +56,14 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
...
-        mainWindow.window?.makeKeyAndOrderFront(nil)
-
         if #available(macOS 14.0, *) {
             NSApp.activate()
         } else {
             NSApp.activate(ignoringOtherApps: true)
         }
+
+        mainWindow.showWindow(nil)
+        mainWindow.window?.makeKeyAndOrderFront(nil)
```
adrien-coye, line 66: @copilot What should we do here to make sure to not only bring to front the window but jump to the proper space ?

adrien-coye (approved):
Tested, the window does now jump back to front.

Regarding spaces: Copilot solution to move the window to the correct space is not perfect but better than nothing. Maybe later we will be able to jump to the space where the window is open. Merging for now.

adrien-coye (approved):

PR #1812, opened 2026-04-21
fix(conan): make --update opt-in to ensure deterministic CI builds

R-Gld:
## Summary
The Conan install scripts were unconditionally passing `--update`, which could cause recipe revision/timestamp drift on local recipes after a checkout and introduce non-determinism in CI builds. This PR makes `--update` opt-in via an explicit flag on both the Bash and PowerShell scripts, keeping CI builds reproducible by default while still allowing on-demand remote refresh.

## Changes
- `build_dependencies.sh`: removed hardcoded `--update` from `conan install`; added `--update` flag (sets `enable_update=true`) and logs `Conan update enabled` state at runtime.
- `build_dependencies.ps1`: removed hardcoded `--update`; added `-Update` switch parameter; refactored `conan install` arguments into an array for clarity; improved remote detection to use JSON output and URL normalization instead of regex string matching.
- `build_dependencies.ps1`: cleaned up code style (consistent brace placement, indentation, string interpolation).
- `recipes/openssl-universal/config.yml`: deleted obsolete config file superseded by the `openssl-macos` recipe migration.
- `THIRD_PARTY_NOTICES.md`: updated OpenSSL macOS recipe reference from `openssl-universal` to `openssl-macos`; moved Qt and several other dependencies under the Conan-managed section; improved formatting (blank lines between entries, line wrapping for long lines).

## Technical Details
The root cause was that `--update` forces Conan to contact remotes and refresh recipe metadata, including timestamps. Even when the recipe content is identical, a new timestamp produces a different revision hash, breaking reproducibility after a clean checkout. By making the flag opt-in, CI pipelines that do not pass `--update` are fully deterministic. The PowerShell remote detection was also improved: it now parses `conan remote list --format=json` and normalizes path separators before comparing URLs, replacing the previous fragile regex match against the raw text output.

## Testing
- Bash syntax verified with `bash -n infomaniak-build-tools/conan/build_dependencies.sh`.
- Help output confirmed to include the `--update` option and its description.
- Existing CI call sites for Linux, macOS, and Windows verified to not pass `--update` / `-Update`, so no behavioral change for current pipelines.

## Notes
Manual usage when a remote refresh is needed:
- Linux/macOS: `bash infomaniak-build-tools/conan/build_dependencies.sh Release --update`
- Windows: `powershell ./infomaniak-build-tools/conan/build_dependencies.ps1 Release -Update`

### DIFF
--- a/THIRD_PARTY_NOTICES.md
+++ b/THIRD_PARTY_NOTICES.md
@@ -54,7 +54,7 @@ The exact set of redistributed binaries can vary by platform:
 - **License:** LGPL-3.0 / GPL-2.0 / GPL-3.0 / Commercial
 - **Copyright:** Copyright (C) The Qt Company Ltd. and other contributors
 - **Repository:** https://www.qt.io/
-- **Note:** Declared in `conanfile.py`.This application is built against Qt 6.2.3. Qt is deployed with the application
+- **Note:** Declared in `conanfile.py`. This application is built against Qt 6.2.3. Qt is deployed with the application
   on supported platforms. Qt source packages are available from https://download.qt.io/
 
 ### Poco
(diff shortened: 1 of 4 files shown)


adrien-coye (approved):
I did review this PR, and it does look good to me. 
I asked for someone else review before we can merge it idealy.

ClementKunz (approved):

PR #1814, opened 2026-04-22
feat: Returning to onboarding view while all users are disconnected

killianmathias:


src/gui4/macOS/kDrive/UI/Controllers/MainWindow/MainWindowController.swift
```diff
@@ -74,6 +76,21 @@ final class MainWindowController: NSWindowController {
             }
     }
 
+    private func observeUsersCache() {
+        cacheObservable.usersPublisher
+            .receiveOnMain(store: &bindStore) { [weak self] _ in
+                guard let self = self else { return }
+                guard case .mainWindow = router.currentRoute else {
```
adrien-coye, line 87: I think I would rather filter out the users and explicitly call  `router.navigate(to: .onboarding())`
This way we have the explicit logic following up the event and the action in the same place abstract of the other logic.

adrien-coye: Discussed face to face, the .receiveOnMain allows you to receive a collection of users.
You want to check for .isEmpty

adrien-coye (commented):
We discussed face to face 👍

src/gui4/macOS/kDrive/UI/Controllers/MainWindow/MainWindowController.swift
adrien-coye: Ok cool I see the `.map { $0.isEmpty }.removeDuplicates()` but you need to check the value you receive from the combine flow. As I understand it the jump to the onboarding only should be performed when we fall to zero and thus when isEmpty is `true`

adrien-coye (approved):
Did a test on my machine, all good. Cool thanks for the change, I see the explicit guard on the value fetched from observation. All good for me 👍

PR #1845, opened 2026-04-27
feat: Using first name instead of name on home page

killianmathias:


src/gui4/macOS/kDriveCoreUI/Utils/Previews/PreviewHelper.swift
```diff
@@ -48,7 +48,8 @@ public enum PreviewHelper {
     public static let user = UIUser(
         dbId: 95014,
         userId: 95014,
-        name: "Tim Cook",
```
adrien-coye, line 51: I'm nitpicking, but the API still return the full { first + last } name in "name"
So name is still "Tim Cook". I'd like to keep it this way (and in other tests) as to make this clear if someone reads the tests.
In the settings we still display the full username for ex.

src/gui4/macOS/kDriveResources/Symbols/Strings+Symbols.swift
```diff
@@ -143,7 +143,7 @@ public enum KDriveLocalizable {
   /// loco:698b244d7e73966ce3096b02
   public static let copyingLink = KDriveLocalizable.tr("Localizable", "copyingLink", fallback: "Copying Link…")
   /// loco:698b169cf72c91b1760a7002
-  public static let dataManagementDescription = KDriveLocalizable.tr("Localizable", "dataManagementDescription", fallback: "You can customize the usage and operational data collected by Infomaniak to fix bugs and improve the application. Both tools used are hosted and managed exclusively by us.\n\n\nThe source code of this application can be verified, the information collected is anonymous and your personal information remains strictly confidential and is never shared with third parties.")
+  public static let dataManagementDescription = KDriveLocalizable.tr("Localizable", "dataManagementDescription", fallback: "You can customize the usage and operational data collected by Infomaniak to fix bugs and improve the application. Both tools used are hosted and managed exclusively by us.\n\nThe source code of this application can be verified, the information collected is anonymous and your personal information remains strictly confidential and is never shared with third parties.")
```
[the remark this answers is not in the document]
adrien-coye: I think we can roll back this change 👍

src/gui4/macOS/kDriveTests/Parsing/Signals/UserAddedParsingTest.swift
```diff
@@ -53,7 +53,8 @@ struct UserAddedParsingTest {
         #expect(signal.num == SignalNum.USER_ADDED)
         #expect(signal.body.userInfo.dbId == 1)
         #expect(signal.body.userInfo.userId == 12_345_678)
-        #expect(signal.body.userInfo.name == "duck mc duckface")
```
adrien-coye, line 56: Same here we can keep the name unchanged

adrien-coye (commented):
Looking good 👍 
I only have some small knitpicks 🙏

adrien-coye (approved):
Thanks for the changes, all good to me 👍

PR #1847, opened 2026-04-27
refactor(appcache): graph-backed cache foundation with selection and onboarding stores

R-Gld:
## Summary

Replaces the flat, vector-based `AppCache` with a normalized graph model (`User -> Account -> Drive -> Sync -> Error`) and introduces two dedicated state stores: `MainSelectionStore` (owns `currentSyncDbId`) and `OnboardingState` (owns onboarding user, drive key selections, and pending sync configs). This is the Linux v4 cache foundation branch. No `CachePipeline` real wiring is included; that belongs to the follow-up `linux-v4/cache/services` branch.

### AppCache (`appcache.h/cpp`, `appcachegraph.cpp`, `appcachequeries.cpp`)
- Internal storage replaced from `std::vector<*Info>` to `std::unordered_map<*DbId, *Node>` keyed by database id. Each node (`UserNode`, `AccountNode`, `DriveNode`, `SyncNode`) carries its entity info plus sorted child id lists, forming an explicit adjacency graph.
- Cascade removal: `removeUser` walks and deletes all descendant accounts, drives, syncs, and errors; the silent cascade helpers (`removeAccountCascade`, `removeDriveCascade`, `removeSyncCascade`) do not emit signals and are called by the public remove methods which emit the appropriate coarse signals after cleanup.
- Snapshot replacement methods (`replaceUsers`, `replaceAccounts`, etc.) prune orphans via `pruneConfiguredGraph()` and rebuild child id lists via `rebuildGraphRelations()` after each bulk load.
- `availableDrives` is now per-user: `replaceAvailableDrivesForUser(UserDbId, ...)` replaces one user's snapshot without touching others; `availableDrivesChanged(UserDbId)` signal carries the user scope.
- Sync errors and server errors are tracked in separate maps (`_syncErrorsByDbId`, `_serverErrorsByDbId`) with their own upsert/remove/replace/signal pairs (`syncErrorsChanged`, `serverErrorsChanged`).
- Selection state (`selectedUserDbId`, `selectedDriveDbId`, `selectedSyncDbId`) and connection state are removed from `AppCache`; these responsibilities are now owned by `MainSelectionStore` and callers respectively.
- New id-based point lookups return `std::optional<*Info>` (e.g. `user(UserDbId)`, `drive(DriveDbId)`, `sync(SyncDbId)`).
- New graph traversal helpers: `accountsForUser`, `drivesForAccount`, `syncsForDrive`, `errorsForSync`; results are stable-sorted by database id.
- New derived read models (`appcachequeries.cpp`): `syncContext(SyncDbId)`, `syncContexts()`, `driveContext(DriveDbId)`, `driveContexts()`, `availableDriveContexts(UserDbId)`, `availableDriveContexts()`; a context is `nullopt` when any node in its parent chain is missing.
- Flat snapshot accessors (`users()`, `accounts()`, etc.) now reconstruct sorted `std::vector` from the maps on each call (copy-out, not reference return).

### Cache types (`cachetypes.h`)
- New header consolidating all cache-layer value types.
- `SyncContext`: bundles `UserInfo`, `AccountInfo`, `DriveInfo`, `SyncInfo`, errors list, and `latestError`.
- `DriveContext`: bundles user, account, drive, and syncs list.
- `AvailableDriveContext`: bundles user, optional account, `DriveAvailableInfo`, `alreadyConfigured` flag, and optional configured `DriveInfo`.
- `AvailableDriveKey`: composite key `(userDbId, accountId, driveId)` with `operator==` and `std::hash` specialization for use in `unordered_set` and `unordered_map`.
- `PendingSyncConfig`: captures local path, target path, target node id, and VFS mode; used by `OnboardingState`.
- `SyncContext`, `DriveContext`, and `AvailableDriveContext` now define `operator== = default`, which lets the main-selection store compare derived contexts directly.

### MainSelectionStore (`mainselectionstore.h/cpp`)
- New `QObject` subclass, sole owner of `currentSyncDbId`; exposed as a read-only `Q_PROPERTY` to QML.
- `selectSync(qint64)` validates that the id exists in the cache before committing; `ensureValidSelection()` re-pins to the first available sync when the graph changes.
- `_lastRequestedSyncDbId` is kept separately so a sync temporarily absent from the cache can be restored if it reappears.
- `currentSyncContext()` delegates to `AppCache::syncContext`; no duplicate entity storage.
- `handleSyncsChanged()` and `handleContextDataChanged()` are fine-grained: `currentSyncContextChanged()` is emitted only when the derived context actually differs, not on every cache signal.

### OnboardingState (`onboardingstate.h/cpp`)
- New `QObject` subclass owning onboarding-flow-specific state only: selected user, a set of selected `AvailableDriveKey` values, and a map of `PendingSyncConfig` keyed by `AvailableDriveKey`.
- `pruneSelectedAvailableDrives()` drops keys that no longer belong to the selected user when the available drives snapshot changes.
- `reset()` clears all three fields and emits all three change signals.
- Holds a `AppCache &` reference; reads entities from the cache but owns none.

### UserService (`userservice.h/cpp`)
- `loadAvailableDrives` uses a per-user generation counter (`_availableDriveLoadGenerations: unordered_map<UserDbId, uint64_t>`) to discard stale async responses when a newer request has been issued for the same user.
- Replaces the old `selectedUserDbId()` guard (which coupled the service to selection state) with the generation check, which is correct regardless of current selection.
- Calls `replaceAvailableDrivesForUser` instead of the removed global `replaceAvailableDrives`.
- Uses a single casted `UserDbId` local inside `loadAvailableDrives()` instead of carrying both raw and casted ids through the method.

### AppClientLinux (`appclientlinux.h`)
- `AppCache _appCache`, `MainSelectionStore _mainSelectionStore`, and `OnboardingState _onboardingState` are owned members of `AppClientLinux`, constructed in declaration order with proper `AppCache &` references.
- Public accessors `appCache()`, `mainSelectionStore()`, `onboardingState()` added for downstream service and adapter wiring.

## Architecture notes

- All cache, selection, and onboarding mutations must run on the Qt main thread. No internal locking exists; callers must not invoke mutating methods from worker threads.
- `AppCache` emits coarse invalidation signals only. Consumers must re-query the full read model on each signal rather than diffing. This matches the macOS pattern.
- Cascade removes do not emit intermediate signals; only the top-level public remove methods emit, preventing observer storms on multi-level deletions.
- `MainSelectionStore` and `OnboardingState` hold `AppCache &` references and are declared after `_appCache` in `AppClientLinux`, guaranteeing correct construction and destruction order.
- `availableDrives` is intentionally scoped per-user and never global. The `AvailableDriveContext` query reconciles addable drives with configured drives (already-configured flag) for display decisions.
- `CachePipeline` service adapter wiring is deferred to `linux-v4/cache/services`.
- `MainSelectionStore` emits `currentSyncContextChanged()` only when the selected sync context actually changes, while still healing selection on cache changes.

### DIFF
--- a/src/gui4/linux/app/cache/appcache.cpp
+++ b/src/gui4/linux/app/cache/appcache.cpp
@@ -28,6 +28,7 @@ AppCache::AppCache(QObject *const parent) :
     QObject(parent) {}
 
 void AppCache::clearAll() {
+    qCInfo(lcAppCache) << "Cache cleared";
     const bool hadUsers = !_usersByDbId.empty();
     const bool hadAccounts = !_accountsByDbId.empty();
     const bool hadDrives = !_drivesByDbId.empty();
@@ -74,6 +75,7 @@ void AppCache::replaceAccounts(const std::vector<AccountInfo> &accounts) {
     _accountsByDbId.clear();
     for (const auto &info: accounts) {
         if (!_usersByDbId.contains(info.userDbId())) {
+            qCWarning(lcAppCache) << "Account dropped during replace | accountDbId:" << info.dbId() << "/ unknown userDbId:" << info.userDbId();
             continue;
         }
         _accountsByDbId[info.dbId()] = AccountNode{info, info.userDbId(), {}};
@@ -92,6 +94,7 @@ void AppCache::replaceDrives(const std::vector<DriveInfo> &drives) {
     _drivesByDbId.clear();
     for (const auto &info: drives) {
         if (!_accountsByDbId.contains(info.accountDbId())) {
+            qCWarning(lcAppCache) << "Drive dropped during replace | driveDbId:" << info.dbId() << "/ unknown accountDbId:" << info.accountDbId();
             continue;
         }
         _drivesByDbId[info.dbId()] = DriveNode{info, info.accountDbId(), {}};
@@ -109,6 +112,7 @@ void AppCache::replaceSyncs(const std::vector<SyncInfo> &syncs) {
     _syncsByDbId.clear();
     for (const auto &info: syncs) {
         if (!_drivesByDbId.contains(info.driveDbId())) {
+            qCWarning(lcAppCache) << "Sync dropped during replace | syncDbId:" << info.dbId() << "/ unknown driveDbId:" << info.driveDbId();
             continue;
         }
         _syncsByDbId[info.dbId()] = SyncNode{info, info.driveDbId(), {}};
@@ -125,6 +129,7 @@ void AppCache::replaceSyncErrors(const std::vector<ErrorInfo> &errors) {
     _syncErrorsByDbId.clear();
     for (const auto &info: errors) {
         if (!_syncsByDbId.contains(info.syncDbId())) {
+            qCWarning(lcAppCache) << "Sync error dropped during replace | errorDbId:" << info.dbId() << "/ unknown syncDbId:" << info.syncDbId();
             continue;
         }
         _syncErrorsByDbId[info.dbId()] = info;
@@ -170,6 +175,7 @@ void AppCache::clearAllAvailableDrives() {
 void AppCache::upsertUser(const UserInfo &info) {
     auto &node = _usersByDbId[info.dbId()];
     node.info = info;
+    qCDebug(lcAppCache) << "User upserted | dbId:" << info.dbId();
     emit usersChanged();
 }
 
@@ -178,6 +184,7 @@ void AppCache::removeUser(const UserDbId userDbId) {
     if (userIt == _usersByDbId.end()) {
         return;
     }
+    qCDebug(lcAppCache) << "User removed | dbId:" << userDbId;
 
     for (const auto accountDbIds = userIt->second.accountDbIds; const auto accountDbId: accountDbIds) {
         removeAccountCascade(accountDbId);
@@ -195,6 +202,7 @@ void AppCache::removeUser(const UserDbId userDbId) {
 
 void AppCache::upsertAccount(const AccountInfo &info) {
     if (!_usersByDbId.contains(info.userDbId())) {
+        qCWarning(lcAppCache) << "Account upsert dropped | accountDbId:" << info.dbId() << "/ unknown userDbId:" << info.userDbId();
         return;
     }
 
@@ -207,6 +215,7 @@ void AppCache::upsertAccount(const AccountInfo &info) {
     node.info = info;
     node.parentUserDbId = info.userDbId();
     linkAccountToUser(info.dbId(), info.userDbId());
+    qCDebug(lcAppCache) << "Account upserted | dbId:" << info.dbId() << "/ userDbId:" << info.userDbId();
     emit accountsChanged();
 }
 
@@ -214,6 +223,7 @@ void AppCache::removeAccount(const AccountDbId accountDbId) {
     if (!_accountsByDbId.contains(accountDbId)) {
         return;
     }
+    qCDebug(lcAppCache) << "Account removed | dbId:" << accountDbId;
     removeAccountCascade(accountDbId);
     emit accountsChanged();
     emit drivesChanged();
@@ -223,6 +233,7 @@ void AppCache::removeAccount(const AccountDbId accountDbId) {
 
 void AppCache::upsertDrive(const DriveInfo &info) {
     if (!_accountsByDbId.contains(info.accountDbId())) {
+        qCWarning(lcAppCache) << "Drive upsert dropped | driveDbId:" << info.dbId() << "/ unknown accountDbId:" << info.accountDbId();
         return;
     }
 
@@ -235,6 +246,7 @@ void AppCache::upsertDrive(const DriveInfo &info) {
     node.info = info;
     node.parentAccountDbId = info.accountDbId();
     linkDriveToAccount(info.dbId(), info.accountDbId());
+    qCDebug(lcAppCache) << "Drive upserted | dbId:" << info.dbId() << "/ accountDbId:" << info.accountDbId();
     emit drivesChanged();
 }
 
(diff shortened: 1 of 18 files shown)


src/gui4/linux/app/cache/appcache.cpp
```diff
@@ -16,251 +16,318 @@
...
-        return;
+void AppCache::replaceUsers(const std::vector<UserInfo> &users) {
+    _usersByDbId.clear();
+    for (const auto &info: users) {
+        _usersByDbId[info.dbId()].info = info;
     }
-    _connected = connected;
-    emit connectedChanged();
+    pruneConfiguredGraph();
+    emit usersChanged();
```
adrien-coye, line 63: Is there a way with your framework to only use usersChanged() and rely on deduplication of events by the library ?

src/gui4/linux/app/cache/appcache.h
```diff
@@ -91,34 +109,76 @@ class AppCache : public QObject {
...
+
+        // Prunes orphan entities after snapshot replacement and rebuilds child id lists from canonical maps.
+        void pruneConfiguredGraph();
+        void rebuildGraphRelations();
+
+        // Resolve configured-account/drive matches for addable-drive read models.
+        [[nodiscard]] std::optional<AccountInfo> accountForAvailableDrive(UserDbId userDbId, AccountId accountId) const;
+        [[nodiscard]] std::optional<DriveInfo> configuredDriveForAvailableDrive(AccountDbId accountDbId, DriveId driveId) const;
+
+        std::unordered_map<UserDbId, UserNode> _usersByDbId;
```
adrien-coye, line 174: On macOS we use something akin to an "ordered_map", as the cache drive the view and we do not want the UI to glitch when order is not preserved _or_ have to track the order in another data structure.

src/gui4/linux/app/cache/appcache.h
```diff
@@ -18,67 +18,85 @@
...
+        // Direct id-based lookups. Missing or orphaned entities are returned as std::nullopt.
+        [[nodiscard]] std::optional<UserInfo> user(UserDbId userDbId) const;
+        [[nodiscard]] std::optional<AccountInfo> account(AccountDbId accountDbId) const;
+        [[nodiscard]] std::optional<DriveInfo> drive(DriveDbId driveDbId) const;
+        [[nodiscard]] std::optional<SyncInfo> sync(SyncDbId syncDbId) const;
+        [[nodiscard]] std::optional<ErrorInfo> syncError(ErrorDbId errorDbId) const;
+        [[nodiscard]] std::optional<ErrorInfo> serverError(ErrorDbId errorDbId) const;
+        [[nodiscard]] std::optional<DriveAvailableInfo> availableDrive(const AvailableDriveKey &key) const;
+
+        // Parent-to-children graph traversal helpers. Results are stable-sorted by database id.
```
adrien-coye, line 65: On macOS, we order by arrival order in the cache. Do you order by dbId @herve-er ? (I might have to update on my side for the sake of consistency)

src/gui4/linux/app/cache/cachetypes.h
```diff
@@ -0,0 +1,96 @@
...
+        UserDbId userDbId{0};
+        AccountId accountId{0};
+        DriveId driveId{0};
+
+        friend bool operator==(const AvailableDriveKey &lhs, const AvailableDriveKey &rhs) {
+            return lhs.userDbId == rhs.userDbId && lhs.accountId == rhs.accountId && lhs.driveId == rhs.driveId;
+        }
+};
+
+struct SyncContext {
```
adrien-coye, line 46: Okay I see the "flat" structures to drive the view 👍

adrien-coye (commented):
Looking good 👍 
I did not review in detail implementation of pruning cache for example, but I did check the architecture.
I have a couple of questions I'd like to ask you face to face 🙏

src/gui4/linux/app/cache/appcachequeries.cpp
```diff
@@ -0,0 +1,404 @@
...
+    for (const auto errorDbId: syncIt->second.errorDbIds) {
+        if (const auto errorInfo = syncError(errorDbId)) {
+            values.push_back(*errorInfo);
+        }
+    }
+    (void) std::ranges::sort(values, [](const ErrorInfo &lhs, const ErrorInfo &rhs) { return lhs.getTime() < rhs.getTime(); });
+    return values;
+}
+
+std::vector<ErrorInfo> AppCache::syncErrors(const SyncDbId syncDbId) const {
```
herve-er: Is this method really needed? It seems to just be a copy of `errorsForSync`.

src/gui4/linux/app/cache/appcache.h
herve-er: I also order by arrival order

src/gui4/linux/app/cache/appcache.cpp
adrien-coye: Discussed with @herve-er and @R-Gld QT does not work like this. All good.

src/gui4/linux/app/cache/appcache.h
adrien-coye: Discussed, Linux will order by ID as less helpful types and libraries are available.
The server does send in order of `dbid` anyway so it should be transparent for the user.
All good 👍

src/gui4/linux/app/cache/appcache.h
adrien-coye: Discussed not easily available or needs to import Boost or something. Therefore, not doing it on Linux.

adrien-coye (commented):
All my remarks are relsoved, I leave it to @herve-er to give the final GO 👍

src/gui4/linux/app/cache/appcachequeries.cpp
R-Gld: well well well 😅 

https://github.com/Infomaniak/desktop-kDrive/pull/1847/commits/3a0ba7d9ddf38c143eab061de80983c696a6837a

herve-er (approved):

PR #1849, opened 2026-04-27
fix: Creation and edition timestamps wrongly set to 0 after hydration

herve-er:


src/libsyncengine/jobs/network/kDrive_API/downloadjob.h
```diff
@@ -36,8 +36,13 @@ class DownloadJob : public AbstractTokenNetworkJob {
                 SyncTime modificationTime = 0;
                 bool isCreate = false;
         };
+        enum class DateTimePolicy {
+            IgnoreDateTime,
+            ApplyDateTime
+        };
+
         DownloadJob(const std::shared_ptr<Vfs> vfs, std::shared_ptr<CacheDirectory> cacheDirectory,
```
ClementKunz, line 44: I don't agree with Copilot, I prefer forcing the caller to pass the value of `DateTimePolicy` explicitly. However, there is a confusion and `UseDateTime` is still used on macOS

src/libsyncengine/propagation/executor/executorworker.cpp
```diff
@@ -831,7 +832,8 @@ ExitInfo ExecutorWorker::generateEditJob(SyncOpPtr syncOp, std::shared_ptr<SyncJ
                     DownloadJob::FileDownloadInfo{_syncPal->driveDbId(), syncOp->affectedNode()->id().value_or(""),
                                                   absoluteLocalFilePath, syncOp->affectedNode()->size(),
                                                   syncOp->affectedNode()->createdAt().value_or(0),
-                                                  syncOp->affectedNode()->modificationTime().value_or(0), false});
+                                                  syncOp->affectedNode()->modificationTime().value_or(0), false},
+                    DownloadJob::DateTimePolicy::ApplyDateTime);
```
ClementKunz, line 836: Formatting looks odd here. It is difficult to see that `DownloadJob::DateTimePolicy::ApplyDateTime` is a parameter

ClementKunz: Forget about it, formatting is fine

src/libsyncengine/jobs/network/kDrive_API/downloadjob.h
herve-er: done

ClementKunz (approved):

PR #1872, opened 2026-05-04
feat: env variable for stdout logs

RaphaelOstier:
This pull request introduces an environment variable, `ENABLE_LOG_TO_CONSOLE`, to control whether logs are printed to the console in debug builds. This provides developers with a convenient way to enable or disable console logging without modifying code, making debugging easier and less intrusive. The changes are applied consistently across both client and server logging components.

**Logging improvements:**

* Added checks for the `ENABLE_LOG_TO_CONSOLE` environment variable in `Logger::doLog` and `CustomRollingFileAppender::append` to conditionally print logs to the console in debug builds. [[1]](diffhunk://#diff-68c0e0c740cd097f516198e208d0e4f540343bfaadc0134ce9a7a365933472bbR175-R177) [[2]](diffhunk://#diff-5c3887eff30b60981a0abdd0af69f5f7d77ac4ef5e70a1f3061d4b071272547fR195) [[3]](diffhunk://#diff-5c3887eff30b60981a0abdd0af69f5f7d77ac4ef5e70a1f3061d4b071272547fR205)
* Updated the singleton initialization in `Log::instance` (Windows, debug builds) to only allocate a console and redirect output streams if `ENABLE_LOG_TO_CONSOLE` is set.

### DIFF
--- a/src/libcommon/utility/utility.cpp
+++ b/src/libcommon/utility/utility.cpp
@@ -1169,7 +1169,10 @@ std::string CommonUtility::envVarValue(const std::string &name, bool &isSet) {
 }
 
 #ifndef NDEBUG
-const bool CommonUtility::logToConsole = CommonUtility::envVarValue("KDRIVE_ENABLE_LOG_TO_CONSOLE") == "1";
+bool CommonUtility::logToConsoleEnabled() {
+    static const bool value = CommonUtility::envVarValue("KDRIVE_ENABLE_LOG_TO_CONSOLE") == "1";
+    return value;
+}
 #endif
 
 int CommonUtility::setenv(const char *const name, const char *const value, const int overwrite) {
--- a/src/libcommon/utility/utility.cpp
+++ b/src/libcommon/utility/utility.cpp
@@ -1168,12 +1168,14 @@ std::string CommonUtility::envVarValue(const std::string &name, bool &isSet) {
     return std::string();
 }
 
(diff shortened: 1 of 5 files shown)


src/libcommon/utility/utility.cpp
```diff
@@ -1133,6 +1133,13 @@ std::string CommonUtility::envVarValue(const std::string &name, bool &isSet) {
     return std::string();
 }
 
+#ifndef NDEBUG
+bool CommonUtility::logToConsoleEnabled() {
+    static const bool value = CommonUtility::envVarValue("KDRIVE_ENABLE_LOG_TO_CONSOLE") == "1";
+    return value;
+}
+#endif
```
ClementKunz: To simplify even more:

```suggestion
bool CommonUtility::logToConsoleEnabled() {
#ifndef NDEBUG
    static const bool value = CommonUtility::envVarValue("KDRIVE_ENABLE_LOG_TO_CONSOLE") == "1";
    return value;
#else
     return false;
#endif
}
```

This way, we do not need the `#ifndef NDEBUG` everywhere else

RaphaelOstier: done https://github.com/Infomaniak/desktop-kDrive/pull/1872/commits/96b8457ed23d8d0effbb628a2f65198a25122bd2

ClementKunz (approved):

PR #1880, opened 2026-05-05
feat(server): improve callback thread management

herve-er:
feat(server): improve callback thread management

Refactor callback thread handling in SocketCommChannel to use direct method reference and add joinCallbackThread() for controlled joining. In SocketCommServer, defer lost connection handling to a separate thread to prevent deadlocks and log a warning if the callback thread cannot be joined.

### DIFF
--- a/src/server/comm/socketcommserver.cpp
+++ b/src/server/comm/socketcommserver.cpp
@@ -334,9 +334,9 @@ void SocketCommServer::execute() {
     }
     _isListening = false;
 }
-void SocketCommServer::joinAndClearPostponedLostConnectionCbks() {
+void SocketCommServer::joinAndClearPostponedLostConnectionCbks() noexcept {
     // Join and remove all postponed lost-connection callback threads
-    for (const auto thread: _postponedLostConnectionCbks) {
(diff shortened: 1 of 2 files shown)


ClementKunz:
Maybe if you can just handle the 2 small SonarCloud issues. Otherwise, LGTM.

src/server/comm/socketcommserver.cpp
```diff
@@ -52,11 +43,7 @@ SocketCommChannel::~SocketCommChannel() {
     }
 
     if (_callbackThread && _callbackThread->joinable()) {
-        if (_callbackThread->get_id() == std::this_thread::get_id()) {
-            _callbackThread->detach();
-        } else {
-            _callbackThread->join();
-        }
+        _callbackThread->join();
     }
```
ClementKunz: Why not using `SocketCommChannel::joinCallbackThread()` here?

herve-er: Done here https://github.com/Infomaniak/desktop-kDrive/pull/1880/commits/326c3969507c6b4348f1c4b20386b20cadc21be9

ClementKunz (approved):

PR #1888, opened 2026-05-07
feat: Support for sync error listing

adrien-coye:
Job ERROR_INFOLIST + Tests


killianmathias (approved):
Tested on my device, it works well! The errors are shown and disappear when there are no errors.

PR #1903, opened 2026-05-12
feat: enhance socket comm layer

herve-er:
This pull request refactors and improves the server communication infrastructure in the kDrive Windows client. The main focus is on enhancing the socket-based communication protocol, introducing explicit connection initialization with cancellation support, and improving error handling and resource management. Additionally, some code quality improvements and minor bug fixes are included.


src/gui4/windows/kDrive client/kDrive client/ServerCommunication/Services/SocketServerCommProtocol.cs
```diff
@@ -34,119 +33,168 @@
...
 namespace Infomaniak.kDrive.ServerCommunication.Services
 {
-    // For the moment, non implemented methods will fallback to the MockServerCommProtocol implementation
     public class SocketServerCommProtocol : Interfaces.IServerCommProtocol
     {
-        private TcpClient? _client;
-        private long _requestIdCounter = 0;
-        private string _inBuffer = "";
+        private Socket? _socket;
+        private readonly SemaphoreSlim _sendLock = new(1, 1);
```
ClementKunz, line 39: Maybe this is a stupid question, but why not use a mutex here instead of a semaphore with a maximum of 1 thread? The goal is to lock it by default?

herve-er: A `Mutex` doesn’t allow taking full advantage of `async/await`.

A mutex will completely block the caller thread until it acquires `_sendLock`. If this happens on the UI thread, it would block the UI thread and freeze the UI.

`SemaphoreSlim` works with:

```csharp
await _sendLock.WaitAsync();
```

so the caller thread is released while waiting for the lock and can continue doing other work until the lock is acquired (for exemple refresh the UI).

herve-er: > The goal is to lock it by default?

A semaphore works differently from a traditional mutex/thread lock.

 It starts with a count of available “slots” (here `1`), each call to `Wait()` (or `WaitAsync()`) acquires one slot until none are left. Once the count reaches `0`, additional callers are blocked until a slot is made available again by calling `Release()`.

 The second `1` in `new SemaphoreSlim(1, 1)` represents the maximum count. It prevents the semaphore count from exceeding `1`, even if `Release()` is called multiple times in a row.

ClementKunz (approved):

PR #1923, opened 2026-05-20
feat(macOS): Handle errors

valentinperignon:


src/gui4/macOS/kDrive/UI/Views/MainWindow/Errors/Components/ErrorsHeaderView.swift
```diff
@@ -0,0 +1,63 @@
...
+
+    let synchroDbId: UISynchro.ID?
+    let errorsCount: Int
+
+    private var title: AttributedString {
+        do {
+            return try AttributedString(markdown: KDriveLocalizable.informationBlockSynchroErrorTitle(errorsCount))
+        } catch {
+            #if DEBUG
+            fatalError("Failed decoding AttributedString: \(error)")
```
adrien-coye, line 35: Nitpick: I think using an assert would do the same thing without the if else danse

src/gui4/macOS/kDrive/Utils/UISynchroErrorCategory.swift
```diff
@@ -0,0 +1,42 @@
...
+    case systemAndPermissions
+    case conflicts
+    case filesToCheck
+    case synchronizationDirectories
+    case storage
+
+    public var title: String {
+        switch self {
+        case .systemAndPermissions:
+            return "!Système et permissions"
```
adrien-coye, line 31: i18n in this PR or a later PR ?

adrien-coye (commented):
From a code perspective all good to me, I just found one nitpick.
From a complexity perspective I understand we have to match what exists, and you showed me the source of truth is basically the windows V4 app.
I see translations but as we discussed this will not be merged right away.

adrien-coye (approved):
Okay, re-reviewed the PR, looks good to me. The temp commits are effectively gone.

PR #1927, opened 2026-05-20
feat(macOS): Show internal distribution channel to staff users only

valentinperignon:


adrien-coye (approved):
Reviewed an tested. All good to me.

PR #1944, opened 2026-05-27
feat(log): persist Linux logs in XDG_STATE_HOME

ClementKunz:
## Summary
- Modified "CommonUtility::logDirectoryPath()" on Linux to store logs in a persistent location
- Follows the XDG Base Directory Specification
- Uses "$XDG_STATE_HOME/{app}/logs/" with fallback to "~/.local/state/{app}/logs/"
- Replaces the previous temporary directory approach where logs were lost after reboot

## Rationale
Log files should persist across system restarts. The previous implementation placed them under /tmp/, causing data loss on reboot. The new path is widely supported across all modern Linux distributions.

## Changes
- "src/libcommon/utility/utility_linux.cpp": new logDirectoryPath() implementation
- "src/libcommon/utility/utility.cpp": removed Linux from the shared logDirectoryPath() block since Linux now has its own implementation in the platform-specific file

### DIFF
--- a/src/libcommon/utility/utility_linux.cpp
+++ b/src/libcommon/utility/utility_linux.cpp
@@ -184,10 +184,12 @@ std::string CommonUtility::fileSystemName(const SyncPath &targetPath) {
 }
 
 ExitInfo CommonUtility::logDirectoryPath(SyncPath &directoryPath) noexcept {
-    // XDG Base Directory Specification: use $XDG_STATE_HOME, fallback to ~/.local/state
+    // XDG Base Directory Specification: use $XDG_STATE_HOME if it is absolute,
+    // otherwise fallback to ~/.local/state
     const auto xdgStateHome = envVarValue("XDG_STATE_HOME");
-    if (!xdgStateHome.empty()) {
-        directoryPath = SyncPath(xdgStateHome);
+    const auto xdgStateHomePath = SyncPath(xdgStateHome);
+    if (!xdgStateHome.empty() && xdgStateHomePath.is_absolute()) {
+        directoryPath = xdgStateHomePath;
     } else {
         auto homeDir = envVarValue("HOME");
         if (homeDir.empty()) homeDir = getHomeDirFromPasswd();
--- a/src/libcommon/utility/utility_linux.cpp
+++ b/src/libcommon/utility/utility_linux.cpp
@@ -1,18 +1,20 @@
-// Infomaniak kDrive - Desktop
-// Copyright (C) 2023-2026 Infomaniak Network SA
-//
-// This program is free software: you can redistribute it and/or modify
-// it under the terms of the GNU General Public License as published by
-// the Free Software Foundation, either version 3 of the License, or
-// (at your option) any later version.
-//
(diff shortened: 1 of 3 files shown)


src/libcommon/utility/utility_linux.cpp
```diff
@@ -171,4 +183,33 @@ std::string CommonUtility::fileSystemName(const SyncPath &targetPath) {
...
+    } else {
+        auto homeDir = envVarValue("HOME");
+        if (homeDir.empty()) homeDir = getHomeDirFromPasswd();
+        if (homeDir.empty()) return {ExitCode::SystemError, ExitCause::NotFound};
+        directoryPath = SyncPath(homeDir) / ".local" / "state";
+    }
+
+    directoryPath /= Str2SyncName(APPLICATION_NAME);
+    directoryPath /= "logs";
+
```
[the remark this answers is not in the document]
ClementKunz: I am not sure that we want to mess with the XDG_STATE_HOME and HOME environment variables. This could have impacts on other processes if the test crashes or fails without resetting the variables to their original value.

src/libcommon/utility/utility_linux.cpp
```diff
@@ -1,41 +1,53 @@
...
+        getpwuid_r(getuid(), &pwd, buf.data(), buf.size(), &result) == 0 && result != nullptr) {
+        return std::string(result->pw_dir);
     }
-    SyncPath homePath(homeDir);
+    return {};
+}
 
+SyncPath CommonUtility::getGenericAppSupportDir() {
+    auto homeDir = CommonUtility::envVarValue("HOME");
+    if (homeDir.empty()) homeDir = getHomeDirFromPasswd();
```
luc-guyot-infomaniak: **Nice to have**: An explanation about this fallback mechanism (when is it useful?) and which tells in particular the nature of the involved password.

src/libcommon/utility/utility_linux.cpp
```diff
@@ -171,4 +183,41 @@ std::string CommonUtility::fileSystemName(const SyncPath &targetPath) {
...
+        if (homeDir.empty()) return {ExitCode::SystemError, ExitCause::NotFound};
+        directoryPath = SyncPath(homeDir) / ".local" / "state";
+    }
+
+    directoryPath /= Str2SyncName(APPLICATION_NAME);
+    directoryPath /= "logs";
+
+    std::error_code ec;
+    const bool directoryPathExists = std::filesystem::exists(directoryPath, ec);
+    if (ec.value() != 0) {
```
luc-guyot-infomaniak:
```suggestion
    if (ec) {
```

src/libcommon/utility/utility_linux.cpp
ClementKunz: efef459c796dca092363b17e5fc848080ec30cca

src/libcommon/utility/utility_linux.cpp
ClementKunz: efef459c796dca092363b17e5fc848080ec30cca

luc-guyot-infomaniak (approved):

PR #1947, opened 2026-05-27
feat(macOS): Display errors

valentinperignon:
CleanShot 2026-06-02 at 10 10 39@2x


src/gui4/macOS/kDrive/UI/Views/MainWindow/Errors/Components/UnknownErrorCellView.swift
```diff
@@ -0,0 +1,84 @@
...
+struct UnknownErrorCellView: View {
+    let error: SynchroError
+
+    var body: some View {
+        VStack(alignment: .leading, spacing: AppPadding.padding8) {
+            ErrorCellView(
+                title: KDriveLocalizable.defaultErrorTitle,
+                description: KDriveLocalizable.unexpectedErrorTeachingTipContent,
+                action: .init(title: KDriveLocalizable.buttonContactSupport) {
+                    // TODO:
```
adrien-coye, line 48: Todo probably sorted in the next PR

src/gui4/macOS/kDrive/UI/Views/MainWindow/Errors/ErrorCellFactory.swift
```diff
@@ -0,0 +1,312 @@
...
+        case .fileLocked:
+            return makeCell(error: error,
+                            title: KDriveLocalizable.errFileLockedTitle,
+                            description: KDriveLocalizable.errFileLockedDescription)
+        case .fileRescued:
+            return makeCell(error: error,
+                            title: KDriveLocalizable.errFileRescuedTitle,
+                            description: KDriveLocalizable.errFileRescuedDescription,
+                            action: .init(title: KDriveLocalizable.buttonOpenFolder) {
+                                // TODO:
```
adrien-coye, line 91: todos of this file probably sorted on the next PR

src/gui4/macOS/kDrive/Utils/UISynchroErrorCategory.swift
```diff
@@ -17,26 +17,50 @@
...
         case .synchronizationDirectories:
-            return "!Dossiers de synchronisation"
+            return KDriveLocalizable.labelSyncFolder
         case .storage:
-            return "!Stockage"
+            return KDriveLocalizable.tabTitleStorage
         }
     }
+
+    public var order: Int {
```
adrien-coye, line 48: Okay it's already a string, no way to do otherwise

src/gui4/macOS/kDriveCore/Extensions/String+Extension.swift
```diff
@@ -0,0 +1,25 @@
...
+ MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ GNU General Public License for more details.
+
+ You should have received a copy of the GNU General Public License
+ along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+import Foundation
+
+public extension String {
```
adrien-coye, line 21: A good candidate to go to swift-foundations, or maybe not if we only use it here.

adrien-coye (commented):
I reviewed in full this PR and tested the code. All good at the moment.

src/gui4/macOS/kDrive/Utils/UISynchroErrorCategory.swift
adrien-coye: Okay so you can use `CaseIterable` and then do something like this:


```
    static func < (lhs: RowType, rhs: RowType) -> Bool {
        guard
            let lhsIndex = allCases.firstIndex(of: lhs),
            let rhsIndex = allCases.firstIndex(of: rhs)
        else {
            return false
        }

        return lhsIndex < rhsIndex
    }
```

It would be a bit slower but for a few cases _and_ the added value of not having to maintain a list I would personally go for this one I think.

adrien-coye: okay lets continue as is.

adrien-coye (approved):
Reviewed latest changes. All good to me 👍

PR #1970, opened 2026-06-03
feat: limit concurent folder size requests

herve-er:


.github/copilot-instructions.md
```diff
@@ -30,4 +30,4 @@ Before proposing code or review comments:
 - Do not duplicate SonarCloud findings verbatim unless they identify a concrete actionable problem in the changed code.
 
 ## Pull Request & Commit Rules (MANDATORY)
-See `AGENTS.md` for the canonical version.
```
ClementKunz, line 33: Shouldn't we ask to follow the rules defined in the "Pull Requests & Commits" section in `AGENTS.md`. So we have only 1 source of truth.

src/gui4/windows/kDrive client/kDrive client/CustomControls/FolderTreeItemTemplateSelectorBase.cs
```diff
@@ -0,0 +1,45 @@
...
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+using Microsoft.UI.Xaml;
+using Microsoft.UI.Xaml.Controls;
+
+namespace Infomaniak.kDrive.CustomControls
+{
+    public abstract class FolderTreeItemTemplateSelectorBase<TTreeItem> : DataTemplateSelector
```
ClementKunz, line 23: Why call it `TTreeItem`? The name is not very explicit.

.github/copilot-instructions.md
herve-er: Since the last version, Visual Studio uses this file for commit/PR message generation, and it does not seem to be able to look at another file.

src/gui4/windows/kDrive client/kDrive client/CustomControls/FolderTreeItemTemplateSelectorBase.cs
herve-er: It's the C# naming convention.

See: https://learn.microsoft.com/en-us/dotnet/csharp/fundamentals/coding-style/identifier-names
Section: "Type parameter naming guidelines"

I guess `T` stands for templated.

.github/copilot-instructions.md
ClementKunz: That's a shame since there are many more rules in our `AGENTS.md` file.
The VS code [doc](https://code.visualstudio.com/docs/agent-customization/custom-instructions) seems to say that the feature is still experimental. That is probably explaining why is does not work for now.

However, I would add anyway another line like: `Follow the project's rules from section "Pull Requests & Commits" in [`AGENTS.md`](https://github.com/Infomaniak/desktop-kDrive/blob/develop/AGENTS.md).` is the case Copilot will be able to read the `AGENTS.md` file in the future.

What do you think?

herve-er: done here: https://github.com/Infomaniak/desktop-kDrive/pull/1970/commits/c2fab8184e40416513ed9f418b9b754f83d05ae1

ClementKunz (approved):

PR #1999, opened 2026-06-09
feat(remote-file-system-observer): Puts RFSO to sleep while syncing

luc-guyot-infomaniak:
These changes ensure that no HTTP requests of type `longpoll` is sent by the `RemoteFileSystemObserver` as long as the synchronization has left the `Idle` state and has not reached the `Done` state yet.

### DIFF
--- a/src/libsyncengine/syncpal/isyncworker.h
+++ b/src/libsyncengine/syncpal/isyncworker.h
@@ -62,6 +62,8 @@ class ISyncWorker {
         } // Minimum pause duration is 1 min
         void resetPauseDuration() { _pauseDuration = defaultPauseDuration; }
 
+        virtual void resume() { start(); }
+
         void setTesting(bool testing) { _testing = testing; }
         [[nodiscard]] std::shared_ptr<CacheDirectory> cacheDirectory() const { return _syncPal->cacheDirectory(); }
 
@@ -69,8 +71,10 @@ class ISyncWorker {
         log4cplus::Logger _logger;
         std::shared_ptr<SyncPal> _syncPal;
 
-        bool _testing{false};
         virtual void init();
+        void startExecutionThread();
+
+        bool _testing{false};
 
     protected:
         //! Wait for a delay. Allows to postpone the start of the worker to smooth the load.
--- a/src/libsyncengine/syncpal/isyncworker.h
+++ b/src/libsyncengine/syncpal/isyncworker.h
@@ -50,7 +50,7 @@ class ISyncWorker {
         bool isRunning() const;
         bool stopAsked() const;
 
-        void setIsRunning(bool isRunning);
+        void setRunningFlagValue(bool isRunning);
         void setStopAsked(bool stopAsked);
 
         ExitCode exitCode() const { return _exitCode; }
--- a/src/libsyncengine/syncpal/isyncworker.h
+++ b/src/libsyncengine/syncpal/isyncworker.h
@@ -35,7 +35,7 @@ const int64_t defaultPauseDuration = 60000; // 1 min
 class ISyncWorker {
     public:
         ISyncWorker(std::shared_ptr<SyncPal> syncPal, const std::string &name, const std::string &shortName,
-                    const std::chrono::seconds &startDelay = std::chrono::seconds(0), bool testing = false);
+                    const std::chrono::seconds &startDelay = std::chrono::seconds(0));
         virtual ~ISyncWorker();
 
         virtual void start();
@@ -64,7 +64,6 @@ class ISyncWorker {
 
         virtual void resume() { start(); }
 
-        void setTesting(bool testing) { _testing = testing; }
         [[nodiscard]] std::shared_ptr<CacheDirectory> cacheDirectory() const { return _syncPal->cacheDirectory(); }
 
     protected:
@@ -73,8 +72,6 @@ class ISyncWorker {
 
         virtual void init();
         void startExecutionThread();
-
-        bool _testing{false};
 
     protected:
         //! Wait for a delay. Allows to postpone the start of the worker to smooth the load.
--- a/src/libsyncengine/syncpal/syncpalworker.cpp
+++ b/src/libsyncengine/syncpal/syncpalworker.cpp
@@ -85,11 +85,12 @@ bool SyncPalWorker::shouldBePaused(const std::shared_ptr<ISyncWorker> w1, const 
 
     const auto networkIssue =
             (w1 && w1->exitCode() == ExitCode::NetworkError) || (w2 && w2->exitCode() == ExitCode::NetworkError);
-    const auto httpBlockingError =
-            (w1 && w1->exitCode() == ExitCode::BackError &&
-             (w1->exitCause() == ExitCause::Http5xx || w1->exitCause() == ExitCause::HttpErr || w1->exitCause() == ExitCause::MissingReplyData)) ||
-            (w2 && w2->exitCode() == ExitCode::BackError &&
-             (w2->exitCause() == ExitCause::Http5xx || w2->exitCause() == ExitCause::HttpErr || w2->exitCause() == ExitCause::MissingReplyData));
+    const auto httpBlockingError = (w1 && w1->exitCode() == ExitCode::BackError &&
+                                    (w1->exitCause() == ExitCause::Http5xx || w1->exitCause() == ExitCause::HttpErr ||
+                                     w1->exitCause() == ExitCause::MissingReplyData)) ||
+                                   (w2 && w2->exitCode() == ExitCode::BackError &&
+                                    (w2->exitCause() == ExitCause::Http5xx || w2->exitCause() == ExitCause::HttpErr ||
+                                     w2->exitCause() == ExitCause::MissingReplyData));
 
     const auto syncDirNotAccessible =
             (w1 && w1->exitCode() == ExitCode::SystemError &&
@@ -236,6 +237,30 @@ ExitInfo SyncPalWorker::ensureBlackListIsPropagated() {
     return ExitCode::Ok;
 }
 
+void SyncPalWorker::startFsoWorkerIfNeeded(std::shared_ptr<ISyncWorker> fsoWorker, bool &isFsoInProgress, bool &syncDirChanged) {
+    if (fsoWorker == nullptr || fsoWorker->isRunning()) return;
+
+    if (isFsoInProgress) {
+        LOG_SYNCPAL_DEBUG(_logger, "Stop FSO worker " << fsoWorker->name());
+        isFsoInProgress = false;
+        stopAndWaitForExitOfWorker(fsoWorker);
+        if (shouldBePaused(fsoWorker, nullptr) && !pauseAsked()) {
+            pause();
+            return;
+        }
+
+        syncDirChanged = fsoWorker->exitCode() == ExitCode::DataError && fsoWorker->exitCause() == ExitCause::SyncDirChanged;
+        if (syncDirChanged) return;
+
+        if (shouldBeStopped(fsoWorker, nullptr) && !stopAsked()) stop();
+
+    } else if (!pauseAsked()) {
+        LOG_SYNCPAL_DEBUG(_logger, "Start FSO worker " << fsoWorker->name());
+        isFsoInProgress = true;
+        fsoWorker->start();
+    }
+}
+
 void SyncPalWorker::execute() {
     ExitCode exitCode(ExitCode::Unknown);
     LOG_SYNCPAL_INFO(_logger, "Worker " << name() << " started");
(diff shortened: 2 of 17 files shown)


src/libsyncengine/syncpal/isyncworker.h
```diff
@@ -62,15 +62,19 @@ class ISyncWorker {
...
 
     protected:
         log4cplus::Logger _logger;
         std::shared_ptr<SyncPal> _syncPal;
 
-        bool _testing{false};
         virtual void init();
+        void startExecutionThread();
+
+        bool _testing{false};
```
ClementKunz: Looks like this is not used anymore

test/libsyncengine/syncpal/testsyncpalworker.cpp
```diff
@@ -150,6 +147,13 @@ void TestSyncPalWorker::testInternalPause1() {
     const auto mockLfso = mockSyncPal->getMockLFSOWorker();
     const auto mockRfso = mockSyncPal->getMockRFSOWorker();
     const auto syncpalWorker = mockSyncPal->getSyncPalWorker();
+    const auto mockExecutorWorker = mockSyncPal->getMockExecutorWorker();
+
+    // Simulate a network error in Executor
+    // mockExecutorWorker->setMockExecuteCallback([]() -> ExitInfo { return ExitInfo(ExitCode::BackError, ExitCause::HttpErr); });
+
+    // Simulate a local file system event to trigger a sync cycle and the network error in the executor.
+    // mockLfso->simulateFSEvent();
```
ClementKunz: So we do not run those tests?

src/libsyncengine/syncpal/syncpalworker.cpp
```diff
@@ -236,6 +297,71 @@ ExitInfo SyncPalWorker::ensureBlackListIsPropagated() {
...
+    static const std::string resumeMessage = "Resuming RFSO until the synchronization leaves the Idle state.";
+
+    switch (_step) {
+        case SyncStep::Idle: {
+            if (!rfsoWorker.isResumable) {
+                startFSOWorker(rfsoWorker);
+                rfsoWorker.isResumable = true;
+            } else {
+                LOG_DEBUG(_logger, resumeMessage);
+                rfsoWorker.worker->resume();
```
ClementKunz, line 300: What's the difference between `startFSOWorker` and `rfsoWorker.worker->resume()`? They both only call the worker's `start` right?

src/libsyncengine/syncpal/syncpalworker.cpp
```diff
@@ -52,81 +52,142 @@ SyncPalWorker::SyncPalWorker(std::shared_ptr<SyncPal> syncPal, const std::string
...
+            return true;
+        }
+    }
+
+    return false;
 }
 
-bool shouldBeStopped(const std::shared_ptr<ISyncWorker> w1, const std::shared_ptr<ISyncWorker> w2 = nullptr) {
-    const auto dbError = (w1 && w1->exitCode() == ExitCode::DbError) || (w2 && w2->exitCode() == ExitCode::DbError);
-    const auto systemError = (w1 && w1->exitCode() == ExitCode::SystemError) || (w2 && w2->exitCode() == ExitCode::SystemError);
```
ClementKunz, line 68: What about the system errors? Should still stop the sync in this case or the logic has changed?

src/libsyncengine/syncpal/syncpalworker.h
```diff
@@ -39,15 +39,26 @@ class SyncPalWorker : public ISyncWorker {
...
-        inline std::chrono::time_point<std::chrono::steady_clock> pauseTime() const { return _pauseTime; }
+        SyncStep step() const { return _step; }
+        std::chrono::time_point<std::chrono::steady_clock> pauseTime() const { return _pauseTime; }
         static std::string stepName(SyncStep step);
 
+        struct Worker {
+                std::shared_ptr<ISyncWorker> worker{nullptr};
+                bool isInProgress{false};
+                bool isResumable{false};
+                ReplicaSide side{ReplicaSide::Both};
```
ClementKunz: Should be initialized to `Unknown` so we can detect eventual bad initialization?

src/libsyncengine/syncpal/syncpalworker.cpp
```diff
@@ -485,134 +569,104 @@ std::string SyncPalWorker::stepName(SyncStep step) {
...
     }
     sentry::syncStepToPTrace(_step, syncDbId())->stop();
     sentry::syncStepToPTrace(step, syncDbId())->start();
 
     _step = step;
+
+    workers = {{ReplicaSide::Local, {.side = ReplicaSide::Local}},
+               {ReplicaSide::Remote, {.side = ReplicaSide::Remote}},
+               {ReplicaSide::Both, {.side = ReplicaSide::Both}}};
+    inputSharedObjects = {{ReplicaSide::Local, nullptr}, {ReplicaSide::Remote, nullptr}};
```
ClementKunz: Using designated initializers does not make this very readable  IMO. Perhaps defining a constructor for `Worker` that takes the `side` as an argument would simplify this notation?

src/libsyncengine/syncpal/syncpalworker.cpp
```diff
@@ -52,81 +52,142 @@ SyncPalWorker::SyncPalWorker(std::shared_ptr<SyncPal> syncPal, const std::string
...
+bool SyncPalWorker::handleRateLimited(const SyncPalWorker::ReplicaWorkers &workers) {
+    auto newPauseDuration = pauseDuration();
+    bool shouldHandleRateLimit = false;
+
+    for (const auto side: {ReplicaSide::Local, ReplicaSide::Remote, ReplicaSide::Both}) {
+        if (const bool hasWorker = workers.contains(side) && workers.at(side).worker; !hasWorker) continue;
+
+        if (workers.at(side).worker->exitCode() == ExitCode::RateLimited) shouldHandleRateLimit = true;
+
+        newPauseDuration = std::max(newPauseDuration, workers.at(side).worker->pauseDuration());
```
ClementKunz: You could break here

src/libsyncengine/syncpal/syncpalworker.cpp
```diff
@@ -52,81 +52,142 @@ SyncPalWorker::SyncPalWorker(std::shared_ptr<SyncPal> syncPal, const std::string
...
+    }
+
+    if (shouldHandleRateLimit && newPauseDuration != pauseDuration()) {
+        LOG_SYNCPAL_INFO(_logger, "Changing pause duration to " << newPauseDuration << " ms due to rate limiting");
+        setPauseDuration(newPauseDuration);
+    }
```
ClementKunz: And maybe put this part in the for loop so we do not need `shouldHandleRateLimit`

src/libsyncengine/syncpal/syncpalworker.cpp
```diff
@@ -236,6 +297,71 @@ ExitInfo SyncPalWorker::ensureBlackListIsPropagated() {
...
+        LOG_SYNCPAL_DEBUG(_logger, "Stop FSO worker " << fsoWorker.worker->name());
+        fsoWorker.isInProgress = false;
+        stopAndWaitForExitOfWorker(fsoWorker.worker);
+        if (shouldBePaused({{fsoWorker.side, fsoWorker}}) && !pauseAsked()) {
+            pause();
+            return;
+        }
+
+        syncDirChanged = worker->exitCode() == ExitCode::DataError && worker->exitCause() == ExitCause::SyncDirChanged;
+        if (syncDirChanged) return;
```
ClementKunz, line 341: Is it really simpler to stop and restart the FSO workers than just setting a flag or something that stop the sending of longpoll during the sync?
Restarting the worker means you have to make sure that the snapshot is not re-generated, you will send again the getSpecialFoldersRemoteIds requests, ...

src/libsyncengine/syncpal/syncpalworker.cpp
```diff
@@ -236,6 +297,71 @@ ExitInfo SyncPalWorker::ensureBlackListIsPropagated() {
...
+            return;
+    }
+}
+
+void SyncPalWorker::adaptFSOWorkerActivityToSyncState(Worker &fsoWorker, bool &syncDirChanged) {
+    auto worker = fsoWorker.worker;
+
+    if (worker == nullptr || worker->isRunning()) return;
+
+    if (fsoWorker.isInProgress) {
```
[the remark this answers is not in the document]
luc-guyot-infomaniak: Addressed with f0c66cc7118fd15aedf6f6a7c027b1028726129e

src/libsyncengine/syncpal/syncpalworker.cpp
```diff
@@ -236,6 +297,71 @@ ExitInfo SyncPalWorker::ensureBlackListIsPropagated() {
...
+            rfsoWorker.isInProgress = false;
+            break;
+        }
+        case SyncStep::Done: {
+            LOG_DEBUG(_logger, resumeMessage);
+            rfsoWorker.worker->resume();
+            break;
+        }
+        default:
+            return;
```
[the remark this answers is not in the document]
luc-guyot-infomaniak: Addressed with f0c66cc7118fd15aedf6f6a7c027b1028726129e.

test/libsyncengine/syncpal/testsyncpalworker.cpp
```diff
@@ -150,6 +147,13 @@ void TestSyncPalWorker::testInternalPause1() {
...
 
     // Simulate a network error in RFSO
```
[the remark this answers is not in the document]
luc-guyot-infomaniak: Addressed with e3317973267d2b0bfda45dae97a6643b18788948.

test/libsyncengine/syncpal/testsyncpalworker.cpp
luc-guyot-infomaniak: These lines are obsolete and have been removed (see e3317973267d2b0bfda45dae97a6643b18788948): I had written them at a time when I thought that RFSO could not be responsible for pausing the simulation anymore, but it turns out that it can.

src/libsyncengine/syncpal/isyncworker.h
luc-guyot-infomaniak: The unused `_testing` flag is now removed: 5a918468147d4863d1d2d91f76ca8e5111404065.

src/libsyncengine/syncpal/syncpalworker.cpp
luc-guyot-infomaniak: Modified here : 80dfad75ba070b00258758563f6b3d7f4c1e3d60.

src/libsyncengine/syncpal/syncpalworker.cpp
luc-guyot-infomaniak: Modified here : 80dfad75ba070b00258758563f6b3d7f4c1e3d60.

src/libsyncengine/syncpal/syncpalworker.h
luc-guyot-infomaniak: Done here: 80dfad75ba070b00258758563f6b3d7f4c1e3d60

src/libsyncengine/syncpal/syncpalworker.cpp
luc-guyot-infomaniak: Done here : 80dfad75ba070b00258758563f6b3d7f4c1e3d60

src/libsyncengine/syncpal/syncpalworker.cpp
luc-guyot-infomaniak: Thanks, that's a mistake.  Fixed with a99632c6dd2ce882d2889919b4b07935353d3707.

src/libsyncengine/syncpal/syncpalworker.cpp
luc-guyot-infomaniak: No. The `resume()` method does not `init()` the FSO. By calling `init()`, we tell the RFSO that there is no need to make `longpoll` requests nor `listing/continue` requests after the initial construction of the snapshot. 

This behaviour is kept unchanged. When resuming, the RFSO should make `longpoll` requests. This is why `init()` should be called.

src/libsyncengine/syncpal/syncpalworker.cpp
luc-guyot-infomaniak: > Is it really simpler to stop and restart the FSO workers than just setting a flag or something that stop the sending of longpoll during the sync?

Yes, I think so: 
- the state of the RFSO is then easier to track as it is either running or stopped. We avoid creating an intermediate step which would not be easy to interpret. (There are already several flags such as `_initializing`, `_updating`, `_stopAsked` which deactivate parts or the whole observer. It is not always straightforward to track how they actually act on the `SyncPalWorker` and `RemoteFilesystemObserver` workflows).
- If the RFSO is not stopped, we have to keep it asleep by means of a loop with some explicit sleep durations (it consumes CPU time) or a `wait_condition`. This new mechanism adds complexity.

src/libsyncengine/syncpal/syncpalworker.cpp
ClementKunz: I am not sure that I understand. The default implementation of `resume` in `ISyncWorker` only call `start` which does the init, while the overrode version of `resume` in `RemoteFileSystemObserverWorker` does basically the same things (except for the update flag). Maybe I am missing something, but I do not even see the need for overriding the `resume` method.

Image

luc-guyot-infomaniak: There is more in the `init`method of `FileSystemObserverWorker`:

```
void FileSystemObserverWorker::init() {
    ISyncWorker::init();
    setUpdateFlagValue(false);
    setInitFlagValue(true);
    invalidateSnapshot();
}
```
We need to avoid `setInitFlagValue(true)` and `invalidateSnapshot()` in `FileSystemObserverWorker::resume()`.

There is no actual need to define a virtual `resume` function, but this allows me to avoid dynamic casts in `syncpalworker.cpp` (we usually prefer polymorphism over dynamic cast, but we can debate in this particular case).

src/libsyncengine/syncpal/syncpalworker.cpp
ClementKunz: From my point of view, the complexity and regression risk are even higher here since you introduce more flags, such as `isInProgress` (duplicate of `_isRunning` from `ISyncWorker`?) or `isResumable`.
I am not against those changes, but I just think they're overly complicated. We should tend to simplify the code, not make it more and more complex (which we do over and over for years now 😅 )

luc-guyot-infomaniak: - The flag `isInProgress` is not a new flag: it is an old flag that has been put inside the `Worker` struct.
- The flag `isResumable` is indeed a new flag used to differentiate a `start` action from a `resume` action. But it is only handled by the `SyncPalWorker`; we keep the changes in `RemoteFileSystemObserver` to the minimum. 
- I believe, on the contrary, that maintaining a simple ON-OFF dichotomy for the RFSO is clearer that maintaining a form of idleness.

ClementKunz (approved):

PR #2006, opened 2026-06-10
fix: deploy Conan libs

ChristopheLarchier:


adrien-coye (approved):
I did test this branch on my Mac and built an installer package, then use the package and run the app. I was able to confirm it works fine 👍

PR #2010, opened 2026-06-10
feat: hash match to upload job

RaphaelOstier:
This pull request refactors and improves the upload logic to avoid unnecessary uploads when the file content has not changed, relying on hash comparisons between local and remote files. It introduces new logic to update only the modification date when the file content matches, and adds comprehensive tests for both the matching and mismatching checksum scenarios. The changes also replace the `shouldDownload` concept with a clearer `hashMatch` flag.

**Upload logic improvements:**

* Added a new method `resolveUploadNeed` in `UploadJob` to determine if an upload is necessary by comparing file hashes, and to update file modification dates without uploading when the content matches. [[1]](diffhunk://#diff-a107392671105321c8d38ee6317c46cf2e5f8106323be36da3cbcd80c429f908R100-R168) [[2]](diffhunk://#diff-97e3b56e2735902fde69aa6f18322faf9723905e7d4d47167dc611332544ee06R53-R60) [[3]](diffhunk://#diff-97e3b56e2735902fde69aa6f18322faf9723905e7d4d47167dc611332544ee06R82-R84)
* Refactored `runJob` in `UploadJob` to use `resolveUploadNeed` and only proceed with uploading if needed. [[1]](diffhunk://#diff-a107392671105321c8d38ee6317c46cf2e5f8106323be36da3cbcd80c429f908R100-R168) [[2]](diffhunk://#diff-97e3b56e2735902fde69aa6f18322faf9723905e7d4d47167dc611332544ee06R53-R60)
* Added `applyFileDates` to update file metadata on the server without uploading the file when hashes match. [[1]](diffhunk://#diff-a107392671105321c8d38ee6317c46cf2e5f8106323be36da3cbcd80c429f908R100-R168) [[2]](diffhunk://#diff-97e3b56e2735902fde69aa6f18322faf9723905e7d4d47167dc611332544ee06R53-R60)

**CheckHashMatchJob refactor:**

* Replaced the `shouldDownload` flag with a `hashMatch` flag for clarity, updated related method names, and set the flag appropriately in `runJob`. [[1]](diffhunk://#diff-f0df3dba286f0b2426ee319f60fc0b9a8904980ef761db5ad0b715b6045dde2bL31-R31) [[2]](diffhunk://#diff-f0df3dba286f0b2426ee319f60fc0b9a8904980ef761db5ad0b715b6045dde2bL50-R50) [[3]](diffhunk://#diff-70ec71f86a41cad2be15ebadd9a5683169ff5e10b69c5e7e51e233a122346711R68) [[4]](diffhunk://#diff-70ec71f86a41cad2be15ebadd9a5683169ff5e10b69c5e7e51e233a122346711L108-R109)
* Updated tests to use the new `hashMatch` logic instead of `shouldDownload`.

**Testing improvements:**

* Added new tests for upload scenarios: one where the checksum mismatches (forcing an upload), and one where the checksum matches (skipping the upload and only updating the modification date). [[1]](diffhunk://#diff-a44011b4b2142730cdb600b754fc9df3f32479959030cd13cf167fda5bfc52ebR1309-R1391) [[2]](diffhunk://#diff-7cb4981dbd48c6446082f1f1058fe4f92be59b6f6ef5d9dfe5dda249e0618330R58-R59) [[3]](diffhunk://#diff-7cb4981dbd48c6446082f1f1058fe4f92be59b6f6ef5d9dfe5dda249e0618330R106-R107)

These changes improve efficiency by preventing redundant uploads and clarify the upload decision logic.


src/libsyncengine/jobs/network/kDrive_API/upload/uploadjob.cpp
```diff
@@ -98,13 +97,75 @@ ExitInfo UploadJob::canRun() {
     return ExitCode::Ok;
 }
 
+ExitInfo UploadJob::resolveUploadNeed() {
+    _shouldUpload = true;
+    FileStat fileStat;
+    IoError ioError = IoError::Success;
+    if (!IoHelper::getFileStat(_absoluteFilePath, &fileStat, ioError, IoHelper::PathCheckOption::Insensitive) ||
+        ioError != IoError::Success) {
+        LOGW_WARN(_logger, L"CheckHashMatchJob: failed to get file size for " << Utility::formatSyncPath(_absoluteFilePath));
```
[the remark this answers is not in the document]
ClementKunz: You should address this

src/libsyncengine/jobs/network/kDrive_API/upload/uploadjob.cpp
```diff
@@ -98,13 +97,75 @@ ExitInfo UploadJob::canRun() {
...
+        return ExitCode::Ok; // Non-fatal: fall through to upload
+    }
+    CheckHashMatchJob hashJob(driveDbId(), _absoluteFilePath, _fileId, fileStat.size);
```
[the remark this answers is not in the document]
ClementKunz: You are extracting the local size here and again inside the job. Where are you retrieving the remote size?
IMO, `CheckHashMatchJob` should not take the size as input; it should only deal with the checksum. You should run this job only if the size differ.

src/libsyncengine/jobs/network/kDrive_API/upload/uploadjob.cpp
```diff
@@ -98,13 +97,75 @@ ExitInfo UploadJob::canRun() {
...
+        }
+    }
+    return ExitCode::Ok;
+}
+
+void UploadJob::computeHydrationStatus() {
+    VfsStatus vfsStatus;
+    if (_vfs) (void) _vfs->status(_absoluteFilePath, vfsStatus);
+    _isHydrated = !_vfs || vfsStatus.isHydrated;
+}
```
ClementKunz: This method is useless. It is called in 1 place and check again `_vfs` whereas it is already in `runJob`.
IMO you should keep this code directly in `runJob`

src/libsyncengine/jobs/network/kDrive_API/upload/uploadjob.cpp
```diff
@@ -98,13 +97,75 @@ ExitInfo UploadJob::canRun() {
...
+
+ExitInfo UploadJob::runJob() noexcept {
+    if (!_fileId.empty() && _vfs) {
+        computeHydrationStatus();
+        if (_isHydrated) {
+            const ExitInfo exitInfo = resolveUploadNeed();
+            if (!_shouldUpload && exitInfo) return ExitCode::Ok;
```
ClementKunz: `_shouldUpload` should be a local variable, not a member of the class. And therefore, `resolveUploadNeed` should return whether we should upload or not (as an argument passed by reference for example).
Always keep the following rule in mind: a variable should always have the narrowest possible scope.

src/libsyncengine/jobs/network/kDrive_API/upload/uploadjob.cpp
```diff
@@ -98,13 +97,75 @@ ExitInfo UploadJob::canRun() {
...
+ExitInfo UploadJob::applyFileDates() {
+    PostFileModificationDateJob postJob(driveDbId(), _fileId, _modificationTimeIn);
+    const ExitInfo exitInfo = postJob.runSynchronously();
+    if (!exitInfo) {
+        LOGW_DEBUG(_logger, L"PostFileModificationDateJob failed: " << exitInfo);
+        return exitInfo;
+    }
+    IoError ioError = IoError::Success;
+    uint64_t fileSize = 0;
+    IoHelper::getFileSize(_absoluteFilePath, fileSize, ioError);
```
ClementKunz: As indicated by Sonar, you should handle errors the might be returned by `IoHelper::getFileSize`

ClementKunz: Moreover, you already retrieved the file size in `resolveUploadNeed()`, no need to do it again here. Pass it as an argument in the function.

src/libsyncengine/jobs/network/kDrive_API/upload/uploadjob.h
```diff
@@ -73,6 +79,9 @@ class UploadJob : public AbstractTokenNetworkJob {
         SyncPath _linkTarget;
         NodeType _targetType = NodeType::File;
 
+        bool _shouldUpload = true;
+        bool _isHydrated;
```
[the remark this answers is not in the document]
ClementKunz: Always initialize your variable that does not have a default constructor (i.e. all basic types)

src/libsyncengine/jobs/network/kDrive_API/checkhashmatchjob.cpp
ClementKunz, line 79: As said before, you do not care about the size in this job

ClementKunz, line 85: This is strange, here you:
- send a synchronous request to the backend
- compute the local checksum (which might takes some time to compute)
- then process the backend's reply

IMO you should first compute the local checksum, so you do not have to send the request if anything goes wrong. To make the function even more readable, you might move the logic about computing the local checksum and handling related errors to a dedicated private method.

ClementKunz: And, to make it even better and more efficient, the local and remote checksum computation should be done in parallel. But this should be the topic of a future task. Keep it simple for now, and once everything works perfectly, we can complexify it.

test/libsyncengine/jobs/network/testnetworkjobs.cpp
ClementKunz, line 1: I think you mention the fact that the backend only accepts modification dates with a 24h offset in the future. You should test the behavior in the case the user's file has a date far into the future.

src/libsyncengine/jobs/network/kDrive_API/upload/uploadjob.cpp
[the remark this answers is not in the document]
herve-er: The decision to provide `CheckHashMatch` with the remote size and let it handle the size comparison logic was made to centralize hash comparison logic in a single job, instead of duplicating it across `DownloadJob`, `UploadJob`, and `UploadSessionJob`.

From my point of view, comparing file sizes can be considered part of the responsibility of `CheckHashMatchJob`. It is simply a shortcut to detect that the hash does not match.


(Anyhow the size provided to the CheckHashMatchJob should be the remoteSize not the localSize @RaphaelOstier  

```cpp
CheckHashMatchJob(DriveDbId driveDbId, const SyncPath &filePath, const NodeId &nodeId,
                          const int64_t remoteSize);
```
)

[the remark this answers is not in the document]
ClementKunz: I think a job should always do as little as possible and that the local size might already be available outside the job (at least for the upload job), so we do not need another IO access.
However, as long as we fix the logic here, I am ok with this implementation.

src/libsyncengine/jobs/network/kDrive_API/upload/uploadjob.cpp
RaphaelOstier: The functions come from downloadJob where I used the variable for tests, which I should have done for the upload as well. I don’t really have any other way to know if it’s uploaded or not?

RaphaelOstier: it is also required in the destructor (which I had forgoten)

test/libsyncengine/jobs/network/testnetworkjobs.cpp
RaphaelOstier: Those tests are already in the PostFileModificationDateJob. the tests in upload are only about whether the job should be uploaded or not.
Also, the backend accepts modification dates more than 24h in the future, they used not to with the specific API used in the PostFileModificationDateJob but now, it simply caps the date at 24 hours in the future just like upload route does.

src/libsyncengine/jobs/network/kDrive_API/upload/uploadjob.cpp
RaphaelOstier: not required anymore since the getFileSize isn't used in the resolveUploadNeed() anymore (it was giving the local size when I wanted the remote size)

src/libsyncengine/jobs/network/kDrive_API/upload/uploadjob.cpp
```diff
@@ -98,13 +98,73 @@ ExitInfo UploadJob::canRun() {
...
+        }
+    }
+    return ExitCode::Ok;
+}
+
+ExitInfo UploadJob::runJob() noexcept {
+    if (!_fileId.empty() && _vfs) {
+        VfsStatus vfsStatus;
+        if (_vfs) (void) _vfs->status(_absoluteFilePath, vfsStatus);
+        _isHydrated = !_vfs || vfsStatus.isHydrated;
```
ClementKunz: Why is this an attribute? It is used only here? Should be a local variable.
Moreover, why do you check this? Only hydrated items should be uploaded, so the upload job should not have been created if the file is not hydrated. Do you have a specific case in mind?

src/libsyncengine/jobs/network/kDrive_API/upload/uploadjob.cpp
```diff
@@ -98,13 +98,73 @@ ExitInfo UploadJob::canRun() {
...
+            return exitInfo;
+        }
+    }
+    return ExitCode::Ok;
+}
+
+ExitInfo UploadJob::runJob() noexcept {
+    if (!_fileId.empty() && _vfs) {
+        VfsStatus vfsStatus;
+        if (_vfs) (void) _vfs->status(_absoluteFilePath, vfsStatus);
```
ClementKunz: you already checked the validity of this pointer just above

src/libsyncengine/jobs/network/kDrive_API/upload/uploadjob.cpp
ClementKunz: Why? The file must be hydrated if we've created an upload job. Is it to handle a specific case where the file has been dehydrated for the start of the job?
Please, either justify this logic with a comment in the destructor or make `_shouldUpload` a local variable

test/libsyncengine/jobs/network/testnetworkjobs.cpp
```diff
@@ -1306,6 +1306,89 @@ void TestNetworkJobs::testUpload() {
...
+        CPPUNIT_ASSERT(!nodeId.empty());
+    }
+
+    // Modify the local file content so the hash no longer matches the remote
+    testhelpers::generateOrEditTestFile(localFilePath);
+    const SyncTime newModificationTime = modificationTime + 10;
+
+    // EDIT upload — checksum mismatch expected → must upload
+    {
+        UploadJob editJob(nullptr, _driveDbId, localFilePath, nodeId, newModificationTime);
```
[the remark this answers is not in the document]
ClementKunz: What about this comment? Seems legit to me

test/libsyncengine/jobs/network/testnetworkjobs.cpp
```diff
@@ -1306,6 +1306,89 @@ void TestNetworkJobs::testUpload() {
...
+        const ExitInfo exitInfo = createJob.runSynchronously();
+        CPPUNIT_ASSERT_MESSAGE(toString(exitInfo), exitInfo);
+        nodeId = createJob.nodeId();
+        CPPUNIT_ASSERT(!nodeId.empty());
+    }
+
+    // EDIT upload — same file content, only the modification time differs → must skip upload
+    const SyncTime newModificationTime = modificationTime + 10;
+    {
+        UploadJob editJob(nullptr, _driveDbId, localFilePath, nodeId, newModificationTime);
```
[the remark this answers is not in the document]
ClementKunz: Same as above. Moreover, your test seems useless; it passes even though you did not check the hash here (since `vfs` is null).
You need a way to check that the hashes match and that the upload therefore did not occur.

test/libsyncengine/jobs/network/testnetworkjobs.cpp
```diff
@@ -1306,6 +1306,89 @@ void TestNetworkJobs::testUpload() {
...
+        const ExitInfo exitInfo = editJob.runSynchronously();
+        CPPUNIT_ASSERT_MESSAGE(toString(exitInfo), exitInfo);
+        // No real upload: nodeId returned by the job should be non-empty (set from setOutputParameters)
```
ClementKunz: `_nodeIdOut` is never empty. It is either set in `UploadJob::handleResponse` if the upload occurs, or in `UploadJob::applyFileDates()` if the hashes match. This is not a good test.

src/libsyncengine/jobs/network/kDrive_API/checkhashmatchjob.cpp
```diff
@@ -104,6 +99,15 @@ ExitInfo CheckHashMatchJob::runJob() noexcept {
         LOGW_WARN(_logger, L"Unable to compute checksum for " << Utility::formatIoError(_filePath, checksumIoError));
         return ExitCode::SystemError;
     }
+    return remoteHashMatch();
+}
+
+ExitInfo CheckHashMatchJob::remoteHashMatch() noexcept {
```
ClementKunz: Why is this method called "remote"? It simply runs the job and compares remote AND local hashes.
Moreover, it feels strange, IMO, that `AbstractTokenNetworkJob::runJob()` is called outside the `runJob()` method.
I would prefer to put the call to `AbstractTokenNetworkJob::runJob()` inside `runJob()`. However, you can put all the logic to retrieve the local checksum inside a dedicated helper function. This would make the method `runJob` more readable.

src/libsyncengine/jobs/network/kDrive_API/checkhashmatchjob.cpp
```diff
@@ -104,6 +99,15 @@ ExitInfo CheckHashMatchJob::runJob() noexcept {
...
+    if (const ExitInfo exitInfo = AbstractTokenNetworkJob::runJob(); !exitInfo) {
+        LOGW_DEBUG(_logger, L"Failed to get remote hash for " << Utility::formatSyncPath(_filePath) << L", skipping hash check: "
+                                                              << exitInfo);
+        return exitInfo;
+    }
 
     if (_localHash != _remoteHash) return ExitCode::Ok;
     _hashMatch = true;
```
ClementKunz: This should be in `handleResponse`

src/libsyncengine/jobs/network/kDrive_API/checkhashmatchjob.cpp
RaphaelOstier: because it checks if the remoteHash match the localHash ?

RaphaelOstier: It became useless when moving the run job and hash check.

src/libsyncengine/jobs/network/kDrive_API/upload/uploadjob.cpp
RaphaelOstier: done https://github.com/Infomaniak/desktop-kDrive/pull/2010/commits/61b2993d87e63c0634233f30534157fa8eab85bf

src/libsyncengine/jobs/network/kDrive_API/upload/uploadjob.cpp
RaphaelOstier: done https://github.com/Infomaniak/desktop-kDrive/pull/2010/commits/61b2993d87e63c0634233f30534157fa8eab85bf

test/libsyncengine/jobs/network/testnetworkjobs.cpp
RaphaelOstier: done https://github.com/Infomaniak/desktop-kDrive/pull/2010/commits/f896ddc46078869c24a4d9b4ea8147b72242b816

src/libsyncengine/jobs/network/kDrive_API/checkhashmatchjob.cpp
RaphaelOstier: done https://github.com/Infomaniak/desktop-kDrive/pull/2010/commits/867ada0936d9bf906d5c8586803dd1a975d254b5

src/libsyncengine/jobs/network/kDrive_API/checkhashmatchjob.cpp
RaphaelOstier: done https://github.com/Infomaniak/desktop-kDrive/pull/2010/commits/867ada0936d9bf906d5c8586803dd1a975d254b5

test/libsyncengine/jobs/network/testnetworkjobs.cpp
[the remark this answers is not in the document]
RaphaelOstier: VFS isn't used anymore to check hydration so it isn't needed at all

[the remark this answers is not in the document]
RaphaelOstier: https://github.com/Infomaniak/desktop-kDrive/pull/2010/commits/f896ddc46078869c24a4d9b4ea8147b72242b816

test/libsyncengine/jobs/network/testnetworkjobs.cpp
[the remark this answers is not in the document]
RaphaelOstier: doesn't pass through because of vfs anymore and test assert shouldUpload() method https://github.com/Infomaniak/desktop-kDrive/pull/2010/commits/f896ddc46078869c24a4d9b4ea8147b72242b816

src/libsyncengine/jobs/network/kDrive_API/upload/uploadjob.cpp
RaphaelOstier: it was indeed not required anymore, and since vfs doesn't do anything anyway weather should upload is True or False : https://github.com/Infomaniak/desktop-kDrive/pull/2010/commits/c49e0b8b2b035a0d09577f0f54c276d1850927b3

src/libsyncengine/jobs/network/kDrive_API/upload/uploadjob.cpp
```diff
@@ -98,13 +98,68 @@ ExitInfo UploadJob::canRun() {
...
+            return exitInfo;
+        }
+    }
+    return ExitCode::Ok;
+}
+
+ExitInfo UploadJob::runJob() noexcept {
+    if (!_fileId.empty()) {
+        const ExitInfo exitInfo = resolveUploadNeed();
+        if (!_shouldUpload && exitInfo) return ExitCode::Ok;
```
ClementKunz, line 129: So here you can have `_shouldUpload` that is `false` but the upload occurs because of `exitInfo` is not `Ok`. Therefor, the following test might pass while the uplaod occurs:
```
CPPUNIT_ASSERT_MESSAGE("Hash match: upload must have been skipped", !editJob.shouldUpload());
```

I still think the `shouldUpload` should be a local variable.
Then, in the tests, perhaps you could check if the upload occurs based an existing vairable such as `_progressSize` that is used to track the progress of uploads and downloads. I am not 100% sure that it works, but that might a lead to make your unit tests more robusts.

ClementKunz: Or at least, `_shouldUpload` must be set to `true` if `resolveUploadNeed` return an exit code different from `Ok`

src/libsyncengine/jobs/network/kDrive_API/checkhashmatchjob.h
```diff
@@ -37,6 +37,7 @@ class CheckHashMatchJob : public AbstractTokenNetworkJob {
         std::string getSpecificUrl() override;
         ExitInfo getFileSize(const SyncPath &path, int64_t &size);
         ExitInfo runJob() noexcept override;
+        ExitInfo remoteHashMatch() noexcept;
```
ClementKunz: This method does not have a definition anymore

src/libsyncengine/jobs/network/kDrive_API/upload/uploadjob.cpp
RaphaelOstier:
```cpp
    if (!_shouldUpload) {
        LOGW_DEBUG(_logger, L"Changing last modified date without uploading: hash match");
        if (const ExitInfo exitInfo = applyFileDates(); !exitInfo) {
            LOGW_DEBUG(_logger, L"applyFileDates failed: " << exitInfo << L" Proceeding UploadJob normally.");
            _shouldUpload = true; //set it to true to proceed with download
            return exitInfo;
        }
    }
```
wouldn't that work ? how can I know in the tests if I uploaded, if I have no variable that tells me if I uploaded ?

src/libsyncengine/jobs/network/kDrive_API/checkhashmatchjob.h
RaphaelOstier: oops https://github.com/Infomaniak/desktop-kDrive/pull/2010/commits/7999f12d6b0d508294a1f5254bce25a008b9b369

src/libsyncengine/jobs/network/kDrive_API/upload/uploadjob.cpp
RaphaelOstier: https://github.com/Infomaniak/desktop-kDrive/pull/2010/commits/f2d3077752dfa0c1b06871ba7d5f2c812eb87e52

ClementKunz (approved):

PR #2021, opened 2026-06-12
fix(appserver): capture variable by copy in lambda

herve-er:
Updated the lambda in `QTimer::singleShot` to capture `startedSyncDbIds` by copy instead of by reference. This change prevents potential issues/crash caused by destruction of the original variable before the lambda execution, improving code safety and predictability.


ChristopheLarchier (approved):

herve-er:
Forced merge → the extended failure was only due to a failure when removing the debug certificate from the CI (it had already been removed by another job).

PR #2029, opened 2026-06-15
feat(search-job): Introduces the `isHydrated` flag in every `SearchInfo` item returned by `SearchJob`

luc-guyot-infomaniak:
These changes introduce the new boolean attribute `isHydrated` to the `SearchInfo` object and let the method `SearchJob::handleResponse` computes the value of this flag for every searched item.


src/libsyncengine/jobs/network/kDrive_API/searchjob.cpp
```diff
@@ -90,10 +90,47 @@ void SearchJob::setQueryParameters(Poco::URI &uri) {
     uri.addQueryParameter("with", "path");
 }
 
-ExitInfo SearchJob::handleResponse(std::istream &is) {
-    if (const auto exitInfo = AbstractTokenNetworkJob::handleResponse(is); !exitInfo) {
-        return exitInfo;
+
+SearchJob::LocalProperties SearchJob::getLocalProperties(const SyncPath &itemPath) const {
```
ChristopheLarchier: I would prefer that the function returns an ExitInfo and that SearchJob::LocalProperties is returned as a parameter.

src/libsyncengine/jobs/network/kDrive_API/searchjob.cpp
```diff
@@ -90,10 +90,47 @@ void SearchJob::setQueryParameters(Poco::URI &uri) {
...
+                std::char_traits<std::remove_cvref_t<decltype(*sharedFolder)>>::length(sharedFolder));
+    }
+
+    if (localProperties.path.native().starts_with(Str("/")) || localProperties.path.native().starts_with(Str("\\"))) {
+        localProperties.path = localProperties.path.relative_path();
+    }
+
+    const SyncPath absolutePath = _syncRootPath / localProperties.path;
+    IoError ioError = IoError::Success;
+    if (!IoHelper::checkIfPathExists(absolutePath, localProperties.isAvailableLocally, ioError,
```
ChristopheLarchier: You don't need to call IoHelper::checkIfPathExists as IoHelper::checkIfFileIsDehydrated will return an ioError if the item doesn't exist (and there is a probability that it exists during the first call but not during the second).

luc-guyot-infomaniak: This is not true for the Linux implementation. Do you advise to resort to `#if defined`?

src/libsyncengine/jobs/network/kDrive_API/searchjob.cpp
luc-guyot-infomaniak: In the initial implementation of `SearchJob`, no `ExitInfo` object was sent from this part of the code. Unexpected errors were just logged, presumably because inaccurate information will not harm the GUI components that consume the search output. 

Is this something you want to change?

src/libsyncengine/jobs/network/kDrive_API/searchjob.cpp
ChristopheLarchier: OK, so, depending on the VFS mode (store sync.virtualFileMode() in the constructor), you could call either IoHelper::checkIfPathExists or IoHelper::checkIfFileIsDehydrated.

src/libsyncengine/jobs/network/kDrive_API/searchjob.cpp
ChristopheLarchier: This function did not exist in the previous code; I think it is better for it to return an ExitInfo and for the caller to decide what to do with it (just log it in this case).

src/libsyncengine/jobs/network/kDrive_API/searchjob.cpp
luc-guyot-infomaniak: Done in 0abc8f4ec6d3f614dcb05a5419c616ccb754434a.

src/libsyncengine/jobs/network/kDrive_API/searchjob.cpp
luc-guyot-infomaniak: Done in 0abc8f4ec6d3f614dcb05a5419c616ccb754434a.

ChristopheLarchier (approved):

PR #2045, opened 2026-06-18
fix(i18n): remove vanished tags, restore translations and add missing strings

ClementKunz:
## Changes

- Fix all translation files where every entry was marked with `type="vanished"` 
- Added missing translations in all languages for new error string in parametersdialog.cpp
- Fixed `update_ui_translations.py` script to include Dutch (nl) language support

## Verification

- [x] No vanished tags remain across all 12 .ts files
- [x] All languages have proper translations
- [x] 0 unfinished entries remain

### DIFF
--- a/translations/client_da.ts
+++ b/translations/client_da.ts
@@ -1,2580 +1,3095 @@
 <?xml version="1.0" encoding="utf-8"?>
 <!DOCTYPE TS>
-<TS version="2.1" language="da_DK">
+<TS language="da_DK" version="2.1">
 <context>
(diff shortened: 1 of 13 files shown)


herve-er (dismissed):
The indentation is gone in this PR. Not a big deal, as those files are not really meant to be human-readable anyway, and the tools are indentation-insensitive.

I'll let you decide whether this was just an easy-to-fix unwanted change or whether it's caused by the `update_ui_translations` script. In the latter case, it can probably stay as-is.

herve-er (approved):

PR #2052, opened 2026-06-19
feat(macOS): Block actions when updating synchro mode

valentinperignon:


adrien-coye (approved):
So I did test this branch.
We identified a potential issue where _with a debugger attached_ we can overwhelm the "signal" queue, thus making the "conversion UI state" take longer to change.
UI will now switch earlier to a locked state. This is all good to go in the PR to go for the QA build.

PR #2053, opened 2026-06-19
feat(macOS): Add confirmation alert when changing synchro mode

valentinperignon:
CleanShot 2026-06-19 at 08 58 56@2x


adrien-coye (approved):
Reviewed the view and tested, let's merge this in the base PR 👍

PR #2060, opened 2026-06-19
refactor(types): improve toString implementation

herve-er:
Rewrote `toString(const std::source_location &)` for better readability, portability, and performance:
- Added platform-specific path separator handling.
- Extracted file name and simplified function name formatting.
- Used `std::to_chars` for efficient line number conversion.
- Constructed formatted string as `file:line[func]`.

test(testtypes): added assertion for `ExitInfo` string conversion to validate the new `toString` implementation.

### DIFF
--- a/src/libcommon/utility/types.cpp
+++ b/src/libcommon/utility/types.cpp
@@ -970,12 +970,11 @@ std::string toString(const std::source_location &loc) {
         func.remove_prefix(pos + 1);
     }
 
-    std::string lineBuffer;
-    lineBuffer.reserve(16); // Enough to hold any line number
-    const auto [ptr, ec] = std::to_chars(lineBuffer.data(), lineBuffer.data() + lineBuffer.size(), loc.line());
+    char lineBuffer[16] = {};
+    const auto [ptr, ec] = std::to_chars(lineBuffer, lineBuffer + sizeof(lineBuffer), loc.line());
 
     // ptr points to the first character after the last written character
-    const auto lineLength = static_cast<size_t>(ptr - lineBuffer.data());
+    const auto lineLength = static_cast<size_t>(ptr - lineBuffer);
 
     std::string result;
     result.reserve(file.size() + 1 + // :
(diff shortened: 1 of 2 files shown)


src/libcommon/utility/types.cpp
```diff
@@ -950,8 +947,49 @@ std::string toString(const TranslationMode e) {
...
+
+    // Remove the return type / calling convention prefix (if any) without breaking
+    // names that contain spaces (e.g. "operator bool").
+    if (const auto nsPos = func.find("::"); nsPos != std::string_view::npos) {
+        if (const auto pos = func.rfind(' ', nsPos); pos != std::string_view::npos) {
+            func.remove_prefix(pos + 1);
+        }
+    } else if (const auto pos = func.find_last_of(' '); pos != std::string_view::npos) {
+        func.remove_prefix(pos + 1);
+    }
```
R-Gld: This `func.find("::")` break the logic if the method return a namespaced type, like `std::string`.

Example, in `DownloadJob::canRun`:
https://github.com/Infomaniak/desktop-kDrive/blob/fbdeca79dc770ef2b72e9691b239bc43cd215acd/src/libsyncengine/jobs/network/kDrive_API/downloadjob.cpp#L116-L142

Here `func` is `KDC::ExitInfo KDC::DownloadJob::canRun(...)`. The first `::` found is the one inside the return type `KDC::ExitInfo`, and there is no space before it, so the return type prefix is never stripped.

So instead of:
```
ExitInfo{DataError-FileExists from (downloadjob.cpp:139[KDC::DownloadJob::canRun])}
```
we get:
```
ExitInfo{DataError-FileExists from (downloadjob.cpp:139[KDC::ExitInfo KDC::DownloadJob::canRun])}
```

herve-er: Done here: https://github.com/Infomaniak/desktop-kDrive/pull/2060/commits/165a3666044863d2f6b6ec7c379ab6ad19f80885

R-Gld (approved):

PR #2066, opened 2026-06-23
refactor(account): unify Account and AccountInfo models

ClementKunz:
Depends on #2075

Unifies the legacy AccountInfo DTO and the DB Account model into a single Account class in libcommon/data/, following the same pattern established for Drive/DriveInfo and DriveAvailable/DriveAvailableInfo.

Key changes:
- Creates unified Account in libcommon/data/account.h/.cpp
- Moves fields, serialization, and QDataStream operators from both AccountInfo and db::Account
- Switches all server, sync engine, GUI legacy, and test consumers to Account
- Removes accountToAccountInfo mapping function (now redundant)
- Updates ParmsDb to use libcommon/data/account.h
- Preserves JSON IPC keys (accountInfo, accountInfoList) for compatibility with gui4/windows/


src/gui/clientgui.h
```diff
@@ -25,11 +25,11 @@
 #include "custommessagebox.h"
 #include "logindialog.h"
 #include "info/userinfoclient.h"
-#include "info/accountinfo.h"
+#include "libcommon/data/account.h"
 #include "info/driveinfoclient.h"
 #include "info/syncinfoclient.h"
 #include "info/syncfileiteminfo.h"
-#include "libcommon/info/accountinfo.h"
+#include "libcommon/data/account.h"
```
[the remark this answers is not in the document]
ClementKunz: Will be done in next PR

src/server/requests/serverrequests.cpp
```diff
@@ -684,7 +675,7 @@ ExitInfo ServerRequests::addSync(const UserDbId userDbId, const AccountId accoun
         account.setDbId(accountDbId);
         account.setAccountId(accountId);
         account.setUserDbId(userDbId);
-        if (const auto exitCode = createAccount(account, accountInfo); exitCode != ExitCode::Ok) {
+        if (const auto exitCode = createAccount(account); exitCode != ExitCode::Ok) {
             LOG_WARN(Log::instance()->getLogger(), "Error in createAccount");
             return exitCode;
         }
```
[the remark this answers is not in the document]
ClementKunz: Will be done in next PR

src/server/appserver.cpp
```diff
@@ -1558,8 +1558,8 @@ void AppServer::onRequestReceived(int id, RequestNum num, const QByteArray &para
                     break;
                 }
 
-                if (accountInfo.dbId() != 0) {
-                    sendAccountAdded(accountInfo);
+                if (account.dbId() != 0) {
+                    sendAccountAdded(account);
```
R-Gld: `account.dbId()` is now always set after `addSync`, even when the account already exists. So 
`ACCOUNT_ADDED` is now sent again for accounts the client already has, not just for new ones.

Before, the guard checked a separate `AccountInfo` that was filled only when a new account was created. This is different from the missing-name issue below and won't be fixed by it,  the check should tell whether the account was really created.

src/server/comm/guijobs/syncaddjob.cpp
```diff
@@ -71,8 +71,8 @@ ExitInfo SyncAddJob::process() {
         return exitInfo;
     }
 
-    if (accountInfo.dbId() != 0) {
-        auto signalAccountAddedJob = std::make_shared<SignalAccountAddedJob>(accountInfo);
+    if (account.dbId() != 0) {
+        auto signalAccountAddedJob = std::make_shared<SignalAccountAddedJob>(account);
```
R-Gld: Sam here: `account.dbId() != 0` is now `true` for every successful `addSync`, even for an existing account so a `SignalAccountAddedJob` is sent for accounts that already exist.

src/server/appserver.cpp
R-Gld: Fixed in the next PR #2075 maybe we should set this PR as depending 2075 so `develop` does not contain the issues ?

src/server/comm/guijobs/syncaddjob.cpp
R-Gld: See https://github.com/Infomaniak/desktop-kDrive/pull/2066#discussion_r3497111154

src/server/appserver.cpp
ClementKunz: Done in PR https://github.com/Infomaniak/desktop-kDrive/pull/2075

src/server/comm/guijobs/syncaddjob.cpp
ClementKunz: Done in PR https://github.com/Infomaniak/desktop-kDrive/pull/2075

src/libcommon/data/user.cpp
```diff
@@ -0,0 +1,101 @@
...
+    _email(email),
+    _avatarUrl(avatarUrl),
+    _avatar(avatar),
+    _toMigrate(toMigrate) {}
+
+void User::toDynamicStruct(Poco::DynamicStruct &dstruct) const {
+    CommonUtility::writeValueToStruct(dstruct, userDbIdKey, _dbId);
+    CommonUtility::writeValueToStruct(dstruct, userUserIdKey, _userId);
+    CommonUtility::writeValueToStruct(dstruct, userNameKey, CommonUtility::str2CommString(_name));
+    CommonUtility::writeValueToStruct(dstruct, userFirstNameKey, CommonUtility::str2CommString(_firstName));
```
R-Gld, line 58: Maybe we could use the fallback to the `_name` used here in the method.

See the comment in `user.h`:

> User logged in a version of kDrive Desktop < 4.0 might not have the firstName field populated until they have network connectivity. In this case, we can use the name field as a fallback to avoid showing an empty name in the UI.

```suggestion
    CommonUtility::writeValueToStruct(dstruct, userFirstNameKey, CommonUtility::str2CommString(firstName()));
```

ClementKunz: Don't you think that this might lead to wrong UI display? For example, for a user named `John Doe`, a label showing the username might show `Doe Doe`, or `John Doe John Doe` if the `name` value contains the full name. IMO it is better to show only `Doe` in that case.
What do you think?

R-Gld: Out of scope for this PR, discussed IRL.

R-Gld (approved):
LGTM

src/libcommon/data/user.cpp
ClementKunz: Screenshot 2026-07-10 at 11 29 34

`name` field is the full name, last name only is never retrieved. So the proposed fallback is safe. Will be done in a dedicated PR

PR #2073, opened 2026-06-24
fix(gui): ensure proper shutdown delay for GUI job cleanup

R-Gld:


herve-er (commented):
Ideally, I would add an `AppServer::quitLater` function that simply calls:

`QTimer::singleShot(quitDelayMs, QCoreApplication::instance(), [] { AppServer::quit(); });`

This would avoid bringing Qt into the communication layer, as we are trying to remove it.

This function could then be used in the few places where we currently use `QTimer::singleShot(..., AppServer::quit());`.

R-Gld:
> Ideally, I would add an `AppServer::quitLater` function that simply calls:
> 
> `QTimer::singleShot(quitDelayMs, QCoreApplication::instance(), [] { AppServer::quit(); });`
> 
> This would avoid bringing Qt into the communication layer, as we are trying to remove it.
> 
> This function could then be used in the few places where we currently use `QTimer::singleShot(..., AppServer::quit());`.

Yep, good idea: https://github.com/Infomaniak/desktop-kDrive/pull/2073/commits/b55ec649ebea3a701346e67403818356d5173bf0

herve-er (approved):

PR #2079, opened 2026-06-25
chore: Bump Periphery and fix issues

valentinperignon:


adrien-coye (approved):
LGTM

adrien-coye:
I'd rather get feedback from @PhilippeWeidmann if possible on this one instead of merging it as-is.

adrien-coye:
Discussed with @PhilippeWeidmann, this tech is not production ready for macOS, thus we remove it for now until it works reliably. cc @valentinperignon

PR #2082, opened 2026-06-25
fix: SplitView should not be collapsable

valentinperignon:


adrien-coye (approved):
Code OK, I checked it also.

PR #2093, opened 2026-06-29
Feat: flatpak build

### LINKED ISSUE #125: kDrive as a flatpak
*Note: Please write your issue only in english*

**Is your feature request related to a problem? Please describe.**
The golden standard of packaged application in linux is Flatpak. All new distros use flatpak. Flatpak is also integrated into distro "app stores". This would enable update to be done automatically vs the current manual updates needed with the appimages (beneficial also for devs they would see way quicker updates ;))

Having kdrive available on flathub is also a source of visibility and will potentially lead to new customers that are FOSS fans.

**Describe the solution you'd like to implement**
Kdrive should be available as a flatpak, optimally on flathub

**Describe alternatives you've considered**
I'm running the appimage now, because i don't have a choice :)

**Additional context**
fully available to test if needed

Persson-dev:
This PR introduces reproducible flatpak build and fixes #125.

I took https://github.com/eylenburg/kdrive-flatpak repo as a base and changed it to compile kDrive for the flatpak instead of using the AppImage. 
Some tweaks in the code needed to be done and I'm aware they are not very clean, especially for the system tray.

Since the flatpak does not have access to filesystem, the .desktop file is patched when building to contain the `x-scheme-handler/kdrive` field. And the config files are now written to the flatpak `$XDG_CONFIG_HOME` persistent directory.

By default, the flatpak is granted access to `~/kDrive` for the app to work properly. Syncing to other directories will in fact, work, but kDrive won't be able to monitor file changes. So syncing in other directories must be done manually by pausing and resuming syncing, which is not convenient. However, the kDrive flatpak can be authorized to fully access a directory with the `--filesystem` flag.

Furthermore, I also updated the keychain submodule to match tag `v1.3.1` because there was a warning treated as an error when building.

Lastly, I think the script can be improved to use conan instead of manually cloning git repos but I had a lot of trouble making conan to work with flatpak, so I decided to not use it for now.


FabianDevel:
Hello, thank you kindly for your contribution. 
However, due to the sandboxing philosophy of flatpak, several problems are emerging. 
To list some : a lot of tweaks would be needed for the system and filesystem permissions (as you said for the systray for example, but also for the app autostart). These permissions could potentially be refused or ignored silently.
We should probably rewrite the deps system because of incompatibility with Conan.
 
These problems are why we don't plan to release a flatpak build, and thus I cannot accept your PR.
What we plan to do however is to support .deb in the future (even if it won't be compatible with all distribs, it will greatly improve the existing, by allowing to have an in-app auto-update for example). 

Thank you again and sorry.

lexo-mfleuti:
@FabianDevel `.deb` is increasingly problematic due to all sorts of compatibility issues. I think the filesharing via File Explorer Extensions and Sync via D-Bus could be the simpler approach. The install routine would need to setup the required permissions. Perhaps 2 software packages would need to be installed as it was with OwnCloud: A sync client and a file explorer plugin. In my opinion this should be the way to go nowadays. It's properly isolated and easier to setup for endusers since they can simply install it via package manager in any distro - including Arch/NixOS etc.

I am not a pro in Flatpak development but I see the everyday issues with .deb installs and maintenance. It's not a good solution in my humble opinion.

beankylla:
> Hello, thank you kindly for your contribution. However, due to the sandbowing philosophy of flatpak, several problems are emerging. To list some : a lot of tweak would be needed for the system and filesystem permissions (as you said for the systray for example, but alos for the app autostart). These permissions could potentially be refused or ignored silently. We should probably rewrite the deps system because of incompatibility with Conan.
> 
> These problems are why we don't plan to release a flatpak build, and thus I cannot accept your PR. What we plan to do however is to support .deb in the future (even if it won't be compatible with all distribs, it will greatly improve the existing, by allowing to have an in-app auto-update for example).
> 
> Thank you again and sorry.

Seems like the bad decision is made for the bad reason.

Releasing a packaged file (.deb) now seems like going against the wind and current. 

Trend is distros going immutable so no more .deb installed just flatpak or other packaged apps.

Also i'm wondering what happened since your Nextcloud fork that you now cannot go flatpak even though NC is working like a charm.

Last point: Did not see any .deb request, but did see numerous people supporting the flatpak release, but maybe not everything is public and there is great demand for the .deb release.

This is a bit like going full blackberry in 2009.

Persson-dev:
Hello @FabianDevel, thank you for your response.

I initially started working on this PR because of #125.
As you pointed out, everything is not perfect but I think the PR can be improved.

- First of all, the filesystem can actually be fully exposed to the application (this is what the Nextcloud client does) and therefore would work properly. I just tried to apply the principle of least privilege. But if it's not convenient for the user, I guess it can be not followed.
- As for the system tray, I did not fully understand why the `ClientGui::osRequireMenuTray()` was needed so I just made it returned true when flatpak was used ...
- Lastly, the integration with conan package manager is just a skill issue on my side and I did not try to make the autostart work, but I think these are doable.

Is it worth for me to investigate those issues or is the flatpak version completely abandoned ?

FabianDevel:
Hello @Persson-dev, 

Thanks again for your involvment.

This is a project that would need a deep dive and understanding of the existing code, the deps system, and would require us to spend some significant amount of time just to figure out every problem the flatpak's sandboxing could cause to the app, and how to fix them.

Unfortunatly we simply don't have the resource to do that for now. And we also don't have the resource to provide support to flatpak builds if there are any problems like there are for NextCloud ([here for example](https://github.com/flathub/com.nextcloud.desktopclient.nextcloud/issues)). We're aware that there are lots of improvements to do on the linux version though.

That's the reason why we officially only support Ubuntu on linux. Maybe we'll look into it in the future, but as of today (and near future), providing flatpak's build is not something we want to do.

emanuc:
> Hello @Persson-dev,
> 
> Thanks again for your involvment.
> 
> This is a project that would need a deep dive and understanding of the existing code, the deps system, and would require us to spend some significant amount of time just to figure out every problem the flatpak's sandboxing could cause to the app, and how to fix them.
> 
> Unfortunatly we simply don't have the resource to do that for now. And we also don't have the resource to provide support to flatpak builds if there are any problems like there are for NextCloud ([here for example](https://github.com/flathub/com.nextcloud.desktopclient.nextcloud/issues)). We're aware that there are lots of improvements to do on the linux version though.
> 
> That's the reason why we officially only support Ubuntu on linux. Maybe we'll look into it in the future, but as of today (and near future), providing flatpak's build is not something we want to do.


Thanks for the detailed explanation. I understand that resources are limited and that maintaining another distribution format comes with additional work.
However, currently the AppImage version has several drawbacks that a Flatpak build would solve:

- seamless automatic updates without having to manually download a new AppImage from your website every time;
- no need to repeatedly set the executable permission after every update (I literally had to do this again after updating today);
- better desktop integration;
- improved security through sandboxing;
- better compatibility thanks to the shared runtime and a more predictable environment across distributions;
- ability to bundle their own libraries without depending on the host system.

I understand that Flatpak's sandbox introduces challenges, but I think those issues are worth addressing considering the long-term benefits for both users and developers.

As an example, NVIDIA already distributes its own app (GeForce NOW) through its own official Flatpak repository, rather than through the Flathub store, which gives them more controlled management of permissions. This could be a good model to follow, showing that maintaining a first-party Flatpak repo is a viable approach even for companies with strict requirements on how their software is packaged and permissioned.

I subscribed to your KSuite service trusting that the Linux application would continue to improve, [and for me that also includes an official Flatpak version](https://github.com/Infomaniak/desktop-kDrive/issues/125#issue-2303601624). Linux users are paying customers too, and I don't think we should feel like second-class customers compared to users on other operating systems.

PS: I use Fedora, and many of your Linux customers don't use Ubuntu either. Officially supporting only Ubuntu doesn't really reflect the diversity of the Linux ecosystem today.

PR #2104, opened 2026-07-02
fix(macOS): Update log path and display error if the path doesn't exist

valentinperignon:


adrien-coye (approved):
LGTM

adrien-coye (approved):
I did test it, it works fine when I delete the cache folder. LGTM 👍

adrien-coye (approved):
Still works

PR #2105, opened 2026-07-02
feat: Reset PreferencesRouter when the preferences window opens

valentinperignon:


adrien-coye (commented):
Code looks good but does not reset _other_ tabs. Checking if that is what is expected or not.

adrien-coye (approved):
I did check the issue I had initially with state not reset in other tabs. All good on my side 🤝

PR #2106, opened 2026-07-02
feat(server): open a local file from a kdrive://open URL

Clandestino775:
## Description

This PR adds support for opening a synchronized file directly from a web browser link, similar to what SharePoint/OneDrive offer, using the URL format:

```
kdrive://open/<file path relative to the sync folder>[?driveId=<server drive id>]
```

Example: `kdrive://open/Documents/Report.docx?driveId=123456`

### Behavior

- The file path is resolved against the local folders of all configured syncs. The optional `driveId` query item (the server-side drive id, as shown in web application URLs) restricts the search to the syncs of one drive.
- **File available locally** → it is opened with its default application.
- **Dehydrated LiteSync placeholder** → its hydration is triggered first, following the exact same path as the "Make available locally" context menu action (pin state on macOS + `SyncPal::addDlDirectJob`), and the file opens once the download completes. The user is notified that the download is in progress.
- **File not synchronized locally yet** → if it is known on the remote side (`SyncPal::checkIfExistsOnServer`), the handler waits for the synchronization to bring it locally (up to 10 minutes), then opens it. On a cold start, resolution is retried for 30 seconds while the syncs come up.
- **Safety**: absolute paths and paths containing `..` components are rejected; executable content (`.exe`, `.bat`, `.sh`, `.jar`, …) is revealed in its parent folder instead of being launched.

### How the URL reaches the application

The `kdrive` scheme is already registered on the three platforms by `Utility::registerLoginRedirection()` (used for the `kdrive://auth-desktop` OAuth redirect), so no new OS registration is needed:

- **Windows / Linux**: the URL arrives as a command-line argument, parsed in `AppServer::parseOptions`. When an instance is already running, it is forwarded through the existing `QtSingleApplication` message channel (new `openFile` message, mirroring the `redirectLogin` one).
- **macOS**: the URL arrives as a `QEvent::FileOpen`, caught by the existing event filter (renamed `AuthorizationCodeEventFilter` → `UrlSchemeEventFilter` since it now handles two URL kinds).

### Implementation

- New `OpenFileUrlHandler` class (`src/server/openfileurlhandler.{h,cpp}`): a small `QTimer`-driven state machine (Resolve → WaitForSync → WaitForHydration) running on the server main thread; all per-tick checks are cheap (DB/snapshot/VFS status lookups).
- URL parsing, path safety validation and the executable-extension check are static and covered by unit tests (`test/server/openfileurlhandler/`).
- `src/server/AGENTS.md` updated with the kdrive:// URL scheme documentation.

### Notes for reviewers

- The `open` host segment is required because the host part of a URL is case-insensitive: putting the first path component in the host position would corrupt case-sensitive folder names.
- On hydration failure detection, two consecutive negative `isDownloadOngoing` checks are required before failing, to avoid a race with the download job completion callback.
- User-facing strings are new `tr()` entries; translation files are expected to be regenerated by the usual lupdate flow.

### DIFF
--- a/src/server/AGENTS.md
+++ b/src/server/AGENTS.md
@@ -11,6 +11,12 @@ Long-running background daemon that owns all sync state. Communicates with the G
 - Auto-updater (Windows): `src/server/updater/windowsupdater.h`
 - Auto-updater (Linux): `src/server/updater/linuxupdater.h`
 - Migration from legacy config: `src/server/migration/`
+- `kdrive://open/...` URL handler (open a synced file from a browser link, hydrating/downloading it first if needed): `src/server/openfileurlhandler.{h,cpp}`
+
+## kdrive:// URL scheme
+The `kdrive` scheme is registered at server startup by `Utility::registerLoginRedirection()` (registry keys on Windows, `lsregister` on macOS, `.desktop` + `xdg-mime` on Linux). Incoming URLs reach the server as a command-line argument (Windows/Linux, forwarded to the running instance through the `QtSingleApplication` message channel) or as a `QEvent::FileOpen` caught by `UrlSchemeEventFilter` (macOS). Two hosts are handled:
+- `kdrive://auth-desktop?...` — OAuth login redirect (see `AppServer::onAuthorizationCodeReceived`).
+- `kdrive://open/<relative path>[?driveId=<server drive id>]` — open a file of a sync (see `OpenFileUrlHandler`).
 
 ## IPC Architecture
 The GUI sends typed **job requests** to the server over the IPC channel. Each capability is a separate job class.
--- a/src/server/CMakeLists.txt
+++ b/src/server/CMakeLists.txt
@@ -42,6 +42,7 @@ endif()
 
 set(server_SRCS
     appserver.h appserver.cpp
+    openfileurlhandler.h openfileurlhandler.cpp
     comm/oldcommserver.h comm/oldcommserver.cpp
     comm/abstractcommchannel.h comm/abstractcommchannel.cpp
     comm/abstractcommserver.h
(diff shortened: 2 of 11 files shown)


FabianDevel:
Thank you kindly for your submission. 

However, we don't plan to offer this feature anytime, as this format is not used for our file's deeplinks, so we don't see the added value as only you could  make your "private deeplink" if you are aware of this url scheme. 
It would also means support from us to this code, which is not planned.

Thus I close this PR.

Clandestino775:
Cher Monsieur Devel,
Nous sommes étonnés que vous n'ayez pas considéré cette PR. En effet vous (Infomaniak) voulez vous mettre en avant comme alternative souveraine face à Microsoft (ce que nous, entreprise Suisse, encourageons vivement). Vous avez mis en place tout un système d'API qui permet une intégration externe de vous outils. Cette PR est le chainon manquant entre vos outils et les intégrations externes. 
Le format de l'URL est simple et standarisé et ne nécessite pas où peu de maintenance. 
Beaucoup de nos clients décident de finalement opter pour MS365 du fait que cette fonctionnalité d'ouverture de documents en local via kDrive n'est pas gérée.
Je vous prie de reconsidérer votre point de vu, cette fonctionnalité est essentiel afin d'offrir un écosystème crédible face à MS365.
Cordialement

FabianDevel:
Bonjour monsieur, nous prenons bonne note de vos explications sur ce que vous attendez vraiment de cette demande de feature, notamment sur l'impact sur vos clients.
Nous allons donc réévaluer en interne cette demande avec l'équipe produit et vous tiendrons au courant du suivi. 
Merci encore pour votre intérêt. 
Cordialement

PR #2116, opened 2026-07-08
fix(server): stop OldCommServer worker thread at shutdown

R-Gld:
This pull request improves the shutdown procedure of the application by ensuring that background worker threads, particularly those associated with the `OldCommServer`, are properly stopped and cleaned up. This prevents intermittent crashes and resource leaks during application exit. Additionally, it ensures that the update manager is destroyed at the correct time to avoid shutdown-related crashes.

Shutdown and thread management improvements:

* Added a `stop()` method to `OldCommServer` that cleanly stops and joins the request worker thread, preventing the thread from outliving the application and causing race conditions or crashes at shutdown. (`src/server/comm/oldcommserver.cpp`, `src/server/comm/oldcommserver.h`) [[1]](diffhunk://#diff-88448ff60ec84d9a03ef3738e97e0e4f86cc33d84e7f2b647d646f858169deaeR91-R101) [[2]](diffhunk://#diff-6f1ba3cfe7cc7088b8f131e4bddebab04cf9fe60918e918ae295ac260ed82d24R56-R58)
* Updated `AppServer::cleanup()` to call `OldCommServer::stop()` if the old communication server is in use, ensuring its worker thread is stopped before shutdown. (`src/server/appserver.cpp`)

Resource destruction timing:

* Modified `AppServer::cleanup()` to explicitly destroy the update manager (`reset()`) after network jobs are cleared but before the Qt event loop ends, preventing crashes due to destruction order at shutdown. (`src/server/appserver.cpp`)

### DIFF
--- a/src/server/comm/oldcommserver.cpp
+++ b/src/server/comm/oldcommserver.cpp
@@ -53,12 +53,7 @@ std::shared_ptr<OldCommServer> OldCommServer::instance(QObject *parent) {
 
 OldCommServer::OldCommServer(QObject *parent) :
     QObject(parent),
-    _requestWorkerThread(new QtLoggingThread()),
-    _requestWorker(new Worker()),
-    _tcpSocket(nullptr),
-    _buffer(QByteArray()),
-    _hasQuittedProperly(false),
-    _stopping(false) {
+    _requestWorker(new Worker()) {
     // Start worker thread
     _requestWorker->moveToThread(_requestWorkerThread);
     connect(_requestWorkerThread, &QThread::started, _requestWorker, &Worker::onStart);
--- a/src/server/comm/oldcommserver.cpp
+++ b/src/server/comm/oldcommserver.cpp
@@ -28,6 +28,7 @@
 #include <QJsonDocument>
 #include <QJsonParseError>
 #include <QJsonObject>
+#include <QThread>
 #include <QTimer>
 
 #include <fstream>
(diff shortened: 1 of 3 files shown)


src/server/comm/oldcommserver.cpp
```diff
@@ -88,6 +88,17 @@ OldCommServer::~OldCommServer() {
...
+void OldCommServer::stop() {
+    // Break out of Worker::onStart, then quit and join the worker thread. Without this the thread outlives shutdown.
+    if (_requestWorker) {
+        _requestWorker->stop();
+    }
+    if (_requestWorkerThread) {
+        _requestWorkerThread->quit();
+        _requestWorkerThread->wait();
+    }
+}
```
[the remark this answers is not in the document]
R-Gld: https://github.com/Infomaniak/desktop-kDrive/pull/2116/commits/595ec34c863e61cd5d97752d7c7eb5ffc888da69

src/server/comm/oldcommserver.h
```diff
@@ -71,6 +74,7 @@ class OldCommServer : public QObject {
         QByteArray _buffer;
 
         bool _hasQuittedProperly;
+        bool _stopping;
```
ClementKunz: We usually prefer to initialize members in header file.

R-Gld: Yep Thx ! 

https://github.com/Infomaniak/desktop-kDrive/pull/2116/commits/0b1a47f49033e30afc9d18da71ca5d10ca2071aa

ClementKunz (approved):

PR #2125, opened 2026-07-10
fix(libcommon): use firstName() in User::toDynamicStruct

ClementKunz:
Fixed User::toDynamicStruct to call firstName() instead of directly using _firstName, making the fallback behavior (using full name when first name is empty) correctly serialized.


R-Gld (approved):
LGTM

Related to https://github.com/Infomaniak/desktop-kDrive/pull/2066#discussion_r3557464092

PR #2126, opened 2026-07-10
fix(Onboarding): Improve onboarding flow

valentinperignon:
### Drive Selection
- Only display the drives of the user currently in the flow
- Fix a bug about the scroll view when many drives are displayed

### Connect new account
- When the user wants to add an account, the app open the onboarding in a separate window so the user can easily come back to the app


adrien-coye (approved):
Code LGTM, I did run an onboarding without issues.

PR #2136, opened 2026-07-13
fix: Signal ordering server

adrien-coye:
PoC for a signal ordering fix that would guarantee ACCOUNT_ADDED -> DRIVE_ADDED -> SYNC_ADDED in order


adrien-coye:
Closing this PoC, if a cache now detect something impossible it will reload on its own

PR #2164, opened 2026-07-16
fix(server): ensure quitLater timer is created on the main thread

ClementKunz:
Ensures QTimer::singleShot is created from the main thread event loop by chaining through QMetaObject::invokeMethod with Qt::QueuedConnection. Prevents unreliable timer transfer when quitLater is called from non-Qt threads (e.g. Poco thread pool) in release builds.


adrien-coye (approved):
I built a release build on top of this branch, and we both validated with @ClementKunz It was working as expected.

Note: XPC call from the GUI in on a dedicated thread, this, I think, is the root cause of the issue.

adrien-coye:
following https://github.com/Infomaniak/desktop-kDrive/pull/2161

PR #2166, opened 2026-07-17
feat: Show icon in the dock when window is opened

valentinperignon:


adrien-coye (approved):
I just did test it. It does add up a nice polish to the app. I like it this way. nice 👍

PR #2168, opened 2026-07-17
feat: Show organization name

valentinperignon:


adrien-coye (approved):
Okay DriveContext instead of Drive to move the data around then a small change of the view. I did test it, all good to me.

PR #2175, opened 2026-07-20
fix: Update permissions view for macOS 15+

valentinperignon:


adrien-coye (approved):
So it only changes textual operations to follow. It looks OK. I merge it now and will be testing it on "intel Sequoia" with the other onboarding changes to validate the global behaviour.

On modern macOS I saw no change during the onboarding

PR #2178, opened 2026-07-21
feat: Update OAuth

valentinperignon:
Depends on #2185


adrien-coye (commented):
I'd like to check with you how it now behaves on my setup with FF as main browser.

PR #2202, opened 2026-07-24
fix(server): add TLS 1.2+ encryption to Windows/Linux IPC socket

LawrenceLeitgib:
## Summary
This PR adds TLS 1.2+ encryption to the Windows/Linux localhost IPC channel by migrating `SocketCommServer` to `Poco::Net::SecureServerSocket`. It introduces runtime generation of an ephemeral self-signed RSA certificate and centralizes TLS context management via `SecureContextSingleton`.

## Changes
- Migrate `SocketCommServer` and socket tests to Poco NetSSL
- Add `SelfSignedCert` utility for runtime RSA self-signed certificate generation
- Introduce `SecureContextSingleton` to manage the shared TLS context
- Update CMake dependencies and IPC documentation in `src/server/AGENTS.md`
- Update socket communication tests to establish TLS connections via `SecureStreamSocket`

## Testing
Socket communication tests have been updated to use `SecureStreamSocket` and validate TLS behavior.

## Notes
- Certificates are stored in the system keychain via `KeyChainManager`.
- This change affects the Windows and Linux IPC layers only; macOS continues to use XPC.


src/server/comm/socketcommserver.cpp
```diff
@@ -300,7 +324,20 @@ void SocketCommServer::execute() {
...
+        Poco::Net::SecureStreamSocket secureSocket(socket);
+        try {
+            if (secureSocket.completeHandshake() != 1) {
+                LOG_WARN(Log::instance()->getLogger(), "TLS handshake incomplete after accept");
+                // Let the channel handle the error through normal polling.
+            }
+        } catch (Poco::Exception &ex) {
+            LOG_WARN(Log::instance()->getLogger(), "TLS handshake failed after accept: " << ex.displayText());
+            // Let the channel handle the error through normal polling.
+        }
```
R-Gld, line 344: `completeHandshake()` runs on the accept thread, on a blocking socket with no timeout. A local process that connects and never sends a ClientHello blocks here forever: the accept loop never returns and no GUI can connect again, even after restarting it.

Worse, the thread is parked in `completeHandshake()` and not in `accept()`, so the unblocking self-connect in `close()` (line 229) no longer does anything either: `_serverSocketThread->join()` never returns and the server stops shutting down cleanly.

The `if (_stopAsked) break;` above covers the self-connect from `close()`, but nothing else.

```suggestion
        // Bound the handshake: without a timeout, a local process that opens a TCP connection and
        // never sends a ClientHello blocks this accept loop forever and no GUI can connect again.
        const Poco::Timespan sndTimeout = socket.getSendTimeout();
        socket.setReceiveTimeout(Poco::Timespan(5, 0));
        socket.setSendTimeout(Poco::Timespan(5, 0));

        // Eagerly complete the TLS handshake so that encrypted data on the wire
        // is already decrypted when the callback thread polls for it later.
        Poco::Net::SecureStreamSocket secureSocket(socket);
        bool handshakeDone = false;
        try {
            handshakeDone = secureSocket.completeHandshake() == 1;
            if (!handshakeDone) LOG_WARN(Log::instance()->getLogger(), "TLS handshake incomplete after accept");
        } catch (Poco::Exception &ex) {
            LOG_WARN(Log::instance()->getLogger(), "TLS handshake failed after accept: " << ex.displayText());
        }

        if (!handshakeDone) {
            try {
                secureSocket.close();
            } catch (Poco::Exception &ex) {
                LOG_WARN(Log::instance()->getLogger(), "Exception in SecureStreamSocket::close: " << ex.displayText());
            }
            continue;
        }

        // Keep a bounded receive timeout for the channel's lifetime: SSL_read() must never park on
        // _socketMutex forever (see the comment on readData).
        secureSocket.setReceiveTimeout(Poco::Timespan(30, 0));
        secureSocket.setSendTimeout(sndTimeout);
```

The 30s receive timeout kept past the handshake is what makes the `Poco::TimeoutException` branch suggested on `readData` effective, restoring the original (infinite) value there would leave that branch as dead code.

A cleaner variant if you'd rather not special-case the timeouts at all: move the handshake into the channel's callback thread, so connection admission and TLS negotiation stop sharing a thread.

src/server/comm/socketcommserver.cpp
R-Gld, line 83: `Poco::TimeoutException` needs to be caught before the generic `Poco::Exception`, otherwise a timeout is treated as a lost connection and a healthy channel gets torn down.

The underlying issue: `receiveBytes` can block indefinitely while holding `_socketMutex`.

`bytesAvailable()` changed contract in this PR. It used to answer "how many bytes are readable" (`available()`), it now answers "there is probably something", the `return 1` fallback fires on `poll(SELECT_READ)`, which only means *bytes are sitting on the descriptor*, not *decryptable application data*. `readData` still assumes the old contract.

So it can now be entered with nothing decodable: an incomplete TLS record, or a TLS control message. `SSL_read` then waits for the rest of the record, on a blocking socket with no timeout, **while holding `_socketMutex`**. That mutex also guards `sendBytes` (line 105), so every thread trying to push a message to the GUI, sync progress, notifications, job replies, piles up behind it.

Realistic trigger: the GUI is killed (`kill -9`, crash, debugger breakpoint) mid-way through a large message. The tail of the record never arrives, there is no usable FIN, and the channel thread stays in `SSL_read` for good.

```cpp
    } catch (const Poco::TimeoutException &) {
        // No application data decodable yet: an incomplete TLS record or a TLS control message makes
        // poll() report the descriptor readable while SSL_read() has nothing to hand back. Not a lost
        // connection, the callback thread will poll again.
        _pendingRead = false;
        return 0;
    } catch (Poco::Exception &ex) {
        LOG_ERROR(Log::instance()->getLogger(), "Exception in StreamSocket::receiveBytes: " << ex.displayText());
        _isClosing = true;
        lostConnectionCbk();
        close();
        _pendingRead = false;
        return 0;
    }
```

(`Poco::TimeoutException` derives from `Poco::Exception`, so it has to come first.)

Returning `0` is already handled correctly upstream: `fetchDataToBuffer` breaks out, `canReadMessage` returns `false`, and the callback thread goes back to polling. No other caller needs touching.

**This is only effective if the socket actually has a receive timeout**, otherwise the new branch is dead code. One line in the accept loop, after the handshake:

```cpp
        // Bound blocking reads: SSL_read() must never park on _socketMutex forever.
        secureSocket.setReceiveTimeout(Poco::Timespan(30, 0));
```

Receive side only, a send timeout would need the same branch in `writeData`, and there "the peer isn't consuming" is arguably a real failure, so that's a separate call.

src/server/comm/socketcommserver.cpp
```diff
@@ -115,7 +126,16 @@ uint64_t SocketCommChannel::writeData(const CommChar *data, uint64_t len) {
...
+                std::lock_guard lock(_socketMutex);
+                readable =
+                        _socket.poll(Poco::Timespan(0, 1000), Poco::Net::Socket::SELECT_READ | Poco::Net::Socket::SELECT_ERROR);
+            }
+            if (!readable) {
+                // This is needed to let others threads get their hands on the locks
+                // the time can be reduce or increase but not deleted
+                std::this_thread::sleep_for(std::chrono::milliseconds(5));
                 continue;
             }
```
R-Gld: Discussed IRL.

This is ~170 wakeups per second per channel, continuously, on a daemon that runs 24/7, where the previous code did one blocking 1s poll and let the kernel wake the thread. The comment is honest that it works around the locking rather than being a deliberate choice, so it seems worth fixing the cause.

I checked the Poco headers: `SecureStreamSocketImpl` overrides `available()` (-> `SSL_pending()`) but **not** `poll()`. The poll therefore only `select()` on the socket descriptor and never touches the `SSL` object, so it does not need `_socketMutex` at all.

There is a second, related issue: polling the descriptor alone misses data already sitting decrypted in the SSL buffer (`SSL_pending() > 0` while the descriptor is not readable), so `bytesAvailable()` has to be consulted first.

```suggestion
            if (bytesAvailable() == 0 &&
                !_socket.poll(Poco::Timespan(1, 0), Poco::Net::Socket::SELECT_READ | Poco::Net::Socket::SELECT_ERROR)) {
                continue;
            }
```

src/server/comm/securecontextsingleton.cpp
```diff
@@ -0,0 +1,72 @@
...
+bool applyPem(SSL_CTX *const sslCtx, const SelfSignedCert::Pem &pem) {
+    std::istringstream certStream(pem.cert);
+    std::istringstream keyStream(pem.key);
+
+    const Poco::Crypto::X509Certificate cert(certStream);
+    const Poco::Crypto::RSAKey key(nullptr, &keyStream, "");
+
+    return SSL_CTX_use_certificate(sslCtx, const_cast<X509 *>(cert.certificate())) == 1 &&
+           SSL_CTX_use_RSAPrivateKey(sslCtx, (key.impl()->getRSA())) == 1 && SSL_CTX_check_private_key(sslCtx) == 1;
+}
```
R-Gld, line 52: `applyPem` can throw before it ever returns: both `Poco::Crypto::X509Certificate(std::istream&)` and `Poco::Crypto::RSAKey(...)` throw on malformed PEM. Combined with the two `throw Poco::RuntimeException` below, every failure path out of `createContext()` is an exception, and the call site is a member initialiser in the `SocketCommServer` constructor (`socketcommserver.cpp:203`), so it propagates out of a constructor through `GuiCommServer` up to wherever the server is built.

On Linux an unavailable or locked keychain (headless session, gnome-keyring not started, kwallet locked) is routine, not exceptional. And a corrupted PEM in the keychain becomes permanently fatal, with no path to regenerate.

```suggestion
bool applyPem(SSL_CTX *const sslCtx, const SelfSignedCert::Pem &pem) {
    try {
        std::istringstream certStream(pem.cert);
        std::istringstream keyStream(pem.key);

        const Poco::Crypto::X509Certificate cert(certStream);
        const Poco::Crypto::RSAKey key(nullptr, &keyStream, "");

        return SSL_CTX_use_certificate(sslCtx, const_cast<X509 *>(cert.certificate())) == 1 &&
               SSL_CTX_use_RSAPrivateKey(sslCtx, key.impl()->getRSA()) == 1 && SSL_CTX_check_private_key(sslCtx) == 1;
    } catch (const Poco::Exception &e) {
        LOG_ERROR(Log::instance()->getLogger(), "Failed to apply TLS material: " << e.displayText());
        return false;
    }
}
```

(needs `#include "libcommonserver/log/log.h"` and `<log4cplus/loggingmacros.h>`)

Beyond the catching: what should the server actually do when TLS material cannot be obtained at all? Right now it is an exception thrown from a constructor. A logged failure with a clean shutdown, or a fallback to an in-memory ephemeral certificate, both seem preferable, but it should be a decision rather than the default.

src/libcommonserver/utility/selfsignedcert.cpp
```diff
@@ -0,0 +1,148 @@
...
+    }
+
+    if (!generate(pem)) return false;
+
+    if (!keychain->writeToken(certKey, pem.cert) || !keychain->writeToken(privKey, pem.key)) {
+        LOG_ERROR(Log::instance()->getLogger(), "Failed to store TLS material in the keychain");
+        return false;
+    }
+    return true;
+}
```
R-Gld, line 103: The certificate is meant to be regenerated on every server start, the PR description says "ephemeral". But this reuses whatever is already in the keychain and only generates on a miss, so it is persistent across restarts. The read path is dead code against the intended behaviour, and the name says the opposite of what should happen.

Two things follow from making it per-boot:

**The private key should not be stored at all.** With a fresh certificate on every start, the keychain is only a publication channel towards the GUI, which reads the certificate to pin it, and that needs the public part only. Storing the private key adds surface for nothing: a stale entry left behind would let a previous session's key be reused to impersonate the server. `keyKeychainKey` can go from the header.

**Publication order matters.** The certificate has to reach the keychain before the port does, otherwise a GUI starting in between pins the previous one. That holds today because `SecureContextSingleton::instance()` runs in the `SocketCommServer` member initialiser (`socketcommserver.cpp:203`), ahead of `listen()` and `saveCommPort()`, but nothing states it, and making the context lazy later would silently break it.

Renaming it to something like `generateAndPublish()` would make the intent obvious, that touches the declaration in selfsignedcert.h and the single call site in securecontextsingleton.cpp:58.

src/server/comm/socketcommserver.cpp
```diff
@@ -96,14 +102,19 @@ uint64_t SocketCommChannel::writeData(const CommChar *data, uint64_t len) {
...
     try {
+        std::lock_guard lock(_socketMutex);
         written = _socket.sendBytes(data, static_cast<int>(len) * commCharSize);
     } catch (Poco::Exception &ex) {
         LOG_ERROR(Log::instance()->getLogger(), "Exception in StreamSocket::sendBytes: " << ex.displayText());
+        lostConnectionCbk();
+        close();
+        _isClosing = true;
         return 0;
     }
```
R-Gld, line 119: Nit: the three other error paths set `_isClosing = true;` first, this one sets it last. Between `close()` and the assignment, `callbackHandler` can still loop on a socket that has just been shut down.

```suggestion
    } catch (Poco::Exception &ex) {
        LOG_ERROR(Log::instance()->getLogger(), "Exception in StreamSocket::sendBytes: " << ex.displayText());
        _isClosing = true;
        lostConnectionCbk();
        close();
        return 0;
    }
```

src/libcommonserver/utility/selfsignedcert.cpp
```diff
@@ -0,0 +1,148 @@
+#include "selfsignedcert.h"
```
R-Gld, line 19: Missing the copyright header.

```suggestion
/*
 * Infomaniak kDrive - Desktop
 * Copyright (C) 2023-2026 Infomaniak Network SA
 *
 * This program is free software: you can redistribute it and/or modify
 * it under the terms of the GNU General Public License as published by
 * the Free Software Foundation, either version 3 of the License, or
 * (at your option) any later version.
 *
 * This program is distributed in the hope that it will be useful,
 * but WITHOUT ANY WARRANTY; without even the implied warranty of
 * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 * GNU General Public License for more details.
 *
 * You should have received a copy of the GNU General Public License
 * along with this program.  If not, see <http://www.gnu.org/licenses/>.
 */

#include "selfsignedcert.h"
```

src/server/mainserver.cpp
```diff
@@ -54,6 +54,9 @@
 #define APP_RLIMIT_STACK 0x4000000
 #endif
 
+#if defined(KD_WINDOWS)
+#include <openssl/applink.c>
+#endif
```
R-Gld, line 60: Two small things:

- Missing blank line before `void signalHandler(...)`.
- Recent `applink.c` wraps itself in `extern "C"`, older ones don't, and if it isn't wrapped, the symbol is C++-mangled, OpenSSL never resolves it at runtime, and the failure is silent and Windows-only. Wrapping the include is harmless either way (nested `extern "C"` is legal).

```suggestion
#if defined(KD_WINDOWS)
extern "C" {
#include <openssl/applink.c>
}
#endif

```

Same include is duplicated in `test/server/test.cpp:21-23`.

src/server/comm/socketcommserver.cpp
LawrenceLeitgib: [e068a40](https://github.com/Infomaniak/desktop-kDrive/pull/2202/commits/e068a40dc6f023d1eca8bf505aac8cbd7b05a2e1)

src/libcommonserver/utility/selfsignedcert.cpp
LawrenceLeitgib: [e068a40](https://github.com/Infomaniak/desktop-kDrive/pull/2202/commits/e068a40dc6f023d1eca8bf505aac8cbd7b05a2e1)

src/server/comm/securecontextsingleton.cpp
LawrenceLeitgib: [116457c](https://github.com/Infomaniak/desktop-kDrive/pull/2202/commits/116457c55c37aeeffdc22557102e838d8bf58cbd)

src/libcommonserver/utility/selfsignedcert.cpp
```diff
@@ -0,0 +1,156 @@
...
+    if (!keychain) {
+        LOG_ERROR(Log::instance()->getLogger(), "Keychain unavailable");
+        return false;
+    }
+
+
+    if (!generate(pem)) return false;
+
+    if (const std::string certKey(certKeychainKey); !keychain->writeToken(certKey, pem.cert)) {
+        LOG_ERROR(Log::instance()->getLogger(), "Failed to store TLS material in the keychain");
```
R-Gld: Leftovers from the rewrite:

- line 95-96: double blank line where the read path was removed,
- line 100: "Failed to store TLS material", only the certificate is stored now, "material" suggested the pair;
- line 99: the init-statement isn't useful we only use `certKey` in the condition.

```suggestion
    if (!generate(pem)) return false;

    if (!keychain->writeToken(std::string(certKeychainKey), pem.cert)) {
        LOG_ERROR(Log::instance()->getLogger(), "Failed to store the TLS certificate in the keychain");
        return false;
    }
```

src/libcommonserver/utility/selfsignedcert.h
```diff
@@ -0,0 +1,42 @@
...
+    public:
+        struct Pem {
+                std::string cert;
+                std::string key;
+        };
+
+        /// Read the certificate/key pair from the keychain, generating and storing
+        /// a new one if absent.
+        /// @return true on success, false otherwise.
+        static bool generateAndPublish(Pem &pem);
```
R-Gld: The doc comment still describes the old behaviour.

```suggestion
        /// Generate a fresh certificate/key pair and publish the certificate to the keychain,
        /// so that the GUI can pin it. The private key never leaves this process.
        /// Must run before the port is published, otherwise a GUI starting in between would pin
        /// the previous certificate.
        /// @return true on success, false otherwise.
        static bool generateAndPublish(Pem &pem);
```

R-Gld (commented):
Two more comments.

R-Gld (approved):

PR #2205, opened 2026-07-24
feat(macOSV4): Continue to login after permissions when no account is logged in

adrien-coye:


adrien-coye:
I did test this branch with a release build

PR #2208, opened 2026-07-27
fix: Bring all windows to foreground on click of toolbar button

killianmathias:
Also works for mac users using the "spaces" feature.

### DIFF
--- a/src/gui4/macOS/kDrive/AppDelegate.swift
+++ b/src/gui4/macOS/kDrive/AppDelegate.swift
@@ -103,6 +103,7 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
     }
 
     @objc func bringAllWindowsToFront() {
+        openMainWindow()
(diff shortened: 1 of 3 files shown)


src/gui4/macOS/kDrive/AppDelegate.swift
```diff
@@ -102,6 +102,24 @@ final class AppDelegate: NSObject, NSApplicationDelegate {
         mainWindow.window?.makeFirstResponder(nil)
     }
 
+    @objc func bringAllWindowsToFront() {
```
adrien-coye, line 105: This needs to keep calling the `openMainWindow` since if all windows are closed clicking on the "systray" does not open the app any more.

adrien-coye (commented):
I did test this PR

adrien-coye (approved):
I did test again using spaces. All good to me, let's go for the beta 👍

PR #2211, opened 2026-07-28
feat(ipc): enable TLS for Linux v4 GUI client with pinned certificate

LawrenceLeitgib:
Depends on #2269

---

<details>
<summary>:information_source: Description en français</summary>

## Résumé
Cette PR sécurise la communication IPC entre le client graphique Linux v4 et le serveur en remplaçant le transport TCP brut par TLS. Le client épingle le certificat auto-signé du serveur lu depuis le trousseau du système via un nouveau `CertReader`.

## Changements
- Remplacement de `QTcpSocket` par `QSslSocket` dans `IpcClient` pour chiffrer la communication IPC sur loopback.
- Ajout de `CertReader` (`communicationlayer/certreader.*`), un accès en lecture seule au certificat TLS public stocké dans le trousseau du système.
- Épinglage du certificat auto-signé comme seule CA de confiance et vérification du nom d'hôte `kDrive-localhost`.
- Centralisation des constantes de trousseau `package` et `service` dans `libcommon/comm.h`, partagées entre le client (`CertReader`) et le serveur (`KeyChainStorage`).
- Mise à jour de `AGENTS.md` pour documenter le transport TLS et `CertReader`.

## Détails techniques
Le client charge le certificat une fois au démarrage depuis le trousseau via `CertReader`, le place comme unique CA dans la configuration SSL du socket, et définit le nom de vérification du pair à `kDrive-localhost` (correspondant au CN du certificat généré côté serveur). Les erreurs SSL sont journalisées et jamais ignorées. Si le certificat n'est pas encore présent dans le trousseau (le serveur ne l'a pas encore publié), `CertReader` renvoie `false` et `IpcClient` planifie une nouvelle tentative de connexion.

## Notes
Aucun test unitaire n'a été ajouté pour `CertReader`. Le certificat épinglé est mis en cache pour la durée de vie du processus ; un redémarrage du client est nécessaire si le serveur régénère son certificat.

</details>

---

## Summary
This PR secures the IPC communication between the Linux v4 GUI client and the server by replacing the raw TCP transport with TLS. The client pins the server's self-signed certificate read from the OS keychain via a new `CertReader`.

## Changes
- Replaced `QTcpSocket` with `QSslSocket` in `IpcClient` to encrypt loopback IPC traffic.
- Added `CertReader` (`communicationlayer/certreader.*`), a read-only accessor for the public TLS certificate stored in the OS keychain.
- Pinned the self-signed certificate as the only trusted CA and verified the peer hostname against `kDrive-localhost`.
- Centralized the keychain `package` and `service` constants in `libcommon/comm.h`, shared between the client (`CertReader`) and the server (`KeyChainStorage`).
- Updated `AGENTS.md` to document the TLS transport and `CertReader`.

## Technical Details
The client loads the certificate once at startup from the keychain via `CertReader`, sets it as the sole CA in the socket's SSL configuration, and sets the peer verify name to `kDrive-localhost` (matching the CN of the server-generated certificate). SSL errors are logged and never ignored. If the certificate is not yet present in the keychain (server has not published it yet), `CertReader` returns `false` and `IpcClient` schedules a connection retry.

## Notes
No unit test was added for `CertReader`. The pinned certificate is cached for the process lifetime; a client restart is required if the server regenerates its certificate.


src/libcommon/comm.h
```diff
@@ -19,6 +19,10 @@
 #pragma once
 #include <cstdint>
 #include <string>
+// keychain key for tls certificate
+inline constexpr char certKeychainKey[] = "kdrive_ipc_tls_cert";
+inline constexpr char package[] = "com.infomaniak.drive";
+inline constexpr char service[] = "desktopclient";
```
R-Gld: `comm.h` shouldn't contain the `constexpr` `package` and `service` constants. `comm.h` is intended to remain dedicated to communication-layer-related code.

Also, they shouldn't have such generic names. Nothing in their names indicates that they're intended for the keychain.

Previously, they were declared as `const std::string`:

```cpp
// src/libcommonserver/keychainmanager/keychainstorage.cpp:27-28
static const std::string package("com.infomaniak.drive");
static const std::string service("desktopclient");
```

They are now declared as `inline constexpr char *`, which causes every keychain call to convert them to `std::string`, making the code less readable.

(once moved and converted to std::string back, don't forget to remove the std::string conversion on the calls)

src/server/CMakeLists.txt
```diff
@@ -5,7 +5,7 @@ else()
 endif()
 
 find_package(Qt6 REQUIRED COMPONENTS Network Svg)
-find_package(Poco 1.13.3 REQUIRED XML)
+find_package(Poco 1.13.3 REQUIRED XML Net NetSSL)
```
R-Gld: ## I wrote this comment on PR #2202 but forgot to publish it. Once fixed, please do not forget to cherry-pick the commit onto the other PR

`NetSSL` is declared here but never linked, and `Crypto` isn't even declared.

`securecontextsingleton.cpp` uses all three families directly: `Poco::Net::Context` and `SecureServerSocket`/`SecureStreamSocket` (NetSSL), `X509Certificate` and `RSAKey` (Crypto), and `SSL_CTX_*` (OpenSSL). Yet no Poco target ever reaches `target_link_libraries(${APPLICATION_EXECUTABLE} ...)` (line 348), the only explicit Poco link in this file is `Poco::XML` on the `updater` target (line 344).

It builds today purely by transitivity:
`${APPLICATION_EXECUTABLE}` -> `updater` (line 348) -> `${libsyncengine_NAME}` (line 345) -> `Poco::NetSSL` (`src/libsyncengine/CMakeLists.txt:243`, no scope keyword so `PUBLIC`) -> `Poco::Crypto`.

That's an accident of the dependency graph, not a declaration. Let's link what we use, see my suggestion on line 348 as well.

```suggestion
find_package(Poco 1.13.3 REQUIRED XML Net NetSSL Crypto)
```

src/server/CMakeLists.txt
R-Gld, line 354: Companion to my comment on line 8: the executable that actually contains `securecontextsingleton.cpp` and `socketcommserver.cpp` links no Poco target at all.

```cpp
target_link_libraries(${APPLICATION_EXECUTABLE}
    Qt6::Svg
    sentry::sentry
    Poco::NetSSL Poco::Crypto
    updater)
```

Note: `Poco::NetSSL` already depends on `Poco::Crypto`, so the latter is technically redundant, I'd still keep it explicit since `SelfSignedCert` and `applyPem()` use `Poco::Crypto` types directly.

test/server/CMakeLists.txt
```diff
@@ -5,7 +5,7 @@ else()
 endif()
 
 find_package(Qt6 REQUIRED COMPONENTS Network Svg)
-find_package(Poco 1.13.3 REQUIRED XML Net)
+find_package(Poco 1.13.3 REQUIRED XML Net NetSSL)
```
R-Gld: Same issue as `src/server/CMakeLists.txt:8`: `NetSSL` is declared, `Crypto` isn't, and neither is linked, `target_link_libraries(${testserver_NAME} ...)` (line 291) only lists `updater`, so the test binary relies on the exact same transitive chain through `libsyncengine`.

```suggestion
find_package(Poco 1.13.3 REQUIRED XML Net NetSSL Crypto)
```

And at line 291:

```cmake
target_link_libraries(${testserver_NAME}
    Poco::NetSSL Poco::Crypto
    updater
)
```

src/libcommonserver/CMakeLists.txt
```diff
@@ -24,6 +24,7 @@ add_definitions(-D_UNICODE)
 set(libcommonserver_SRCS
     # Utility
     utility/utility.h utility/utility.cpp
+    utility/selfsignedcert.h utility/selfsignedcert.cpp
```
R-Gld, line 27: This is the target that *actually* compiles the `Poco::Crypto` usage (`RSAKey`, `X509Certificate` in `SelfSignedCert::generate()`), and it's the most under-declared of the three:

- line 9: `find_package(Poco 1.13.3 REQUIRED Foundation Net JSON Util)`, no `Crypto`
- line 139: `Poco::Foundation Poco::Net Poco::JSON Poco::Util`, no `Crypto`

It doesn't break because `libcommonserver` is `STATIC` (line 70), so the symbols only need to resolve at the final executable link, where `libsyncengine` happens to drag `Poco::Crypto` in.

```cmake
# line 9
find_package(Poco 1.13.3 REQUIRED Foundation Net Crypto JSON Util)
```

```cmake
# line 139
    Poco::Foundation Poco::Net Poco::Crypto Poco::JSON Poco::Util
```

src/gui4/linux/communicationlayer/ipcclient.cpp
```diff
@@ -101,6 +103,19 @@ void IpcClient::connectToServer(const quint16 port) {
...
+ * Loads the pinned TLS certificate from the keychain.
+ * @return true if the certificate was successfully loaded, false otherwise.
+ */
+bool IpcClient::loadPinnedCertificate() {
+    if (!_pinnedCert.isNull()) {
+        return true;
+    }
+    const CertReader reader{std::string(certKeychainKey)};
+    return reader.readCertificate(_pinnedCert);
+}
```
R-Gld, line 133: `_pinnedCert` is cached for the process lifetime and never invalidated, but `attemptInitialConnection()` is a retry loop.

Scenario: the GUI starts and pins C1 -> the server restarts and publishes C2 -> `_hasConnectedOnce` is still `false`, so `onErrorOccurred` (line 252) routes the handshake failure to `scheduleInitialConnectionRetry()` -> next attempt reuses C1 from the cache -> fails again. Forever, at `initialConnectionRetryDelayMs` intervals, with only a throttled warning.

The PR notes say "a client restart is required if the server regenerates its certificate",  but that's an unbounded silent spin, not a degraded mode. Invalidating the cache on handshake failure makes it self-healing:

```cpp
void IpcClient::onSslErrors(const QList<QSslError> &errors) {
    _pinnedCert.clear();   // force a keychain re-read on the next attempt
    for (const QSslError &error: errors) {
        qCWarning(lcIpcClient) << "SSL/TLS error:" << error.errorString();
    }
}
```

(That means `onSslErrors` can no longer be `static`,  see `ipcclient.h:72`.)

src/server/comm/socketcommserver.cpp
```diff
@@ -146,7 +152,21 @@ void SocketCommChannel::callbackHandler() {
...
+            std::lock_guard lock(_socketMutex);
+            avail = _socket.available();
+        }
+        if (avail > 0) return static_cast<uint64_t>(avail);
+        // For TLS sockets, available() may return 0 even when encrypted data is pending.
+        // Fall back to poll() to detect readability.
+        {
+            std::lock_guard lock(_socketMutex);
+            if (_socket.poll(Poco::Timespan(0, 0), Poco::Net::Socket::SELECT_READ)) {
+                return 1; // Data is likely available to be read
```
R-Gld: ## As with the other comments on code outside `./src/gui4/linux`, this should be fixed by cherry-picking the changes into the other PR (#2202) (or done there and cherry pick here ofc).

`bytesAvailable()` isn't used as a boolean by all callers, it's also used as a read size:

```cpp
// src/server/comm/guicommserver.cpp:98
if (uint64_t charRead = readData(data, (std::min)(bytesAvailable(), static_cast<uint64_t>(1024))); charRead > 0)
```

So returning `1` from the `poll()` fallback makes the first `readData()` of every TLS record read a single `CommChar`. It recovers on the next iteration (once a record is decrypted, `SSL_pending` > 0), so it works, but the contract is broken and it's a trap for the next caller.

Two clean options:
- split the concerns: keep `bytesAvailable()` exact, add an `isReadable()` for the poll loop at line 135;
- or at minimum, make `fetchDataToBuffer()` always pass `1024`.

src/gui4/linux/communicationlayer/certreader.cpp
```diff
@@ -0,0 +1,72 @@
...
+CertReader::CertReader(std::string keychainKey) :
+    _keychainKey(std::move(keychainKey)) {}
+
+bool CertReader::readPem(std::string &outPem) const {
+    outPem.clear();
+    keychain::Error error{};
+    outPem = keychain::getPassword(std::string(package), std::string(service), _keychainKey, error);
+
+    if (error.type == keychain::ErrorType::NotFound) {
+        // Entry not present yet — not an error; caller may retry.
```
R-Gld:
```suggestion
        // Entry not present yet, not an error; caller may retry.
```

test/server/comm/testsocketcomm.cpp
```diff
@@ -113,6 +123,11 @@ void TestSocketComm::testServerCallbacks() {
     // Close the client side channel to trigger lost connection callback
     clientSideChannel->close();
 
+    // Force a read on the server side channel to detect the peer close and
+    // trigger lostConnectionCbk() — required because available() can no longer
```
R-Gld:
```suggestion
    // trigger lostConnectionCbk(), required because available() can no longer
```

src/server/comm/socketcommserver.cpp
```diff
@@ -146,7 +152,21 @@ void SocketCommChannel::callbackHandler() {
...
+        {
+            std::lock_guard lock(_socketMutex);
+            avail = _socket.available();
+        }
+        if (avail > 0) return static_cast<uint64_t>(avail);
+        // For TLS sockets, available() may return 0 even when encrypted data is pending.
+        // Fall back to poll() to detect readability.
+        {
+            std::lock_guard lock(_socketMutex);
+            if (_socket.poll(Poco::Timespan(0, 0), Poco::Net::Socket::SELECT_READ)) {
```
R-Gld: Why lock on poll since we saw together that poll is not using the ssl objet not ts ?

src/libcommonserver/utility/selfsignedcert.cpp
```diff
@@ -0,0 +1,155 @@
...
+        LOG_ERROR(Log::instance()->getLogger(), "X509_gmtime_adj (notBefore) failed: " << sslError());
+        return false;
+    }
+    if (X509_gmtime_adj(X509_getm_notAfter(x509), certValiditySeconds) == nullptr) {
+        LOG_ERROR(Log::instance()->getLogger(), "X509_gmtime_adj (notAfter) failed: " << sslError());
+        return false;
+    }
+
+    X509_NAME *const name = X509_get_subject_name(x509);
+    if (X509_NAME_add_entry_by_txt(name, "CN", MBSTRING_ASC, reinterpret_cast<const uint8_t *>("kDrive-localhost"), -1, -1, 0) !=
```
R-Gld: `X509_set_version(x509, 2)` on l57 declares a v3 certificate but no extension is ever added: no `subjectAltName`, no `basicConstraints`, no `keyUsage`/`extendedKeyUsage`.

Hostname verification on the client therefore relies entirely on the **CN fallback**, which Qt and OpenSSL only apply when SAN is *completely* absent, and which is on its way out. And since the client installs this certificate as its sole CA (`ipcclient.cpp:155`), it's acting as a trust anchor without `CA:TRUE`.

src/gui4/linux/communicationlayer/ipcclient.cpp
```diff
@@ -131,8 +146,18 @@ void IpcClient::attemptInitialConnection() {
...
+    }
+
+    QSslConfiguration config = _socket->sslConfiguration();
+    config.setCaCertificates({_pinnedCert});
+    _socket->setSslConfiguration(config);
+    _socket->setPeerVerifyName(QStringLiteral("kDrive-localhost"));
+
     _socket->abort();
-    _socket->connectToHost(QHostAddress::LocalHost, port);
+    _socket->connectToHostEncrypted(QStringLiteral("127.0.0.1"), port);
```
R-Gld: You should set the PeerVerifyName inside the connectToHostEncrypted method (see [qt doc](https://doc.qt.io/qt-6/qsslsocket.html#connectToHostEncrypted-1)), there is a parameters.

src/server/comm/socketcommserver.cpp
```diff
@@ -96,14 +108,19 @@ uint64_t SocketCommChannel::writeData(const CommChar *data, uint64_t len) {
...
     try {
+        std::lock_guard lock(_socketMutex);
         written = _socket.sendBytes(data, static_cast<int>(len) * commCharSize);
     } catch (Poco::Exception &ex) {
         LOG_ERROR(Log::instance()->getLogger(), "Exception in StreamSocket::sendBytes: " << ex.displayText());
+        _isClosing = true;
+        lostConnectionCbk();
+        close();
         return 0;
     }
```
R-Gld, line 127: `writeData()` now calls `lostConnectionCbk()` + `close()` on exception, a behaviour change that isn't mentioned in the PR description.

If `readData()` (lines 78/88) has already reported the loss for the same socket, `lostConnectionCbk` fires twice, and each invocation spawns a postponed-callback thread in `SocketCommServer::execute()` (line ~360). Is that callback idempotent? If not, `joinAndClearPostponedLostConnectionCbks()` gets a duplicate entry for the same channel.

The `_isClosing` flag is set right before each call, gating the callback on `!_isClosing.exchange(true)` would make it fire exactly once.

src/server/comm/securecontextsingleton.cpp
```diff
@@ -0,0 +1,77 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#include "securecontextsingleton.h"
+
+#include "libcommonserver/utility/selfsignedcert.h"
```
R-Gld, line 21: `LOG_ERROR` and `Log::instance()` are used at line 49, but neither `securecontextsingleton.h` nor `selfsignedcert.h` pulls in the logger. This only compiles because `libcommonserver/log/log.h` happens to be in the server's precompiled header (`src/server/CMakeLists.txt:312`).

```suggestion
#include "libcommonserver/utility/selfsignedcert.h"
#include "libcommonserver/log/log.h"
```

src/server/comm/socketcommserver.cpp
```diff
@@ -300,7 +322,38 @@ void SocketCommServer::execute() {
...
+            } catch (Poco::Exception &ex) {
+                LOG_WARN(Log::instance()->getLogger(), "Exception in SecureStreamSocket::close: " << ex.displayText());
+            }
+            continue;
+        }
+
+        // Keep a bounded receive timeout for the channel's lifetime: SSL_read() must never park on
+        // _socketMutex forever (see the comment on readData).
+        secureSocket.setReceiveTimeout(Poco::Timespan(30, 0));
+        secureSocket.setSendTimeout(sndTimeout);
```
R-Gld: I know I suggested this in the other PR but we shouldn't use "magic values", same for the 30s at line 353.

src/server/AGENTS.md
```diff
@@ -50,11 +65,15 @@ rg -n "vfs|Vfs|VFS" src/server/ -g "*.h" -l
 ```
 
 ## Common Gotchas
-- macOS uses **XPC** exclusively for GUI IPC; Windows/Linux use TCP sockets. They are separate implementations — changes to IPC logic may need to be applied to both `xpccommserver_mac.mm` and `socketcommserver.cpp`.
+
+- macOS uses **XPC** exclusively for GUI IPC; Windows/Linux use TCP sockets. They are separate implementations — changes
+  to IPC logic may need to be applied to both `xpccommserver_mac.mm` and `socketcommserver.cpp`.
```
R-Gld: Contradicts the "IPC Architecture" section updated above (line ~40), which now says TLS over TCP.

```suggestion
- macOS uses **XPC** exclusively for GUI IPC; Windows/Linux use TLS over TCP sockets. They are separate implementations —
  changes to IPC logic may need to be applied to both `xpccommserver_mac.mm` and `socketcommserver.cpp`.
```

src/gui4/linux/communicationlayer/certreader.cpp
```diff
@@ -0,0 +1,72 @@
...
+
+namespace KDC {
+
+CertReader::CertReader(std::string keychainKey) :
+    _keychainKey(std::move(keychainKey)) {}
+
+bool CertReader::readPem(std::string &outPem) const {
+    outPem.clear();
+    keychain::Error error{};
+    outPem = keychain::getPassword(std::string(package), std::string(service), _keychainKey, error);
```
R-Gld: We don't need the clear since we assign outPem just after.

```suggestion
    keychain::Error error{};
    outPem = keychain::getPassword(std::string(package), std::string(service), _keychainKey, error);
```

test/server/test.cpp
```diff
@@ -18,6 +18,10 @@
 
 #include "testincludes.h"
 
+#if defined(KD_WINDOWS)
+#include <openssl/applink.c>
+#endif
```
R-Gld, line 25: Mentionned in [this comment on the other PR](https://github.com/Infomaniak/desktop-kDrive/pull/2202#discussion_r3673036721) that this should be wrapped with extern C too but not applied.

src/gui4/linux/communicationlayer/certreader.cpp
```diff
@@ -0,0 +1,72 @@
...
+#include <QLoggingCategory>
+
+#include <utility>
+
+Q_LOGGING_CATEGORY(lcCertReader, "gui.v4.certreader", QtInfoMsg)
+
+namespace KDC {
+
+CertReader::CertReader(std::string keychainKey) :
+    _keychainKey(std::move(keychainKey)) {}
```
R-Gld: Design remark, since there is only one and only possible key, you shouldn't parameterize CertReader with the key. Should be moved from the const directly here.

src/server/comm/socketcommserver.cpp
```diff
@@ -146,7 +152,21 @@ void SocketCommChannel::callbackHandler() {
 
 uint64_t SocketCommChannel::bytesAvailable() const {
     try {
-        return static_cast<uint64_t>((std::max) (0, _socket.available()));
+        int32_t avail = 0;
```
R-Gld: `socket#available` returns an `int`, and `int32_t` is not strictly the same type. You should use the return type of the method here. This is one of the cases where we can ignore Sonar.

R-Gld (commented):
If you make the changes to the server part here, please don't commit everything in a single commit. Split the changes into separate commits so you can easily cherry-pick the relevant ones into the other PR instead of picking everything at once.

Ideally, make one commit per review comment so I can review the changes properly.

src/libcommon/comm.h
LawrenceLeitgib: 9d42089f3f3d53512512bd827d46abfbf99c896c

src/server/CMakeLists.txt
LawrenceLeitgib: 330c746353dc526195c01e470ead82135b159665

src/server/CMakeLists.txt
LawrenceLeitgib: 330c746353dc526195c01e470ead82135b159665

test/server/CMakeLists.txt
LawrenceLeitgib: ad043e00129f1d7ff706fa6e2bb2a4e9d9eae4c9

src/libcommonserver/CMakeLists.txt
LawrenceLeitgib: ad043e00129f1d7ff706fa6e2bb2a4e9d9eae4c9

src/gui4/linux/communicationlayer/ipcclient.cpp
LawrenceLeitgib: 86db6cabe2bab6db3bb9fcaf5c9ce10301f70cff

src/server/comm/socketcommserver.cpp
LawrenceLeitgib: 29b696bcaf04db2c969aa0dcceff5584f1e606b2

src/server/comm/socketcommserver.cpp
LawrenceLeitgib: 29b696bcaf04db2c969aa0dcceff5584f1e606b2

src/libcommonserver/utility/selfsignedcert.cpp
LawrenceLeitgib: b1c74734f7fe32159d3557bfc103b0f8c5b92e4d

src/gui4/linux/communicationlayer/ipcclient.cpp
LawrenceLeitgib: b1c74734f7fe32159d3557bfc103b0f8c5b92e4d

src/server/comm/socketcommserver.cpp
LawrenceLeitgib: f7c1903d20597cd1228952a86c422296a80483c0

src/server/comm/securecontextsingleton.cpp
LawrenceLeitgib: 7becd43ae8fc3d1f1336c741ec7359f4b0a2afbc

src/server/comm/socketcommserver.cpp
LawrenceLeitgib: 1952a14fd34a737a1494cea3e7ce209d29beada3

test/server/test.cpp
LawrenceLeitgib: 8542c4618afbb1fd7c03d329718daf58a0a52078

src/gui4/linux/communicationlayer/certreader.cpp
LawrenceLeitgib: 6eed71f99d2de41cf693a7f182a50ba2e6b7adf1

src/server/comm/socketcommserver.cpp
LawrenceLeitgib: 2abc04f99ff0ff96a238099ae9828df96d348abd

src/gui4/linux/communicationlayer/certreader.h
```diff
@@ -0,0 +1,56 @@
...
+         * @return true if the certificate was successfully read and parsed, false otherwise.
+         */
+        bool readCertificate(QSslCertificate &certificate) const;
+
+    private:
+        /**
+         * Reads the PEM string from the keychain.
+         * @param outPem The output string to hold the PEM data.
+         * @return true if the PEM string was successfully read, false otherwise.
+         */
```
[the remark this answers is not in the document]
LawrenceLeitgib: 5d1d4198a0c019f5cc7657d87d2e8b76aedead62

src/gui4/linux/communicationlayer/certreader.h
```diff
@@ -0,0 +1,44 @@
...
+        /**
+         * Reads and parses the certificate from the keychain.
+         * @param certificate The output QSslCertificate to hold the parsed certificate.
+         * @return true if the certificate was successfully read and parsed, false otherwise.
+         */
+        static bool readCertificate(QSslCertificate &certificate);
+
+    private:
+        static bool readPem(std::string &outPem);
+};
```
R-Gld: You should delete the constructors here and rename readPem to readPemFromKeychain or something like that.

LawrenceLeitgib: Here is the commit that fix [a5ea7b6](https://github.com/Infomaniak/desktop-kDrive/pull/2211/commits/a5ea7b6ff959c13e7928badceeea45b4bb5bdc47), if everything is alright I will start cherry picking the commit into the other developed based branch

test/server/CMakeLists.txt
R-Gld: 🤔

R-Gld (approved):

PR #2186, opened 2026-07-22
feat: Handle drive with no principal synchro

valentinperignon:

PR #2215, opened 2026-07-28
feat: Handle drive with no principal synchro

adrien-coye:
based on https://github.com/Infomaniak/desktop-kDrive/pull/2186


adrien-coye (commented):
I did test the changes, the advanced sync is now available in the menu when I disable the advanced sync.

PR #2222, opened 2026-07-29
fix: Hide back button when unnecessary

killianmathias:


adrien-coye (approved):
Code LGTM. I did test it it works well on my machine.

PR #2224, opened 2026-07-30
fix: Display available storage size more prominently in settings

killianmathias:


adrien-coye (approved):
Nice catch, it does match the system settings now. 👍

PR #2226, opened 2026-07-30
fix: Accent color and dark mode

killianmathias:


adrien-coye (changes requested):
It looks like there is an issue in dark mode. I need to double check something before merging this one.

adrien-coye (approved):
I did test the latest changes all good to me, I no longer see a problem with the dark mode 🙏
Thanks for the latest changes

PR #2230, opened 2026-07-31
fix: Align selected item preference display with main window pattern

killianmathias:


adrien-coye (approved):
Code looks good, fix does work well in my testing. Good job 🤝

PR #2232, opened 2026-07-31
fix(app-server): Avoid a scenario of massive file deletion on Windows

luc-guyot-infomaniak:
These changes add a new column `toDelete` in the `sync` table of `ParmDb` to mark the synchronizations for which the deletion has been requested by the user. If some synchronization deletion fails, `AppServer` will not try to restart such a synchronization, avoiding thus any risk of propagating the deletion of dehydrated placeholders that might have happened during the deletion attempt. (The deletion of dehydrated placeholders is triggered when stopping the VFS on Windows).

### DIFF
--- a/src/server/appserver.cpp
+++ b/src/server/appserver.cpp
@@ -3373,7 +3373,7 @@ ExitInfo AppServer::startSyncs(User &user, const std::unordered_set<SyncDbId> to
                     LOG_WARN(_logger, "Error in checkIfSyncIsValid for syncDbId=" << sync.dbId() << " : " << exitInfo);
                     addError(Error(sync.dbId(), ERR_ID, exitInfo));
 
-                    if (exitInfo.cause() == ExitCause::SyncDeletionFailed) deleteSync(sync.dbId());
+                    if (exitInfo.cause() == ExitCause::SyncDeletionFailed) deleteSyncAsBackgroundTask(sync.dbId());
 
                     continue;
                 }
--- a/src/server/appserver.cpp
+++ b/src/server/appserver.cpp
@@ -2935,11 +2935,11 @@ ExitCode AppServer::migrateConfiguration(bool &proxyNotSupported) {
 
     MigrationParams mp = MigrationParams();
     std::vector<std::pair<migrateptr, std::string>> migrateArr = {
-            {&MigrationParams::migrateGeneralParams, "migrateGeneralParams"},
-            {&MigrationParams::migrateAccountsParams, "migrateAccountsParams"},
-            {&MigrationParams::migrateTemplateExclusion, "migrateFileExclusion"},
+        {&MigrationParams::migrateGeneralParams, "migrateGeneralParams"},
+        {&MigrationParams::migrateAccountsParams, "migrateAccountsParams"},
+        {&MigrationParams::migrateTemplateExclusion, "migrateFileExclusion"},
 #if defined(KD_MACOS)
-            {&MigrationParams::migrateAppExclusion, "migrateAppExclusion"},
+        {&MigrationParams::migrateAppExclusion, "migrateAppExclusion"},
 #endif
     };
 
--- a/src/server/appserver.cpp
+++ b/src/server/appserver.cpp
@@ -653,6 +653,15 @@ void AppServer::quitLater(const int32_t delayMs) {
 // This task can be long and block the GUI
 void AppServer::stopSyncTask(const SyncDbId syncDbId,
                              const SyncPal::DbBehaviorAfterStop behavior /*= SyncPal::DbBehaviorAfterStop::Keep*/) {
+    if (behavior == SyncPal::DbBehaviorAfterStop::Remove) {
+        // Mark the sync for deletion in the parameters DB.
+        if (bool found = false; !ParmsDb::instance()->setSyncToDelete(syncDbId, true, found)) {
+            LOG_WARN(_logger, "Error in setSyncToDelete for syncDbId=" << syncDbId);
+        } else if (!found) {
+            LOG_WARN(_logger, "Sync not found in DB for syncDbId=" << syncDbId);
+        }
+    }
+
     // Stop sync and remove it from syncPalMap
     if (const auto exitInfo = stopSyncPal(syncDbId, SyncPal::PauseCaller::Sync, behavior); !exitInfo) {
         LOG_WARN(_logger, "Error in stopSyncPal for syncDbId=" << syncDbId << " : " << exitInfo);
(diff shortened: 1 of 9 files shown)


src/server/appserver.cpp
```diff
@@ -4005,6 +4014,13 @@ ExitInfo AppServer::updateAllUsersInfo(const UpdateFollowUpAction action) {
 
 ExitInfo AppServer::initSyncPal(const Sync &sync, const NodeSet &blackList, bool start, const std::chrono::seconds &startDelay,
                                 bool resumedByUser, bool firstInit) {
+    if (sync.toDelete()) {
+        LOG_WARN(_logger, "Synchronization with syncDbId="
+                                  << sync.dbId()
+                                  << " should have been deleted, but is still present in the database. It will be ignored.");
+        return {ExitCode::SystemError, ExitCause::SyncDeletionFailed};
```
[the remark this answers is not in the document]
luc-guyot-infomaniak: Addressed by 3e46e1a2a5a7cf6ac2b48c02bc8520f872ed7d36.

src/server/appserver.cpp
```diff
@@ -612,6 +612,15 @@ void AppServer::quitLater(const int32_t delayMs) {
 // This task can be long and block the GUI
 void AppServer::stopSyncTask(const SyncDbId syncDbId,
                              const SyncPal::DbBehaviorAfterStop behavior /*= SyncPal::DbBehaviorAfterStop::Keep*/) {
+    if (behavior == SyncPal::DbBehaviorAfterStop::Remove) {
+        // Mark the sync for deletion in the parameters DB.
+        if (bool found = false; !ParmsDb::instance()->setSyncToDelete(syncDbId, true, found)) {
+            LOG_WARN(_logger, "Error in setSyncToDelete for syncDbId=" << syncDbId);
```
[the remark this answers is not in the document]
luc-guyot-infomaniak: Addressed by 3e46e1a2a5a7cf6ac2b48c02bc8520f872ed7d36.

src/server/appserver.cpp
```diff
@@ -2607,6 +2621,10 @@ ExitInfo AppServer::checkIfSyncIsValid(const BaseSync &sync) {
                                                            << Utility::formatSyncPath(sync_.localPath()));
             return {ExitCode::InvalidSync, ExitCause::SyncDirNestingError};
         }
+
+        if (sync_.toDelete()) {
```
ChristopheLarchier: I would have performed this test solely for the sync passed as a parameter (at the beginning of the method).

src/server/appserver.cpp
```diff
@@ -3213,6 +3231,12 @@ std::string liteSyncActivationLogMessage(const bool enabled, const SyncDbId sync
 
 // This function will pause the synchronization in case of errors.
 ExitInfo AppServer::tryCreateAndStartVfs(const Sync &sync, bool &startPostponed) noexcept {
+    if (sync.toDelete()) {
```
ChristopheLarchier: Seems unnecessary because the check is done in checkIfSyncIsValid.

src/server/appserver.cpp
```diff
@@ -4005,6 +4029,13 @@ ExitInfo AppServer::updateAllUsersInfo(const UpdateFollowUpAction action) {
 
 ExitInfo AppServer::initSyncPal(const Sync &sync, const NodeSet &blackList, bool start, const std::chrono::seconds &startDelay,
                                 bool resumedByUser, bool firstInit) {
+    if (sync.toDelete()) {
```
ChristopheLarchier: Seems unnecessary because the check is done in checkIfSyncIsValid.

src/server/appserver.cpp
```diff
@@ -2607,6 +2621,10 @@ ExitInfo AppServer::checkIfSyncIsValid(const BaseSync &sync) {
                                                            << Utility::formatSyncPath(sync_.localPath()));
             return {ExitCode::InvalidSync, ExitCause::SyncDirNestingError};
         }
+
+        if (sync_.toDelete()) {
+            return {ExitCode::SystemError, ExitCause::SyncDeletionFailed};
```
ChristopheLarchier: Currently we do nothing with this ExitCause::SyncDeletionFailed.
In the caller method AppServer::startSyncs, we could retry to delete the sync in background (i.e. call stopSyncTask with Remove behavior again).

src/server/appserver.cpp
```diff
@@ -612,6 +612,20 @@ void AppServer::quitLater(const int32_t delayMs) {
...
+#endif
+        } else if (!found) {
+            LOG_WARN(_logger, "Sync not found in DB for syncDbId=" << syncDbId);
+        }
+    }
+
+
     // Stop sync and remove it from syncPalMap
     if (const auto exitInfo = stopSyncPal(syncDbId, SyncPal::PauseCaller::Sync, behavior); !exitInfo) {
         LOG_WARN(_logger, "Error in stopSyncPal for syncDbId=" << syncDbId << " : " << exitInfo);
```
ChristopheLarchier, line 671: In case of error here we should return.
The DB may not have been deleted, and the subsequent stopVfs command will remove the dehydrated placeholders.

src/server/appserver.cpp
```diff
@@ -612,6 +612,20 @@ void AppServer::quitLater(const int32_t delayMs) {
 // This task can be long and block the GUI
 void AppServer::stopSyncTask(const SyncDbId syncDbId,
                              const SyncPal::DbBehaviorAfterStop behavior /*= SyncPal::DbBehaviorAfterStop::Keep*/) {
+    if (behavior == SyncPal::DbBehaviorAfterStop::Remove) {
+        // Mark the sync for deletion in the parameters DB.
+        if (bool found = false; !ParmsDb::instance()->setSyncToDelete(syncDbId, true, found)) {
+            LOG_WARN(_logger, "Error in setSyncToDelete for syncDbId=" << syncDbId);
+#if defined(KD_WINDOWS)
+            LOG_FATAL(_logger, "Quitting to avoid risks of unwanted deletion.");
```
ChristopheLarchier: Why not adding an error and returning?

src/server/appserver.cpp
luc-guyot-infomaniak: You are right, it is incorrect as is. Fixed by 2e9b308637052fcfbd27c8e5df9f1d5c2ccc0bce.

src/server/appserver.cpp
luc-guyot-infomaniak: Thanks. I double-checked: indeed, such check is not necessary. Fixed by 2e9b308637052fcfbd27c8e5df9f1d5c2ccc0bce.

src/server/appserver.cpp
luc-guyot-infomaniak: Yes. Done here: 2e9b308637052fcfbd27c8e5df9f1d5c2ccc0bce.

src/server/appserver.cpp
luc-guyot-infomaniak: Done here: 6aa8bd51a664a8e999cf5f7d86ff3a8faf00dbef.

src/server/appserver.cpp
luc-guyot-infomaniak: Done here: 6aa8bd51a664a8e999cf5f7d86ff3a8faf00dbef.

src/server/appserver.cpp
luc-guyot-infomaniak: Done here: 6aa8bd51a664a8e999cf5f7d86ff3a8faf00dbef.

ChristopheLarchier (approved):

PR #2234, opened 2026-07-31
fix(winui): incorporate tls

herve-er:


src/gui4/windows/kDrive client/kDrive client/ServerCommunication/Services/SecureSocketConnection.cs
```diff
@@ -0,0 +1,96 @@
...
+
+                var networkStream = new NetworkStream(socket, ownsSocket: true);
+                var sslStream = new SslStream(networkStream, leaveInnerStreamOpen: false);
+
+                var options = new SslClientAuthenticationOptions
+                {
+                    TargetHost = host,
+                    EnabledSslProtocols = SslProtocols.Tls12 | SslProtocols.Tls13,
+                    RemoteCertificateValidationCallback =
+                        (sender, certificate, chain, errors) => ValidateServerCertificate(certificate, pinnedCertificate)
```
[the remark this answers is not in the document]
herve-er: Done

src/gui4/windows/kDrive client/kDrive client/ServerCommunication/Services/SocketServerCommProtocol.cs
```diff
@@ -312,7 +365,17 @@ private async Task OnReadyReadAsync()
 
                 if (_socket.Available > 0)
                 {
-                    int bytesRead = await _socket.ReceiveAsync(_receiveBuffer).ConfigureAwait(false);
+                    // Read decrypted application data from the TLS stream. A single socket read
+                    // may carry a TLS record that decrypts to more (or fewer) bytes than are
+                    // reported by Socket.Available.
+                    int bytesRead = await _stream.ReadAsync(_receiveBuffer).ConfigureAwait(false);
```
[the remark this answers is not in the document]
herve-er: Done

src/gui4/windows/kDrive client/kDrive client/ServerCommunication/Services/WindowsKeychainStore.cs
```diff
@@ -0,0 +1,92 @@
...
+    /// Reads secrets from the Windows Credential Manager, the store the kDrive server writes to.
+    /// The target name and blob encoding mirror the server-side keychain implementation
+    /// (<c>package.service/user</c>, UTF-8 blob), so the client reads back exactly what the
+    /// server stored.
+    /// </summary>
+    public sealed class WindowsKeychainStore : IKeychainStore
+    {
+        // Must match the server-side keychain identifiers (see keychainstorage.cpp).
+        private const string Package = "com.infomaniak.drive";
+        private const string Service = "desktopclient";
```
R-Gld, line 36: FYI, I told @LawrenceLeitgib to move the keys Package and Service in a common lib to deduplicate the constants since I can use them in the linux GUI4. 

See [the comment](https://github.com/Infomaniak/desktop-kDrive/pull/2211/changes#r3688989070).

```suggestion
        // Must match the server-side keychain identifiers (see ./src/libcommon/utility/utility.h).
        private const string Package = "com.infomaniak.drive";
        private const string Service = "desktopclient";
```

src/gui4/windows/kDrive client/kDrive client/ServerCommunication/Services/SocketServerCommProtocol.cs
R-Gld, line 277: AI finding
-----

With `SslStream`, low-level errors are not surfaced as `SocketException` but as `IOException` or `ObjectDisposedException`, the `catch (SocketException)` block here is dead code. `SocketException` isn't in the documented exception list of [`SslStream.ReadAsync`](https://learn.microsoft.com/en-us/dotnet/api/system.net.security.sslstream.readasync), and [`NetworkStream.ReadAsync`](https://learn.microsoft.com/en-us/dotnet/api/system.net.sockets.networkstream.readasync) already wraps socket errors into an `IOException`.

So those errors land in the generic `catch (Exception)`, which logs and keeps looping. And `IsAuthenticated` stays `true` after a failed read, [`_exception`](https://github.com/dotnet/runtime/blob/release/9.0/src/libraries/System.Net.Security/src/System/Net/Security/SslStream.cs#L462) is only set by the handshake and dispose paths, never by the read path. Concretely:

- **Graceful FIN** → `bytesRead == 0`, `ConnectionLost` raised. This is the only path the tests cover, since `CrashClientConnectionAsync` disposes the socket without `LingerOption(true, 0)`.
- **RST / server killed / decrypt failure** → `IOException`, loop re-enters, `ReadAsync` throws again immediately: hot loop burning a core and flooding the log, `ConnectionLost` never raised.
- **Concurrent `DisposeConnection()`** → `ObjectDisposedException` sets the disposed sentinel, `IsAuthenticated` flips false, loop exits silently, `ConnectionLost` never raised.

Since `ServerCommService.OnConnectionLost` is what closes the app on server loss, in the last two cases the client never exits and every subsequent request fails (`_pollingTask` is also never reset to `null`, so the loop can't be restarted either).

herve-er: Done here: https://github.com/Infomaniak/desktop-kDrive/pull/2234/commits/ba6698595332ce9ab9b25571b52c8bc3d58fad6f

R-Gld (approved):
LGTM

LawrenceLeitgib (approved):
LGTM

R-Gld (approved):

PR #2240, opened 2026-07-31
feat: Add new UI components, animations, and color properties

R-Gld:


R-Gld:
This PR include only SVG's and QML generated animations.

PR #2248, opened 2026-08-03
feat(gui4): Enhance UI error handling and sidebar functionality

R-Gld:
Depends on #2247 

image


src/gui4/linux/app/mainwindow/syncselectormodel.cpp
```diff
@@ -120,6 +131,9 @@ void SyncSelectorModel::rebuild() {
     if (selectedRow() != previousSelectedRow) {
         emit selectedRowChanged();
     }
+    if (unselectedErrorCount() != previousUnselectedErrorCount) {
```
ChristopheLarchier, line 134: Could the unselectedErrorCount() have changed since the last call 6 lines upper if the selected row is the same?
If no, you could move this code into the previous if section.

R-Gld: Yes, it can change independently here.

`unselectedErrorCount()` iterates over `_entries`, which is replaced with data derived from `driveContexts()` between the two calls. Since `rebuild()` is triggered by `syncErrorsChanged`, an unselected synchronization's error count may change while the selected row remains the same. Therefore, this notification needs to remain outside the selected-row condition.

ChristopheLarchier (approved):

R-Gld:
Rebased to be mergeable. https://github.com/Infomaniak/desktop-kDrive/compare/efb5bef7e86a287bf85b98820af0fb3467a5ea67..7b93c17f9f6e26de656850fb2f8a14768044f066

PR #2266, opened 2026-08-06
fix(folder-tree-item-widget): Do not update blacklist paths if the synchronization DB identifier associated to the widget is not known yet

luc-guyot-infomaniak:


luc-guyot-infomaniak:
Abandoned as it has not a high priority with respect the release `3.8.6`. A new PR targeting `develop` will be created.

PR #2273, opened 2026-08-07
fix(folder-tree-item-widget): Do not update blacklist paths if the synchronization DB identifier associated to the widget is not known yet

luc-guyot-infomaniak:
This PR prevents spurious errors from happening during the creation of a non-_LiteSync_ synchronization. 
The errors are caused by a synchronization database identifier improperly set with `0` when the `AppServer` executes a request of type `NODE_PATH`. Although these errors seem harmless, they are reported in the logs and can cause confusion.


R-Gld (approved):
LGTM

PR #2285, opened 2026-08-11
fix(conflicting-files-corrector): Check error paths validity before resolving conflicts

luc-guyot-infomaniak:
In a corrupted `ParmsDb`, the paths recorded in the `error` table can be empty and will cause unwanted files deletion if a conflict resolution is triggered by the user.

This PR makes sure that no local delete operation is performed when the `error` paths are empty.


luc-guyot-infomaniak:
Abandoned since these changes will be merged back from `release/release-3.8.6`, see https://github.com/Infomaniak/desktop-kDrive/pull/2287.

PR #2286, opened 2026-08-11
fix: Continuation crash

adrien-coye:
- Encapsulate XPC queries behind XPCConnectionProvider.sendQuery(_:)
- Prevent callers from accessing raw XPC proxies
- Use remoteObjectProxyWithErrorHandler for server queries and login-agent endpoint requests
- Propagate XPC transport failures through async throws
- Use a lock-protected one-shot continuation gate so the first callback wins synchronously
- Preserve callback arrival order and prevent double-resume crashes
- Restore request-correlated XPC failure logs with request ID, type, error, and elapsed time


adrien-coye:
I did test this and it works nicely to prevent UI crash when the server does crash.
I'm not no worried about a UI crash but way more of a freeze or a not clean state where the user would be locked.
Thus I'm merging this PR for the next beta as this is a meaningful enough QOL improvement.

PR #2292, opened 2026-08-13
feat(macOS): Shortcut creation on new syncro action

adrien-coye:


adrien-coye:
Okay so a bit hacky but got the deprecated Carbon call working from swift.
Tested so to watch for potential future issues with this.

PR #2294, opened 2026-08-13
fix: Invalid token error

killianmathias:


adrien-coye (approved):
Code LGTM 👍
I also tested on my side the remote log-out -> error -> log in again, worked on my side.

This goes nicely into a 4.0.0.7 🥳 

A ticket was open so the server can send the correct error code right away without requirering to restart it to get the correct error code.

PR #2297, opened 2026-08-14
fix(test-update-tree-worker): Ensures that node1 and node2 are deterministically used as parent and child in testResetNodesRecursiveDelete

luc-guyot-infomaniak:
These changes ensure that we retrieves nodes from the update tree in an obvious deterministic way before calling `resetNodes` in the test `testResetNodesRecursiveDelete`.

This avoids the possible unwanted scenario during which the first of the two nodes is the update tree root node.

### DIFF
--- a/test/libsyncengine/update_detection/update_detector/testupdatetreeworker.cpp
+++ b/test/libsyncengine/update_detection/update_detector/testupdatetreeworker.cpp
@@ -1326,17 +1326,17 @@ void TestUpdateTreeWorker::testResetNodesRecursiveDelete() {
                                              [](const auto &pair) { return pair.second->name() == Str("node1"); });
     CPPUNIT_ASSERT(parent != _remoteUpdateTree->_validNodes.end());
 
-    const auto children = std::ranges::find_if(_remoteUpdateTree->_validNodes.begin(), _remoteUpdateTree->_validNodes.end(),
-                                               [](const auto &pair) { return pair.second->name() == Str("node2"); });
-    CPPUNIT_ASSERT(children != _remoteUpdateTree->_validNodes.end());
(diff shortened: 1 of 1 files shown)


test/libsyncengine/update_detection/update_detector/testupdatetreeworker.cpp
ClementKunz, line 1349: Shouldn't we test the existence of those nodes before running `resetNodes`?

luc-guyot-infomaniak: Good point, thanks! Done here: 0c8ecc6c92b790f38e2824b50c91d01d8eb46b97

test/libsyncengine/update_detection/update_detector/testupdatetreeworker.cpp
```diff
@@ -1322,19 +1322,21 @@ void TestUpdateTreeWorker::testResetNodesRecursiveDelete() {
     _remoteUpdateTree->insertNode(node2);
 
 
-    auto parent = _remoteUpdateTree->_validNodes.begin();
-    if (parent->second == _remoteUpdateTree->rootNode()) ++parent;
+    const auto parent = std::ranges::find_if(_remoteUpdateTree->_validNodes.begin(), _remoteUpdateTree->_validNodes.end(),
+                                             [](const auto &pair) { return pair.second->name() == Str("node1"); });
```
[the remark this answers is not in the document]
luc-guyot-infomaniak: Addressed here: 0c8ecc6c92b790f38e2824b50c91d01d8eb46b97

ClementKunz (approved):

PR #2302, opened 2026-08-14
fix(sync-delete-job): Do not consider the absence in SyncDb of an excluded dehydrated placeholder as an error

luc-guyot-infomaniak:
If a dehydrated placeholder has been excluded from the synchronization, its deletion by `SyncLocalDeleteJob::hardDeleteDehydratedPlaceholders` raises an error because the method tries to remove a `SyncDb` database entry that does not exist.

The current changes ignore the expected database error.


test/libsyncengine/jobs/local/testlocaljobs.cpp
```diff
@@ -341,4 +344,45 @@ void KDC::TestLocalJobs::testLocalDeleteJob() {
...
+    _syncPal->setLocalPath(temporaryDirectory.path());
+    (void) std::filesystem::create_directories(localDirPath);
+
+    const bool liteSyncIsEnabled = true;
+    LocalDeleteJobMock deleteJob(_syncPal, SyncPath{localDirPath.filename()}, liteSyncIsEnabled, NodeId{"1234"});
+    // The excluded dehydrated placeholder will be excluded from sync because of the pattern `_blacklisted_20220913_130102` found
+    // in its name.
+    const auto excludedDehydratedPlaceholderName = Str("dehydrated_placeholder_blacklisted_20220913_130102_666.txt");
+
+    std::ofstream ofs{localDirPath / excludedDehydratedPlaceholderName};
```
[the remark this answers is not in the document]
luc-guyot-infomaniak: Fixed here: d992b99a7314871eb3f30ce07a1fa45afdb7d8fa

src/libsyncengine/jobs/local/synclocaldeletejob.cpp
```diff
@@ -218,12 +219,16 @@ ExitInfo SyncLocalDeleteJob::hardDeleteDehydratedPlaceholders() {
...
+            if (const auto exitInfo = hardDelete(entry.path()); !exitInfo) return exitInfo;
 
             const auto relativeLocalPath = CommonUtility::relativePath(_syncPal->localPath(), entry.path());
-            exitInfo = deleteFromDB(relativeLocalPath);
-            if (!exitInfo) return exitInfo;
+
+            if (const auto exitInfo = deleteFromDB(relativeLocalPath);
+                exitInfo == ExitInfo{ExitCode::DataError, ExitCause::DbEntryNotFound} &&
+                ExclusionTemplateCache::instance()->isExcluded(relativeLocalPath)) {
+                continue; // The item was already removed from the DB because it is excluded, so we can ignore this error.
```
ClementKunz: Should we really also check `&& ExclusionTemplateCache::instance()->isExcluded(relativeLocalPath)` in this case? We wanted to remove an item from the DB, it is not in DB -> goal reached.
What do you think? I am afraid that there might be other cases that might be excluded because of this test.

luc-guyot-infomaniak: I have followed your suggestion (see e3dc3352ed8532a55c60b22fc40c39fec53a81aa), with a slight hesitation: if the file has not been excluded as we expected it to be, is it wise to continue erasing?

As long as dehydrated placeholders are detected accurately, this should not be a problem.

ClementKunz (approved):

PR #2307, opened 2026-08-17
feat: Implement file icon resolver and add new SVG icons and translations

R-Gld:


src/gui4/linux/app/fileiconresolver.cpp
```diff
@@ -0,0 +1,216 @@
...
+    if (const auto iconIt = _iconNamesByFileName.constFind(fileName); iconIt != _iconNamesByFileName.cend()) {
+        return *iconIt;
+    }
+
+    QString iconName = iconNameForFileName(fileName);
+    if (iconName.isEmpty()) {
+        iconName = iconNameForMimeType(QMimeDatabase{}.mimeTypeForFile(fileName, QMimeDatabase::MatchExtension));
+    }
+    if (_iconNamesByFileName.size() >= maxCacheEntries) {
+        _iconNamesByFileName.clear();
```
LawrenceLeitgib, line 212: Are we sure its the best idea to clear the cache ? Shouldn't we remove the oldest element (e.g. by putting the cache into another structure to have a better access to the oldest element)

src/gui4/linux/app/fileiconresolver.cpp
```diff
@@ -0,0 +1,216 @@
...
+    }
+    (void) _iconNamesByFileName.insert(fileName, iconName);
```
LawrenceLeitgib, line 214: I am also a bit confused about the usefulness of this cache. We might as well have a cache for each extension instead of the file name, because the extension is the only criterion to determine the icon.

If finding the icon takes too long → caching the extension might be better.

If finding the icon takes as long as finding an element in a cache → we might as well remove the cache.

But caching the whole file name seems a bit pointless.

LawrenceLeitgib (changes requested):

src/gui4/linux/app/fileiconresolver.cpp
LawrenceLeitgib: discussed IRL

src/gui4/linux/app/fileiconresolver.cpp
LawrenceLeitgib: discussed IRL

LawrenceLeitgib (approved):

src/gui4/linux/app/fileiconresolver.cpp
```diff
@@ -0,0 +1,216 @@
...
+#include <QSet>
+
+// this let us use the u""_s instead of QStringLiteral("")
+using namespace Qt::StringLiterals;
+
+namespace KDC {
+
+namespace {
+
+constexpr qsizetype maxCacheEntries = 1024;
```
R-Gld: Fixed in the next PR, :(

PR #2333, opened 2026-08-20
fix(local-file-system-observer): Avoids appending _pendingFileEvents …to itself, which can happen when the local live snapshot is invalid

luc-guyot-infomaniak:
These changes prevent an undefined behaviour in `LocalFileSystemObserverWorker::changesDetected` when called with the argument `_pendingFileEvents` at line 591 while `_liveSnapshot.isValid()` is `false`.

In this case, the method appends (a `const` reference to) the `std::list` variable named `_pendingFileEvents` to itself, which leads to undefined behaviour.

### DIFF
--- a/src/libsyncengine/update_detection/file_system_observer/localfilesystemobserverworker.cpp
+++ b/src/libsyncengine/update_detection/file_system_observer/localfilesystemobserverworker.cpp
@@ -109,17 +109,10 @@ ExitInfo LocalFileSystemObserverWorker::handleDeleteOp(const SyncPath &absoluteP
     return ExitCode::Ok;
 }
 
-ExitInfo LocalFileSystemObserverWorker::changesDetected(const std::list<std::pair<SyncPath, OperationType>> &changes) {
+ExitInfo LocalFileSystemObserverWorker::localChangesDetected(const std::list<std::pair<SyncPath, OperationType>> &changes) {
     const std::scoped_lock lock(_recursiveMutex);
 
-    // Warning: OperationType retrieved from FSEvent (macOS) seems to be unreliable in some cases. One event might contain
-    // several operations. Only Delete event seems to be 100% reliable Move event from outside the synced dir to inside it will
-    // be considered by the OS as move while must be considered by the synchronizer as Create.
-    if (!_liveSnapshot.isValid()) {
-        // Snapshot generation is ongoing, queue the events and process them later
-        _pendingFileEvents.insert(_pendingFileEvents.end(), changes.begin(), changes.end());
-        return ExitCode::Ok;
-    }
+    if (!_liveSnapshot.isValid()) return ExitCode::Ok;
 
     auto tmpChanges = changes;
     for (auto changeIt = tmpChanges.begin(); changeIt != tmpChanges.end(); ++changeIt) {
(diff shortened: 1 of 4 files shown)


src/libsyncengine/update_detection/file_system_observer/localfilesystemobserverworker.cpp
```diff
@@ -578,7 +588,7 @@ ExitInfo LocalFileSystemObserverWorker::generateInitialSnapshot() {
     const std::scoped_lock lock(_recursiveMutex);
     if (!_pendingFileEvents.empty()) {
         LOG_SYNCPAL_DEBUG(_logger, "Processing pending file events");
-        if (const auto exitInfo = changesDetected(_pendingFileEvents); !exitInfo) {
+        if (const auto exitInfo = localChangesDetected(_pendingFileEvents); !exitInfo) {
```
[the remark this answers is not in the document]
luc-guyot-infomaniak: Addressed by 88e17e481651f43924bca80370514deabfcf0d0e.

src/libsyncengine/update_detection/file_system_observer/localfilesystemobserverworker.cpp
```diff
@@ -109,17 +109,10 @@ ExitInfo LocalFileSystemObserverWorker::handleDeleteOp(const SyncPath &absoluteP
     return ExitCode::Ok;
 }
 
-ExitInfo LocalFileSystemObserverWorker::changesDetected(const std::list<std::pair<SyncPath, OperationType>> &changes) {
+ExitInfo LocalFileSystemObserverWorker::localChangesDetected(const std::list<std::pair<SyncPath, OperationType>> &changes) {
```
herve-er: The `local` prefix seems redundant with the class name, which is already quite specific about the local nature of all the operations (`LocalFileSystemObserverWorker`).

herve-er: Edit after further review:

The name is quite misleading, as both `localChangesDetected` and `ChangesDetected` handle locally detected changes.

```suggestion
ExitInfo LocalFileSystemObserverWorker::processDetectedChanges(const std::list<std::pair<SyncPath, OperationType>> &changes) {
```

src/libsyncengine/update_detection/file_system_observer/localfilesystemobserverworker.cpp
```diff
@@ -109,17 +109,10 @@ ExitInfo LocalFileSystemObserverWorker::handleDeleteOp(const SyncPath &absoluteP
...
 
-    // Warning: OperationType retrieved from FSEvent (macOS) seems to be unreliable in some cases. One event might contain
-    // several operations. Only Delete event seems to be 100% reliable Move event from outside the synced dir to inside it will
-    // be considered by the OS as move while must be considered by the synchronizer as Create.
-    if (!_liveSnapshot.isValid()) {
-        // Snapshot generation is ongoing, queue the events and process them later
-        _pendingFileEvents.insert(_pendingFileEvents.end(), changes.begin(), changes.end());
-        return ExitCode::Ok;
-    }
+    if (!_liveSnapshot.isValid()) return ExitCode::Ok;
```
herve-er, line 115: If, for any reason, the `_liveSnapshot` is invalid at this point, returning `OK` will lead to the `_pendingFileEvents` list being dropped without the contained events being processed, which will lead to inconsistencies.

I suggest either returning a `LogicError` or always processing the events.

herve-er (changes requested):

src/libsyncengine/update_detection/file_system_observer/localfilesystemobserverworker.cpp
luc-guyot-infomaniak: Adressed by 640606ed8a949a4536d79a7c3b1c1845e27f4af1.

src/libsyncengine/update_detection/file_system_observer/localfilesystemobserverworker.cpp
luc-guyot-infomaniak: If the snapshot is invalid, then it will be fully rebuilt via `generateInitialSnapshot`.  Thus processing pending events is not useful in this case. In addition, processing the pending events in this case, or returning `LogicError` would break the (unchanged) logic of `LocalFileSystemObserverWorker::changesDetected`: 
```
 if (!_liveSnapshot.isValid()) {
        // Snapshot generation is ongoing, queue the events and process them later
        _pendingFileEvents.insert(_pendingFileEvents.end(), changes.begin(), changes.end());
        return ExitCode::Ok;
    }
```

herve-er: Thank for the precision, LGTM

herve-er (approved):

PR #2337, opened 2026-08-21
fix(linux): set RUNPATH for shared libraries in Conan packages

R-Gld:


conanfile.py
```diff
@@ -104,6 +105,50 @@ def configure(self):
         if self.settings.os == "Macos":
             self.settings.os.version = "10.15"
 
+    def _set_linux_package_runpath(self, package_name):
+        """
+        Set $ORIGIN as RUNPATH on shared libraries owned by a Conan package.
```
luc-guyot-infomaniak:
```suggestion
        Set RUNPATH to $ORIGIN on shared libraries owned by a Conan package.
```

luc-guyot-infomaniak (approved):

PR #2340, opened 2026-08-21
fix(trust-store): add fallback CA bundle probing for Linux

LawrenceLeitgib:
<details>
<summary>:information_source: Description en français</summary>

## Résumé
Le chargement des certificats CA système sur Linux se reposait uniquement sur `SSL_CTX_set_default_verify_paths()`, qui retourne toujours 1 même si les chemins compilés dans OpenSSL n'existent pas sur le système. Sur les distributions non standard (SUSE, Alpine/musl, ou des systèmes avec des liens symboliques cassés), cela provoquait des échecs de vérification TLS silencieux. Les versions modernes de RHEL/Fedora fonctionnaient déjà grâce à un lien symbolique vers `/etc/pki/tls/certs/ca-bundle.crt`.

## Changements
- Utilisation de `X509_get_default_cert_file()` / `X509_get_default_cert_dir()` pour résoudre les chemins par défaut (qui honorent `SSL_CERT_FILE` / `SSL_CERT_DIR`), puis `SSL_CTX_load_verify_locations()` pour charger réellement les certificats
- Vérification du nombre de certificats chargés via `X509_STORE_get0_objects()` pour éviter les faux positifs sur des fichiers vides ou malformés
- Ajout d'une liste de chemins de bundles CA connus par distribution comme repli (RHEL/Fedora, SUSE, Alpine/musl)
- Ajout d'un rapport Sentry en cas d'échec total, pour s'aligner sur le contrat déjà utilisé par macOS et Windows

## Détails techniques
`SSL_CTX_set_default_verify_paths()` ne fait que configurer les chemins dans le `SSL_CTX` sans vérifier qu'ils existent sur le disque et sans charger les certificats. Elle retourne 1 même si le fichier est absent ou vide. L'ancien code retournait donc `true` (faux positif) et chaque handshake TLS échouait silencieusement. La nouvelle approche utilise `load_verify_locations()` qui lit et parse réellement le bundle, puis vérifie le nombre de certificats chargés pour confirmer le succès.

</details>

---

## Summary
Linux CA loading relied solely on `SSL_CTX_set_default_verify_paths()`, which returns 1 even when the compiled-in paths don't exist on disk. On non-standard distros (SUSE, Alpine/musl, or systems with broken symlinks) this caused silent TLS verification failures. Modern RHEL/Fedora already worked thanks to a symlink, but this was not guaranteed. This PR resolves the default paths via `X509_get_default_cert_file()` (which also honors env vars), loads certs explicitly via `SSL_CTX_load_verify_locations()`, and falls back to probing known distro paths.

## Changes
- Use `X509_get_default_cert_file()` / `X509_get_default_cert_dir()` to resolve default paths (honors `SSL_CERT_FILE` / `SSL_CERT_DIR` env vars), then `SSL_CTX_load_verify_locations()` to actually parse and load the certificates
- Verify loaded cert count via `X509_STORE_get0_objects()` to avoid false positives on empty or malformed bundle files
- Added known distro CA bundle paths as fallback for systems where the compiled-in OPENSSLDIR doesn't match reality (RHEL/Fedora, SUSE, Alpine/musl)
- Added Sentry error reporting on total failure, matching the existing macOS and Windows contract

## Technical Details
`SSL_CTX_set_default_verify_paths()` only sets the paths in the `SSL_CTX` without checking they exist on disk and without loading any certificates. It returns 1 even if the file is missing, empty, or a dangling symlink. The old code returned `true` (a false positive) and every TLS handshake would silently fail. The new approach uses `SSL_CTX_load_verify_locations()` which reads and parses the bundle immediately, then checks `X509_STORE_get0_objects()` to confirm at least one certificate was loaded. This works for both the default path (primary, covers Ubuntu/Debian + env var overrides) and the distro-specific fallback paths.

## Testing
Verify on Ubuntu that the default path is resolved and loaded (check the debug log line). On a system with a missing `/etc/ssl/cert.pem`, confirm the fallback probes pick up the alternate distro path and TLS still works.

### DIFF
--- a/src/libcommonserver/utility/truststorehelper_linux.cpp
+++ b/src/libcommonserver/utility/truststorehelper_linux.cpp
@@ -72,10 +72,16 @@ bool tryLoadBundleDir(SSL_CTX *ctx, const char *path) {
 
 // Well-known distro CA bundle paths (Debian/Ubuntu/Arch, RHEL/Fedora, SUSE, Alpine/musl).
 constexpr std::array knownCaBundles = {
-    "/etc/ssl/certs/ca-certificates.crt",
-    "/etc/pki/tls/certs/ca-bundle.crt",
-    "/etc/ssl/ca-bundle.pem",
-    "/etc/ssl/cert.pem",
+        "/etc/ssl/certs/ca-certificates.crt",
+        "/etc/pki/tls/certs/ca-bundle.crt",
+        "/etc/ssl/ca-bundle.pem",
+        "/etc/ssl/cert.pem",
+};
+
+// Well-known hashed CApath directories (loaded lazily; cannot be cert-count verified).
+constexpr std::array knownCaDirs = {
+        "/etc/ssl/certs",
+        "/etc/pki/tls/certs",
 };
 
 } // namespace
@@ -98,7 +104,8 @@ bool TrustStoreHelper::loadSystemCAs(SSL_CTX *ctx) {
             LOG_DEBUG(Log::instance()->getLogger(), "Loaded system CA certificates from SSL_CERT_FILE=" << envFile);
             return true;
         }
-        LOG_WARN(Log::instance()->getLogger(), "SSL_CERT_FILE=" << envFile << " set but no usable certificates loaded, continuing");
+        LOG_WARN(Log::instance()->getLogger(),
+                 "SSL_CERT_FILE=" << envFile << " set but no usable certificates loaded, continuing");
     }
 
     // 1b. SSL_CERT_DIR env var override
@@ -135,10 +142,18 @@ bool TrustStoreHelper::loadSystemCAs(SSL_CTX *ctx) {
         }
     }
 
+    // 3. Last resort: probe well-known hashed CApath directories.
+    for (const char *path: knownCaDirs) {
+        if (tryLoadBundleDir(ctx, path)) {
+            LOG_DEBUG(Log::instance()->getLogger(), "Loaded system CA certificates from directory " << path);
+            return true;
+        }
+    }
+
     LOG_ERROR(Log::instance()->getLogger(), "Failed to load system CA certificates from any known path or default");
     sentry::Handler::captureMessage(sentry::Level::Error, "TrustStoreHelper::loadSystemCAs",
-                                   "Failed to load system CA certificates on Linux (no usable CA bundle found via "
-                                   "SSL_CERT_FILE/SSL_CERT_DIR env vars, OpenSSL default paths, or known distro paths)");
+                                    "Failed to load system CA certificates on Linux (no usable CA bundle found via "
+                                    "SSL_CERT_FILE/SSL_CERT_DIR env vars, OpenSSL default paths, or known distro paths)");
     return false;
 }
 
--- a/src/libcommonserver/utility/truststorehelper_linux.cpp
+++ b/src/libcommonserver/utility/truststorehelper_linux.cpp
@@ -27,7 +27,9 @@
 #include <openssl/ssl.h>
 #include <openssl/x509.h>
 
+#include <array>
 #include <cstdlib>
+#include <cstdint>
 #include <unistd.h>
 
 namespace KDC {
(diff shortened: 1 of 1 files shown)


src/libcommonserver/utility/truststorehelper_linux.cpp
```diff
@@ -20,29 +20,126 @@
...
+    if (!path || path[0] == '\0' || access(path, R_OK) != 0) {
+        return false;
+    }
+    if (SSL_CTX_load_verify_locations(ctx, path, nullptr) == 1 && certCount(ctx) > 0) {
+        return true;
+    }
+    return false;
+}
+
+// Try to load from a hashed cert directory.  Unlike file loads, certs in a
```
luc-guyot-infomaniak:
```suggestion
// Try to load from a hashed cert directory. Unlike file loads, certs in a
```

src/libcommonserver/utility/truststorehelper_linux.cpp
```diff
@@ -20,29 +20,126 @@
...
+// CApath directory are loaded lazily at handshake time, so certCount() is
```
luc-guyot-infomaniak:
```suggestion
// CA path directory are loaded lazily at handshake time, so certCount() is
```

src/libcommonserver/utility/truststorehelper_linux.cpp
```diff
@@ -20,29 +20,126 @@
...
+// always 0 here — we can only trust the return value of load_verify_locations.
+bool tryLoadBundleDir(SSL_CTX *ctx, const char *path) {
```
luc-guyot-infomaniak, line 66: **Suggestion.** Reuse the code of this function inside `tryLoadBundleFile` by means of a new helper, say `tryLoadBundle` or `tryLoadBundleItem`.

src/libcommonserver/utility/truststorehelper_linux.cpp
```diff
@@ -20,29 +20,126 @@
...
+    if (!path || path[0] == '\0' || access(path, R_OK) != 0) {
+        return false;
+    }
+    return SSL_CTX_load_verify_locations(ctx, nullptr, path) == 1;
+}
+
+// Well-known distro CA bundle paths (Debian/Ubuntu/Arch, RHEL/Fedora, SUSE, Alpine/musl).
```
luc-guyot-infomaniak, line 73: If it is doable, you could break the comment line into smaller comments and assign them to specific strings below.

src/libcommonserver/utility/truststorehelper_linux.cpp
```diff
@@ -20,29 +20,126 @@
...
+        if (tryLoadBundleDir(ctx, defaultDir)) {
+            LOG_DEBUG(Log::instance()->getLogger(), "Loaded system CA certificates from default directory " << defaultDir);
+            return true;
+        }
+    }
+
+    // 2. Fallback: probe well-known distro paths for systems where the compiled-in
+    // OPENSSLDIR doesn't match reality (RHEL/Fedora, SUSE, musl/Alpine).
+    for (const char *path: knownCaBundles) {
+        if (tryLoadBundleFile(ctx, path)) {
```
luc-guyot-infomaniak, line 139: Do we know, when to look for a directory or a file? Can we easily determine which case to handle? If not, what about trying both?

src/libcommonserver/utility/truststorehelper_linux.cpp
LawrenceLeitgib: This PR will be closed so not needed for now

src/libcommonserver/utility/truststorehelper_linux.cpp
LawrenceLeitgib: same as above

src/libcommonserver/utility/truststorehelper_linux.cpp
LawrenceLeitgib: From what I understood default is directory first in most cases

LawrenceLeitgib:
Closing this PR. We decided the fallback to alternative CA paths was too prone to subtle bugs to be worth keeping. If a user hits a trust-store issue, they can point OpenSSL at the right location themselves via the SSL_CERT_FILE (bundle file) or SSL_CERT_DIR (hashed cert directory) environment variables.

PR #2351, opened 2026-08-25
fix: fix mac os server translations

ClementKunz:
## Summary
- fix translation loading for server notifications in development builds
- generate Qt translations before loading gui4 so macOS bundle install picks up client_*.qm files
- keep release package notifications localized according to OS language


ClementKunz:
Done in https://github.com/Infomaniak/desktop-kDrive/pull/2336

PR #2365, opened 2026-08-27
fix(long-poll-job): Do not send a LongPoll request in absence of an access token. Throw a InvalidTokenError instead

luc-guyot-infomaniak:
This PR ensures that no `listing/longpoll` is sent if the access token is missing at the time of constructing the `LongPollJob` object.

This allows us to skip a useless request and write a clearer error message in the log of the server component when the token has been remotely revoked.


R-Gld (approved):
LGTM

PR #2377, opened 2026-08-31
fix(macOS): Serialize concurrent permission status requests

adrien-coye:


adrien-coye (commented):
code looks good, I did produce an installer on top of this branch and tested the installer / onboarding and tested the full disk access querry. All good to me.

PR #2384, opened 2026-08-31
fix(extended-tests): Use symlinks with relative paths as targets to test symlink synchronization

luc-guyot-infomaniak:


luc-guyot-infomaniak:
Superseded by https://github.com/Infomaniak/desktop-kDrive/pulls/luc-guyot-infomaniak.

PR #2389, opened 2026-09-01
feat: add support for client entry point

herve-er:
# Allow kDrive to be launched from the client executable & bring the app to the foreground on manual launch

## Launch from the client executable

When a user pins kDrive to the taskbar, the pinned shortcut launches the **client executable** directly.

image

This scenario is currently unsupported because:

* The communication port is not provided to the client.
* The TLS materials are not available.
* There is no guarantee that a server instance is already running.

This PR adds support for this scenario with the following flow:

1. If a client is launched and is unable to connect to the server, it starts a new server instance and then closes itself.
2. The standard server startup mechanism then takes over:

   * If another server is already running, the new server asks it to bring its associated client to the foreground and then exits.
   * If no server is running, it starts normally, initializes the communication port and TLS materials, and then properly launches a new client.
> [!NOTE]
> The server is launched with the `--synthesis` option, ensuring that the application is brought to the foreground immediatly

## Show the app in the foreground when manually launched

Currently, if kDrive is not running at all, manually starting it only makes it appear in the taskbar/system tray, potentially hidden in the submenu, even when the `--synthesis` option is used.

This leads to a frustrating experience where the user has to launch the application a second time for the UI to be brought to the foreground.

To address this, this PR ensures that `--synthesis` (and `--settings`) are handled from the initial server launch and forwarded to the client. The client therefore knows immediately whether it should bring the UI to the foreground.

The application shortcuts now launch kDrive with the `--synthesis` argument, ensuring that **any manual launch results in visible UI feedback**, regardless of whether kDrive was already running.

The **automatic startup behavior remains unchanged**.


src/gui4/windows/kDrive client/kDrive client/ServerCommunication/Services/TcpServerCommClient.cs
```diff
@@ -184,76 +184,59 @@ public async Task<bool> InitConnection(CancellationToken cancellationToken)
...
+            {
+                Logger.Log(Logger.Level.Info, "Connection initialization was canceled.");
+                
+            }
+            catch (Exception ex) when (ex is SocketException or System.Security.Authentication.AuthenticationException or IOException)
+            {
+                Logger.Log(Logger.Level.Warning, $"Connection failed: {ex.Message}");
+             
             }
+            ConnectionLost?.Invoke(this, new ConnectionLostArgs(ConnectionLostArgs.ConnectionLostReason.ServerUnreachable));
```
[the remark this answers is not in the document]
herve-er: * The server launches the client only once the comm port and TLS materials are ready.
* When the server is launched, it will start a new client or show any existing one.

src/gui4/windows/kDrive client/kDrive client/App.xaml.cs
```diff
@@ -114,10 +118,36 @@ protected override async void OnLaunched(LaunchActivatedEventArgs args)
...
+
+                if (arguments.Contains("--synthesis"))
+                {
+                    Logger.Log(Logger.Level.Info, "--synthesis arg detected, opening Synthesis window in foreground.");
+                    CreateWindow(CreateWindowOptions.Foreground);
+                }
+                else if (arguments.Contains("--settings"))
+                {
+                    Logger.Log(Logger.Level.Info, "--settings arg detected, opening Settings window in foreground.");
+                    CreateWindow(CreateWindowOptions.Foreground);
```
ChristopheLarchier: Should be OpenSettings I guess.

herve-er: Done here: https://github.com/Infomaniak/desktop-kDrive/pull/2389/commits/0764d1b3492ad232397c746843f3bb29bf3c030d

ChristopheLarchier (approved):

PR #2397, opened 2026-09-02
fix(macOS): Show sync folder in system errors

adrien-coye:
Screenshot 2026-09-02 at 14 00 28


valentinperignon (approved):

adrien-coye (commented):
I did validate this with a ramdisk, it does now correctly explicit the folder name when the device is removed.

PR #2401, opened 2026-09-02
fix(macOS): Keep drive list coherent after sync deletion

adrien-coye:


adrien-coye:
I was looking for a quick fix / quick win. This does not pass the quality bar.

PR #2414, opened 2026-09-04
refactor(parameters): unify Parameters and ParametersInfo models

ClementKunz:
## Summary

This PR merges the duplicated `Parameters` and `ParametersInfo` models into a single `Parameters` class in `src/libcommon/data`.

## Changes
- combine the shared DTO fields and serialization logic in the unified `Parameters` model
- remove the legacy `ParametersInfo` conversion layer and update server, GUI, and DB callers
- migrate dialog geometry handling to the QMap-based API used by the shared model
- refresh the corresponding tests and CMake entries

## Validation
- CMake builds for libcommon, libparms, libsyncengine, server, GUI, and all test targets succeeded
- direct test execution is blocked in this sandbox by a pre-existing `sentry_init` initialization failure unrelated to the refactor


src/server/requests/serverrequests.cpp
```diff
@@ -278,13 +278,13 @@ ExitCode ServerRequests::getSyncList(std::vector<Sync> &list) {
...
-    parametersToParametersInfo(ParametersCache::instance()->parameters(), parametersInfo);
+ExitCode ServerRequests::getParameters(Parameters &parametersInfo) {
+    parametersInfo = ParametersCache::instance()->parameters();
     return ExitCode::Ok;
 }
 
-ExitCode ServerRequests::updateParameters(const ParametersInfo &parametersInfo) {
-    parametersInfoToParameters(parametersInfo, ParametersCache::instance()->parameters());
+ExitCode ServerRequests::updateParameters(const Parameters &parametersInfo) {
+    ParametersCache::instance()->parameters() = parametersInfo;
```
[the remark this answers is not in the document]
ClementKunz: f55f3782c21f40d43631a33ab8af7f7281d83d98

src/server/requests/serverrequests.cpp
```diff
@@ -278,19 +278,19 @@ ExitCode ServerRequests::getSyncList(std::vector<Sync> &list) {
...
-    parametersToParametersInfo(ParametersCache::instance()->parameters(), parametersInfo);
+ExitCode ServerRequests::getParameters(ServerParameters &parameters) {
+    parameters = ParametersCache::instance()->parameters();
     return ExitCode::Ok;
 }
 
-ExitCode ServerRequests::updateParameters(const ParametersInfo &parametersInfo) {
-    parametersInfoToParameters(parametersInfo, ParametersCache::instance()->parameters());
+ExitCode ServerRequests::updateParameters(const ServerParameters &parameters) {
+    ParametersCache::instance()->parameters() = parameters;
```
R-Gld: This copies the whole ServerParameters, including the server-only fields (`seenVersion`, `uploadSessionParallelJobs`, `updateFileAvailable`, `updateTargetVersion(String)`, `autoUpdateAttempted`).

The GUI never knows those values, both callers build a fresh object before reading the client payload (PARAMETERS_UPDATE at appserver.cpp:2233, and `ParametersUpdateJob` via `dynamicVar2Struct<ServerParameters>`). 

So every GUI settings save resets them to their defaults, in the cache and in the DB.

R-Gld: Opus 5.5 finding

Side effect of the full copy: this now writes `std::string` members (`seenVersion`, `updateFileAvailable`, ...) from a GuiJobManager pool thread (`ParametersUpdateJob::process` is queued via `queueAsyncJob`), while the UpdateManager thread reads `seenVersion()` in `AbstractUpdater::isVersionSkipped` and executor workers read `uploadSessionParallelJobs()`. Concurrent write/read on a non-atomic std::string is UB (torn read, crash in the copy). Before this PR `updateParameters` never touched those fields. Assigning only the `Parameters` base subobject removes the race for these fields.

src/gui4/macOS/CppInterop/interop/cstypes.h
R-Gld: This file (and `interop/comm.h`, `XPC/xpcGuiProtocol.h`, `XPC/xpcLoginItemProtocol.h`) went from a symlink to a regular file.

Maybe a checkout with `core.symlinks=false` ?

src/libcommonserver/data/serverparameters.h
```diff
@@ -0,0 +1,78 @@
...
+                   purgeOldLogs() == parameters.purgeOldLogs() && proxyConfig() == parameters.proxyConfig() &&
+                   darkTheme() == parameters.darkTheme() && dialogGeometry() == parameters.dialogGeometry() &&
+                   maxAllowedCpu() == parameters.maxAllowedCpu() && distributionChannel() == parameters.distributionChannel() &&
+                   sentryEnabled() == parameters.sentryEnabled() && matomoEnabled() == parameters.matomoEnabled() &&
+                   _updateFileAvailable == parameters._updateFileAvailable &&
+                   _updateTargetVersion == parameters._updateTargetVersion &&
+                   _updateTargetVersionString == parameters._updateTargetVersionString &&
+                   _autoUpdateAttempted == parameters._autoUpdateAttempted && _seenVersion == parameters._seenVersion &&
+                   _uploadSessionParallelJobs == parameters._uploadSessionParallelJobs;
+        }
```
R-Gld: Nit, if we compare all the fields of the class here, we sould use the `= default`.

```suggestion
        bool operator==(const ServerParameters &parameters) const = default;
```

src/libcommon/data/parameters.h
```diff
@@ -28,58 +28,83 @@
...
         [[nodiscard]] DistributionChannel distributionChannel() const { return _distributionChannel; }
         void setDistributionChannel(const DistributionChannel channel) { _distributionChannel = channel; }
+
         bool sentryEnabled() const { return _sentryEnabled; }
-        void setSentryEnabled(bool value) { _sentryEnabled = value; }
-        bool matomoEnabled() const { return _matomoEnabled; }
-        void setMatomoEnabled(bool value) { _matomoEnabled = value; }
+        void setSentryEnabled(const bool value) { _sentryEnabled = value; }
 
-        friend bool operator==(const ParametersInfo &lhs, const ParametersInfo &rhs) = default;
```
R-Gld: Shoudln't we have it here too ? 

```cpp
bool operator==(const Parameters &) const = default;
```
(sry github does not let me do a suggestion)

src/libcommon/data/parameters.h
```diff
@@ -28,58 +28,83 @@
...
 
+#include <string>
+
 namespace KDC {
 
-class ParametersInfo {
+class Parameters {
     public:
+        Parameters() = default;
+        virtual ~Parameters() = default;
```
R-Gld: Why do we need the dtor to be virtual here ?

src/libparms/db/parmsdb.cpp
```diff
@@ -18,18 +18,56 @@
...
 
 #include <sqlite3.h>
 
 #include <fstream>
 #include <string>
 
+namespace {
+
+// Encodes the dialog geometry map into the BLOB format stored in the "parameters" table:
+// one "objectName;base64Geometry" pair per line.
```
R-Gld: The comment says "objectName;base64Geometry", but `dialogGeometryToBlob` stores the raw binary `saveGeometry()` bytes (no base64). Since `blobToDialogGeometry` splits on '\n' and ';', any geometry whose bytes contain 0x0A (e.g. x=10, width=522) or 0x3B is dropped or mangled on reload.

The issue is inherited but since this code is rewritten here, either fix the comment to describe the actual format or encode the value with `toBase64()` / `fromBase64()` (but this would need a migration for existing rows ...).

src/libparms/db/parmsdb.h
```diff
@@ -36,6 +36,7 @@
 #include "libcommonserver/db/db.h"
 
 #include "libcommon/data/drive.h"
+#include "../../libcommonserver/data/serverparameters.h"
```
R-Gld:
```suggestion
#include "libcommonserver/data/serverparameters.h"
```

src/libparms/db/parmsdb.cpp
```diff
@@ -18,18 +18,56 @@
 
 #include "parmsdb.h"
 
-
 #include "libcommon/utility/utility.h"
 #include "libcommon/utility/logiffail.h"
 
 #include "libcommonserver/io/iohelper.h"
 #include "libcommonserver/utility/utility.h"
+#include "../../libcommonserver/data/serverparameters.h"
```
R-Gld:
```suggestion
#include "libcommonserver/data/serverparameters.h"
```

src/libsyncengine/requests/parameterscache.h
```diff
@@ -18,8 +18,8 @@
 
 #pragma once
 
+#include "../../libcommonserver/data/serverparameters.h"
```
R-Gld:
```suggestion
#include "libcommonserver/data/serverparameters.h"
```

src/server/requests/serverrequests.h
```diff
@@ -28,12 +28,11 @@
 #include "libcommon/info/nodeinfo.h"
 #include "libcommon/data/syncfileitem.h"
 #include "libcommon/data/error.h"
-#include "libcommon/info/parametersinfo.h"
 #include "libcommon/data/proxyconfig.h"
 #include "libcommon/data/exclusiontemplate.h"
 #include "libcommon/data/exclusionapp.h"
-#include "libparms/db/parameters.h"
 #include "libsyncengine/login/login.h"
+#include "../../libcommonserver/data/serverparameters.h"
```
R-Gld:
```suggestion
#include "libcommonserver/data/serverparameters.h"
```

src/libcommonserver/data/serverparameters.h
```diff
@@ -0,0 +1,78 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+
+#pragma once
+
+#include "data/parameters.h"
```
R-Gld: Since libcommonserver and libcommon has their own `data` folder, maybe, we could precise completely

```suggestion
#include "libcommonserver/data/parameters.h"
```

src/libparms/db/parmsdb.cpp
```diff
@@ -18,18 +18,56 @@
...
+    return std::make_shared<std::vector<char>>(arr.begin(), arr.end());
+}
+
+KDC::Parameters::DialogGeometry blobToDialogGeometry(const std::shared_ptr<std::vector<char>> &blob) {
+    KDC::Parameters::DialogGeometry dialogGeometry;
+    if (!blob) {
+        return dialogGeometry;
+    }
+
+    const QByteArray arr(blob->data(), static_cast<int>(blob->size()));
```
R-Gld: Really nitpick, but the [ctor of QByteArray](https://doc.qt.io/qt-6/qbytearray.html#public-functions) takes a qsizetype as second parameters.

```suggestion
    const QByteArray arr(blob->data(), static_cast<qsizetype>(blob->size()));
```

src/server/comm/guijobs/parametersinfojob.h
```diff
@@ -20,7 +20,6 @@
 
 #include "server/comm/guijobs/abstractguijob.h"
 
-#include "libcommon/info/parametersinfo.h"
```
R-Gld: Nitpick too but this class compiles only thanks to the transitive includes under abstractguijob.h since we keep a ServerParameters field but not the include associate.

Same in parametersupdatejob.h:32

src/libcommon/data/parameters.cpp
```diff
@@ -0,0 +1,122 @@
...
+static const auto parametersMonoIcons = "monoIcons";
+static const auto parametersAutoStart = "autoStart";
+static const auto parametersMoveToTrash = "moveToTrash";
+static const auto parametersNotificationsDisabled = "notificationsDisabled";
+static const auto parametersUseLog = "useLog";
+static const auto parametersLogLevel = "logLevel";
+static const auto parametersExtendedLog = "extendedLog";
+static const auto parametersPurgeOldLogs = "purgeOldLogs";
+static const auto parametersProxyConfigInfo = "proxyConfigInfo";
+static const auto parametersDarkTheme = "darkTheme";
```
R-Gld: Nit: `parametersDarkTheme` is only used under `#ifdef KD_MACOS` (l.53 and l.90) but is defined on all platforms, so Clang on Linux emits -Wunused-const-variable. I think we should move it back under `#ifdef KD_MACOS` or mark it `[[maybe_unused]]`.

src/libcommon/data/parameters.cpp
```diff
@@ -0,0 +1,122 @@
...
+                    assert(value.isStruct());
+                    const auto &structValue = value.extract<Poco::DynamicStruct>();
+                    DialogGeometry dialogGeometry;
+
+                    for (const auto &[key, blob64]: structValue) {
+                        const auto blob64Str = blob64.convert<std::string>();
+                        CommString commStr;
+                        CommonUtility::convertFromBase64Str(blob64Str, commStr);
+                        std::string str = CommonUtility::commString2Str(commStr);
+                        (void) dialogGeometry.insert(QString::fromStdString(key), QByteArray(str.data()));
```
R-Gld: Agree with the [copilot comment](https://github.com/Infomaniak/desktop-kDrive/pull/2414/changes#r4091175637) (`saveGeometry()` blobs start with `01 D9 D0 CB 00 03`)

src/libcommonserver/data/serverparameters.h
```diff
@@ -0,0 +1,78 @@
...
+
+namespace KDC {
+
+class ServerParameters : public Parameters {
```
R-Gld: [Sonar comment related](https://sonarcloud.io/project/issues?id=Infomaniak_desktop-kDrive&issues=AaDOEYG9Zg7Gg0ShnavQ&open=AaDOEYG9Zg7Gg0ShnavQ&pullRequest=2414).

Same root cause as the `virtual ~Parameters()` comment: the user-declared destructor suppresses the implicit move of `Parameters`, so `ServerParameters`'s move falls back to copying its base (not noexcept because of the ProxyConfig strings).

R-Gld (commented):
A lot of nitpick but only the the comments on serverrequests.cpp:287 and cstypes.h are blocking.

PR #2453, opened 2026-09-16
fix: prevent stale errors from being added

herve-er:


src/server/requests/serverrequests.cpp
```diff
@@ -1586,6 +1586,16 @@ ExitInfo ServerRequests::getErrorList(const int limit, std::vector<Error> &list)
...
     list.clear();
     for (const Error &error: errorList) {
+        if (error.isStale()) {
+            bool found = false;
+            LOG_INFO(Log::instance()->getLogger(), "Deleting stale error with dbId: " << error.dbId()); 
+
+            if (!ParmsDb::instance()->deleteError(error.dbId(), found)) {
+                LOG_WARN(Log::instance()->getLogger(), "Error in ParmsDb::deleteError");
+            }
+            continue;
```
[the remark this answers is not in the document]
herve-er: Done

src/server/requests/serverrequests.cpp
```diff
@@ -1578,18 +1578,38 @@ ExitCode ServerRequests::getErrorList(const ErrorLevel level, const SyncDbId syn
...
+                }
+                continue;
+            }
+
+
+            if (isDisplayableError(error)) {
+                list.push_back(error);
+            }
+        }
+    } while (staleCount > maxStaleCount);
```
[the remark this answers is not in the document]
herve-er: Done

src/server/requests/serverrequests.h
```diff
@@ -74,7 +74,7 @@ struct SYNCENGINE_EXPORT ServerRequests {
         static ExitCode setExclusionAppList(bool def, const QList<ExclusionApp> &list);
 
         static ExitCode getErrorList(ErrorLevel level, SyncDbId syncDbId, int limit, QList<Error> &list);
-        static ExitInfo getErrorList(int limit, std::vector<Error> &list);
+        static ExitInfo getErrorList(int32_t limit, std::vector<Error> &list, bool &hasMore);
```
ClementKunz, line 77: If I understand it correctly, the first method is used in the old comm layer, and the second one is used in the new comm layer. Perhaps just add a short comment explaining this in order to simplify cleaning once the old comm layer is not required anymore.

src/server/requests/serverrequests.cpp
```diff
@@ -1577,20 +1577,53 @@ ExitCode ServerRequests::getErrorList(const ErrorLevel level, const SyncDbId syn
...
 
+    return true;
+}
+
+} // namespace
+
+ExitInfo ServerRequests::getErrorList(const int32_t limit, std::vector<Error> &list, bool &hasMore) {
+    constexpr int32_t maxStaleCount = 10;
+
+    int32_t staleCount;
```
ClementKunz: Should be initialized

src/server/requests/serverrequests.cpp
```diff
@@ -1577,20 +1577,53 @@ ExitCode ServerRequests::getErrorList(const ErrorLevel level, const SyncDbId syn
...
+            if (deleteIfStale(error)) {
+                ++staleCount;
+                continue;
+            }
+
+            if (isDisplayableError(error)) {
+                list.push_back(error);
+            }
+        }
+    } while (staleCount > maxStaleCount);
```
ClementKunz, line 1625: I am not sure I understand this logic. If there are more than 10 stale errors, we query again the first N (where N = `limit`) errors; otherwise, we just return the error list. We should try to match the required number of errors asked (`limit`) if possible, don't you think?

src/server/requests/serverrequests.h
herve-er: That's it. There is already a general comment about `ServerRequests` at the top of the class definition 😉 :
https://github.com/Infomaniak/desktop-kDrive/blob/0a7318d2f3cdb055fb82b3009d5ac5d79213ef35/src/server/requests/serverrequests.h?plain=1#L47C35-L47C35

src/server/requests/serverrequests.cpp
herve-er: Exactly matching the limit was the original intent, and the Windows GUI was mistakenly relying on it to determine whether more errors were available.

But I then realized that, because of `isDisplayableError(...)`, there were already many cases where we wouldn't return exactly `limit` errors even if we had fetched `limit` or more.

That's why I decided to treat `limit` as a true maximum rather than as an implicit min/max value, and added a `hasMore` flag instead.

From there, I chose to tolerate up to 10 stale errors to avoid refetching and reprocessing 1,000 errors (`limit` is currently set to 1,000 by both the Windows and macOS GUIs).

Let me know if this sounds good to you.

I admit this is a niche optimization with very limited impact. If you prefer, I can totally refetch as soon as at least one stale error is detected.

src/server/requests/serverrequests.cpp
herve-er: Done

src/server/requests/serverrequests.cpp
ClementKunz: I do agree that we might not need to re-fetch all the errors when only a few have been removed.
Let's keep it that way, thanks 👍

ClementKunz (approved):

ClementKunz (approved):

src/server/requests/serverrequests.cpp
```diff
@@ -1224,7 +1224,7 @@ bool ServerRequests::isDisplayableError(const Error &error) {
         case TooManyDeleteOperations:
             return false;
         case Unknown: {
-            return error.inconsistencyType() != InconsistencyType::PathLength;
+            return error.inconsistencyType() != InconsistencyType::PathLength && !error.isStale();
```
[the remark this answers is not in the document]
ClementKunz: No error should exist with `ExitCode::Ok`

PR #2473, opened 2026-09-18
fix: fix RecoveryUpdater build

herve-er:
This pull request updates the `Package-RecoveryUpdater` function in `build-drive.ps1` to support selecting between the old and new GUI when packaging the recovery updater. The main change is the addition of a `newGui` boolean parameter, which is now properly passed through the build pipeline and used to determine the icon path.

The build was previously failling as icon path is not the same between legacy gui and the new one


R-Gld (approved):
LGTM

PR #2476, opened 2026-09-21
fix: check installer signature in kDrive recovery updater

herve-er:
fix: check installer signature in kDrive recovery updater


R-Gld (approved):
LGTM

